PR #17, opened 2020-09-11
Enforce minimum gas price on local transactions

### LINKED ISSUE #16: Issuing invalid transaction stops node from processing messages
Issuing an invalid transaction to a node currently causes that node to put it in a block and then stop being able to process additional messages.

This can be reproduced in avash by starting a local network with `scripts/five_node_staking.lua`.

1) Create a new account in metamask
2) Create a new keystore user and import the genesis private key on both the X and C chains
3) ExportAVAX from the X chain to the genesis private key address on the C chain
4) ImportAVAX on the C chain to the address created in metamask
5) Use advanced options to specify a transaction with gas below the minimum gas price of 47 gwei

The node that it's issued to should probably reject the transaction outright ideally in a way that allows apps such as metamask to display the error appropriately to users. Instead, the node it's issued to gets to a block height one above the rest of the nodes and then completely stops processing consensus messages.

aaronbuchwald:
Fixes #16

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -243,6 +243,7 @@ func (vm *VM) Initialize(
 	config.Miner.GasPrice = minGasPrice
 	config.GPO.Default = minGasPrice
 	config.TxPool.PriceLimit = minGasPrice.Uint64()
+	config.TxPool.NoLocals = true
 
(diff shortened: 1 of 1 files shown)


tyler-smith:
This looks okay to me but it would be great if @Determinant could confirm.

PR #35, opened 2020-09-17
Fix panic on shutdown form orphaned goroutine

aaronbuchwald:


plugin/evm/vm.go
```diff
@@ -745,6 +723,52 @@ func (vm *VM) writeBackMetadata() {
 	atomic.StoreUint32(&vm.writingMetadata, 0)
 }
 
+// awaitTxPoolStabilized waits for a txPoolHead channel event
+// and notifies the VM when the tx pool has stabilized to the
+// expected block hash
+// Waits for signal to shutdown from txPoolStabilizedShutdownChan chan
+func (vm *VM) awaitTxPoolStabilized() {
+	vm.shutdownWg.Add(1)
```
StephenButtolph, line 731: This should be called in the main goroutine, not the spawned goroutine.

plugin/evm/vm.go
```diff
@@ -745,6 +723,52 @@ func (vm *VM) writeBackMetadata() {
...
+				vm.txPoolStabilizedLock.Lock()
+				if vm.txPoolStabilizedHead == h.Block.Hash() {
+					vm.txPoolStabilizedOk <- struct{}{}
+					vm.txPoolStabilizedHead = common.Hash{}
+				}
+				vm.txPoolStabilizedLock.Unlock()
+			default:
+			}
+		case <-vm.txPoolStabilizedShutdownChan:
+			vm.shutdownWg.Done()
```
StephenButtolph, line 746: Would it make sense for this to be in a defer at the start of the function?

plugin/evm/vm.go
```diff
@@ -745,6 +723,52 @@ func (vm *VM) writeBackMetadata() {
...
+			return
+		}
+	}
+}
+
+func (vm *VM) awaitSubmittedTxs() {
```
StephenButtolph, line 752: Same comments as `awaitTxPoolStabilized`

plugin/evm/vm.go
aaronbuchwald: Yup that would be much better

plugin/evm/vm.go
aaronbuchwald: Updated

StephenButtolph (approved):

PR #43, opened 2020-10-16
Handle block number requests

### LINKED ISSUE #28: debug_traceTransaction RPC command not available
Blockscout uses the `debug_traceTransaction` RPC call when indexing transactions. Without it the explorer is less useless, but it also seems to be causing Blockscout performance issues.

We need a way to enable this RPC, which belongs to the `debug` API group. 

RPC documentation: https://github.com/ethereum/go-ethereum/wiki/Management-APIs#debug_tracetransaction

aaronbuchwald:
Treat requests for Pending and Latest blocks as requests for the last Accepted block where possible
Adds JSON config to be passed in as command line arg to evm plugin
Config specifies RPC gas caps and which APIs to enable - this resolves #28 pending the corresponding change in avalanchego


Determinant (approved):
LGTM

PR #49, opened 2020-11-05
Fix db

aaronbuchwald:
Separate database into prefix dbs
Optimize get block status based on acceptedDB


aaronbuchwald:
requires database migration

PR #56, opened 2020-11-17
Mc transfer

aaronbuchwald:


dasconnor (commented):
Didn't go through everything, but I had some thoughts on the Solidity

params/protocol_params.go
```diff
@@ -95,6 +95,8 @@ const (
 	ExtcodeHashGasConstantinople uint64 = 400  // Cost of EXTCODEHASH (introduced in Constantinople)
 	ExtcodeHashGasEIP1884        uint64 = 700  // Cost of EXTCODEHASH after EIP 1884 (part in Istanbul)
 	SelfdestructGasEIP150        uint64 = 5000 // Cost of SELFDESTRUCT post EIP 150 (Tangerine)
+	AssetBalanceApricot          uint64 = 20
```
aaronbuchwald, line 98: Add comment explaining this

core/vm/contracts_stateful.go
```diff
@@ -0,0 +1,140 @@
...
+	} else if mcerr != 0 {
+		log.Info("incompatible account for mc transfer")
+		return nil, remainingGas, ErrIncompatibleAccount
+	}
+
+	snapshot := evm.StateDB.Snapshot()
+
+	if !evm.StateDB.Exist(to) {
+		log.Info("Creating account entry", "address", to.Hex())
+		remainingGas -= params.CallNewAccountGas
```
aaronbuchwald, line 119: Add overflow check

core/vm/contracts_stateful.go
```diff
@@ -0,0 +1,140 @@
...
+// TODO if value is passed into pre-compiled contract, then it can potentially be redirected from going to precompiled address
+// to going to the address specified by the call
+func (t *nativeAssetCall) Run(evm *EVM, caller ContractRef, input []byte, suppliedGas uint64, readOnly bool) (ret []byte, remainingGas uint64, err error) {
+	// input: encodePacked(address 20 bytes, assetID 32 bytes, assetAmount 32 bytes, callData variable length bytes)
+	if readOnly {
+		return nil, remainingGas, errors.New("cannot execute native asset transfer within read only call")
+	}
+
+	if len(input) < 84 {
+		return nil, remainingGas, fmt.Errorf("missing inputs to native asset transfer, expected input of length at least 92, but found %d", len(input))
```
aaronbuchwald, line 95: Correct error message

core/vm/evm.go
```diff
@@ -34,28 +36,31 @@ var emptyCodeHash = crypto.Keccak256Hash(nil)
...
-	case evm.chainRules.IsYoloV1:
-		precompiles = PrecompiledContractsYoloV1
-	case evm.chainRules.IsIstanbul:
-		precompiles = PrecompiledContractsIstanbul
-	case evm.chainRules.IsByzantium:
-		precompiles = PrecompiledContractsByzantium
-	default:
-		precompiles = PrecompiledContractsHomestead
-	}
+func (evm *EVM) precompile(addr common.Address) (StatefulPrecompiledContract, bool) {
```
StephenButtolph, line 51: Shouldn't we handle this the same way, by having our own hardfork identifier?

core/vm/evm.go
```diff
@@ -265,7 +271,7 @@ func (evm *EVM) Call(caller ContractRef, addr common.Address, input []byte, gas
 	return ret, gas, err
 }
 
-// This allows the user transfer balance of a specified coinId in addition to a normal Call().
+// CallExpert allows the user transfer balance of a specified coinId in addition to a normal Call().
```
StephenButtolph, line 274: This call gives me the spooks. Probably going to want to deprecate it past the hardfork, but we'll still need to support it for verifying the transitions pre-hardfork.

arc20.sol
```diff
@@ -0,0 +1,116 @@
...
+        balanceOf[msg.sender] += depositAmount;
+        emit Deposit(msg.sender, depositAmount);
+        return depositAmount;
+    }
+    
+    function withdraw(uint256 amount) public {
+        require(balanceOf[msg.sender] >= amount, "Insufficient funds for withdrawal");
+        
+        // use multicoin call here to transfer
+        balanceOf[msg.sender] -= amount;
```
StephenButtolph, line 67: Do we need to remove the `msg.sender` entry if it's value goes to zero? or does the EVM already handle that auto-magically?

arc20.sol
```diff
@@ -0,0 +1,116 @@
...
+        return data;
+    }
+    
+    function contractBalance() internal returns (uint256) {
+        return getBalance(address(this), assetID);
+    }
+    
+    function updateContractBalance() public {
+        _totalSupply = contractBalance();
+    }
```
StephenButtolph, line 48: This should be removed.

arc20.sol
```diff
@@ -0,0 +1,116 @@
...
+    
+    function assetCall(address addr, uint256 aID, uint256 assetAmount, bytes memory callData) public returns (bytes memory) {
+        (bool success, bytes memory data) = transferAddr.call(abi.encodePacked(addr, aID, assetAmount, callData));
+        require(success, "assetCall failed");
+        return data;
+    }
+    
+    function contractBalance() internal returns (uint256) {
+        return getBalance(address(this), assetID);
+    }
```
StephenButtolph, line 44: Lets try to be more clear on the term `balance` for if we are talking about the erc20 balance or the underlying ANT balance.

arc20.sol
```diff
@@ -0,0 +1,116 @@
...
+    constructor(string memory _name, string memory _symbol, uint8 _decimals, uint256 _assetID) public {
+        name = _name;
+        symbol = _symbol;
+        decimals = _decimals;
+        assetID = _assetID;
+    }
+    
+    function getBalance(address addr, uint256 aID) public returns (uint256) {
+        (bool success, bytes memory data) = balanceAddr.call(abi.encodePacked(addr, aID));
+        require(success, "getBalance failed");
```
StephenButtolph, line 32: This should be an assert.

arc20.sol
```diff
@@ -0,0 +1,116 @@
...
+    function updateContractBalance() public {
+        _totalSupply = contractBalance();
+    }
+    
+    event Deposit(address indexed depositor, uint256 value);
+    event Withdrawal(address indexed withdrawer, uint256 value);
+    
+    function deposit() public returns (uint256) {
+        uint256 balance = contractBalance();
+        require(balance >= _totalSupply, "Balance was less than total supply. Critical error occurred.");
```
StephenButtolph, line 55: This should be an assert.

arc20.sol
```diff
@@ -0,0 +1,116 @@
...
+
+    event Approval(address indexed owner, address indexed spender, uint256 value);
+
+    mapping(address => mapping(address => uint256)) public allowance;
+
+    function approve(address spender, uint256 value)
+        public
+        returns (bool success)
+    {
+        allowance[msg.sender][spender] = value;
```
StephenButtolph, line 98: This seems incredibly unsafe, but apparently that's the erc20 interface... so I guess it's fine.

arc20.sol
```diff
@@ -0,0 +1,116 @@
...
+        // use multicoin call here to transfer
+        balanceOf[msg.sender] -= amount;
+        _totalSupply -= amount;
+        // Transfer multicoin asset to sender
+        assetCall(msg.sender, assetID, amount, "");
+        // TODO remove this check, must allow balances to be different otherwise
+        // someone sending funds without calling deposit could cause withdraw to
+        // malfunction
+        uint256 newBalance = contractBalance();
+        require(_totalSupply == newBalance, "Total supply does not equal contract balance");
```
StephenButtolph, line 75: +1 for removing.

core/vm/contracts_stateful.go
aaronbuchwald: Updated

arc20.sol
```diff
@@ -0,0 +1,116 @@
...
+    uint256 public assetID;
+    string public name;
+    string public symbol;
+    uint8 public decimals;
+
+    uint256 public _totalSupply;
+    
+    constructor(string memory _name, string memory _symbol, uint8 _decimals, uint256 _assetID) public {
+        name = _name;
+        symbol = _symbol;
```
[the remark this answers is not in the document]
aaronbuchwald: This is really just an example, so I think it's best to leave this out.

arc20.sol
```diff
@@ -0,0 +1,116 @@
...
+        decimals = _decimals;
+        assetID = _assetID;
+    }
+    
+    function getBalance(address addr, uint256 aID) public returns (uint256) {
```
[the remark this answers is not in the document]
aaronbuchwald: getBalance calls the precompiled contract to get the balance of an assetID owned by an address. This has nothing to do with the balance of the erc20 and was just included for convenience of testing. I removed it in the updated version.

arc20.sol
```diff
@@ -0,0 +1,116 @@
...
+    
+    // Track how many tokens are owned by each address.
+    mapping (address => uint256) public balanceOf;
+
+    uint256 public assetID;
+    string public name;
+    string public symbol;
+    uint8 public decimals;
+
+    uint256 public _totalSupply;
```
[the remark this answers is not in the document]
aaronbuchwald: Updated.

arc20.sol
```diff
@@ -0,0 +1,116 @@
...
+    
+    function updateContractBalance() public {
+        _totalSupply = contractBalance();
+    }
+    
+    event Deposit(address indexed depositor, uint256 value);
+    event Withdrawal(address indexed withdrawer, uint256 value);
+    
+    function deposit() public returns (uint256) {
+        uint256 balance = contractBalance();
```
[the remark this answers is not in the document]
aaronbuchwald: With WETH, when deposit is called it can simply read msg.value to figure out the deposit amount from the sender. There's no equivalent with other assets. So the solution here is to have the contract keep track of its current balance of the asset, then when someone wants to deposit, they send the funds and call deposit. The contract can then calculate the difference between what it previously though its balance was and the updated amount and use that as the deposit amount.

arc20.sol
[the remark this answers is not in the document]
aaronbuchwald: Response above.

arc20.sol
```diff
@@ -0,0 +1,116 @@
...
+    }
+    
+    function updateContractBalance() public {
+        _totalSupply = contractBalance();
+    }
+    
+    event Deposit(address indexed depositor, uint256 value);
+    event Withdrawal(address indexed withdrawer, uint256 value);
+    
+    function deposit() public returns (uint256) {
```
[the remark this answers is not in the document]
aaronbuchwald: Response above.

core/vm/contracts_stateful.go
```diff
@@ -0,0 +1,140 @@
...
+	common.BytesToAddress([]byte{2}):  newWrappedPrecompiledContract(&sha256hash{}),
+	common.BytesToAddress([]byte{3}):  newWrappedPrecompiledContract(&ripemd160hash{}),
+	common.BytesToAddress([]byte{4}):  newWrappedPrecompiledContract(&dataCopy{}),
+	common.BytesToAddress([]byte{5}):  newWrappedPrecompiledContract(&bigModExp{}),
+	common.BytesToAddress([]byte{6}):  newWrappedPrecompiledContract(&bn256AddIstanbul{}),
+	common.BytesToAddress([]byte{7}):  newWrappedPrecompiledContract(&bn256ScalarMulIstanbul{}),
+	common.BytesToAddress([]byte{8}):  newWrappedPrecompiledContract(&bn256PairingIstanbul{}),
+	common.BytesToAddress([]byte{9}):  newWrappedPrecompiledContract(&blake2F{}),
+	common.BytesToAddress([]byte{10}): &nativeAssetBalance{gasCost: params.AssetBalanceApricot},
+	common.BytesToAddress([]byte{11}): &nativeAssetCall{gasCost: params.AssetCallApricot},
```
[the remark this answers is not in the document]
aaronbuchwald: Updated to start these at the opposite end of the spectrum.

core/vm/evm.go
aaronbuchwald: Yup, this has been update for both the precompiled contracts and the instruction sets in order to deprecate callEx and balancemc.

core/vm/evm.go
aaronbuchwald: Updated to be deprecated after the hard fork.

arc20.sol
aaronbuchwald: I believe this is handled auto-magically since I can't seem to find any example erc20s that use delete instead of just subtracting it.

arc20.sol
aaronbuchwald: Removed.

arc20.sol
aaronbuchwald: Updated.

Issue #46, opened 2020-11-03
Block.Hash produces a wrong hash


sosedoff:
Running `block.Hash()` produces a wrong hash when comparing to the actual hash from the RPC response.
Filing the ticket per @aaronbuchwald request.

Code to reproduce the issue: 

```go
package main

import (
	"context"
	"log"

	"github.com/ethereum/go-ethereum/ethclient"
)

func main() {
	client, err := ethclient.Dial("https://api.avax-test.network/ext/bc/C/rpc")
	if err != nil {
		log.Fatal(err)
	}

	header, err := client.HeaderByNumber(context.Background(), nil)
	if err != nil {
		log.Fatal(err)
	}

	log.Println("block number:", header.Number)
	log.Println("block hash:", header.Hash().Hex())

        // This will fail as `header.Hash()` returns an invalid hash value
	header2, err := client.HeaderByHash(context.Background(), header.Hash())
	if err != nil {
		log.Fatal(err)
	}

	log.Println("fetched number:", header2.Number)
	log.Println("fetched hash:", header2.Hash().Hex())
}
```

aaronbuchwald:
Closing this issue since it has been resolved by using ethclient. The issue was a result of changes to the Eth Block format to include an extra data field and a version field, to support atomic transactions. This resulted in the hash calculated by the node and the hash calculated by client libraries that ignored this field producing a different hash from the block.

PR #58, opened 2020-11-19
Ethclient

aaronbuchwald:
Provides solution for #46 . block.hash is producing an incorrect hash because some of the data is not present in the native Ethereum version of a block. Therefore, when it takes a hash of the block, it's missing part of the data and creating a different hash. Using this version of ethclient should resolve the issue.


go.mod
```diff
@@ -4,7 +4,9 @@ go 1.14
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.5.7
+	github.com/ava-labs/avalanche-testing v0.9.6
 	github.com/ava-labs/avalanchego v1.0.4-update-id-2
+	github.com/ava-labs/gecko v0.6.1-rc.1
```
tasinco, line 9: this should be avalanchego 1.0.5+ ?

aaronbuchwald: Yes, good catch. Idk why this gets overrided sometimes when building.

go.mod
```diff
@@ -23,6 +25,7 @@ require (
 	github.com/pborman/uuid v1.2.0
 	github.com/prometheus/tsdb v0.10.0
 	github.com/rjeczalik/notify v0.9.2
+	github.com/sirupsen/logrus v1.6.0
```
aaronbuchwald, line 28: Need to remove this.

ethclient/ethclient.go
```diff
@@ -0,0 +1,575 @@
...
+	var head *types.Header
+	var body rpcBlock
+	if err := json.Unmarshal(raw, &head); err != nil {
+		return nil, err
+	}
+	if err := json.Unmarshal(raw, &body); err != nil {
+		return nil, err
+	}
+	// Quick-verify transaction and uncle lists. This mostly helps with debugging the server.
+	if head.UncleHash == types.EmptyUncleHash && len(body.UncleHashes) > 0 {
```
aaronbuchwald, line 125: Remove uncle related calls

ethclient/ethclient.go
```diff
@@ -0,0 +1,575 @@
...
+	StartingBlock hexutil.Uint64
+	CurrentBlock  hexutil.Uint64
+	HighestBlock  hexutil.Uint64
+	PulledStates  hexutil.Uint64
+	KnownStates   hexutil.Uint64
+}
+
+// SyncProgress retrieves the current progress of the sync algorithm. If there's
+// no sync currently running, it returns nil.
+func (ec *Client) SyncProgress(ctx context.Context) (*coreth.SyncProgress, error) {
```
aaronbuchwald, line 327: Remove if not implemented in coreth

PR #60, opened 2020-11-22
block extra data parse fix

tasinco:
golang json.Unmarshal appears to only decode []byte from base64 strings..
1. Parse it to a string.
2. Hex decode the string.

### DIFF
--- a/ethclient/ethclient.go
+++ b/ethclient/ethclient.go
@@ -97,11 +97,11 @@ func (ec *Client) BlockNumber(ctx context.Context) (uint64, error) {
 }
 
 type rpcBlock struct {
-	Hash         common.Hash      `json:"hash"`
-	Transactions []rpcTransaction `json:"transactions"`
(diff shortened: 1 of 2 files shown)


ethclient/ethclient.go
```diff
@@ -121,6 +121,16 @@ func (ec *Client) getBlock(ctx context.Context, method string, args ...interface
 	if err := json.Unmarshal(raw, &body); err != nil {
 		return nil, err
 	}
+
+	var blockExtraData *[]byte
+	if len(body.BlockExtraData) != 0 {
+		blockExtraDataDecoded, err := hexutil.Decode(body.BlockExtraData)
+		if err != nil {
+			return nil, err
```
aaronbuchwald, line 129: Could you change this to `fmt.Errorf("problem parsing block extra data: %w", err)` so the error is clearer if there's ever a problem?

aaronbuchwald (approved):

PR #62, opened 2020-11-25
Dev

aaronbuchwald:


coreth.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/event"
 	"github.com/ethereum/go-ethereum/trie"
+
 	//"github.com/ethereum/go-ethereum/event"
```
StephenButtolph, line 23: Why do we have commented out imports?

ethclient/ethclient.go
```diff
@@ -0,0 +1,586 @@
...
+// 		return nil, err
+// 	}
+// 	return &coreth.SyncProgress{
+// 		StartingBlock: uint64(progress.StartingBlock),
+// 		CurrentBlock:  uint64(progress.CurrentBlock),
+// 		HighestBlock:  uint64(progress.HighestBlock),
+// 		PulledStates:  uint64(progress.PulledStates),
+// 		KnownStates:   uint64(progress.KnownStates),
+// 	}, nil
+// }
```
StephenButtolph, line 359: Generally speaking, we should avoid committing commented out code. Is there a reason we have some here? (and in the rest of this file?)

plugin/evm/static_service.go
```diff
@@ -8,15 +8,14 @@ import (
...
-	"github.com/ava-labs/avalanchego/utils/formatting"
 )
 
 // StaticService defines the static API services exposed by the evm
 type StaticService struct{}
 
 // BuildGenesis returns the UTXOs such that at least one address in [args.Addresses] is
 // referenced in the UTXO.
-func (*StaticService) BuildGenesis(_ context.Context, args *core.Genesis) (formatting.CB58, error) {
+func (*StaticService) BuildGenesis(_ context.Context, args *core.Genesis) ([]byte, error) {
```
StephenButtolph, line 18: Should this return `{bytes, encoding}`?

plugin/evm/vm.go
```diff
@@ -110,26 +111,29 @@ func maxDuration(x, y time.Duration) time.Duration {
...
+		c.RegisterType(&secp256k1fx.TransferInput{}),
+		c.RegisterType(&secp256k1fx.MintOutput{}),
+		c.RegisterType(&secp256k1fx.TransferOutput{}),
+		c.RegisterType(&secp256k1fx.MintOperation{}),
+		c.RegisterType(&secp256k1fx.Credential{}),
+		c.RegisterType(&secp256k1fx.Input{}),
+		c.RegisterType(&secp256k1fx.OutputOwners{}),
 	)
+
+	errs.Add(Codec.RegisterCodec(0, c))
```
StephenButtolph, line 136: This can just be included in the previous `Add` chain. Also we should use `codecVersion` here rather than `0`.

node/node.go
```diff
@@ -113,8 +113,8 @@ func New(conf *Config) (*Node, error) {
 		eventmux:      new(event.TypeMux),
 		log:           conf.Logger,
 		stop:          make(chan struct{}),
-		server:        &p2p.Server{Config: conf.P2P},
-		databases:     make(map[*closeTrackingDB]struct{}),
+		// server:        &p2p.Server{Config: conf.P2P},
```
StephenButtolph, line 116: Should we just remove this? or should we keep it for posterity? (same below)

ethclient/ethclient.go
```diff
@@ -0,0 +1,586 @@
+// Copyright 2016 The go-ethereum Authors
```
StephenButtolph, line 1: Honestly, I'm not sure exactly what this file is doing. But it seems to mainly be a drop in from geth.

ethclient/ethclient.go
aaronbuchwald: Since this is taken from geth, I find it much easier to know what changes have been made by simply commenting out code that we don't need anymore as opposed to looking through the git history or pulling up the geth codebase to compare and see what's changed.

ethclient/ethclient.go
aaronbuchwald: Yup this is just a drop in from geth. Since there were a few small changes to the block format in coreth, using geth's ethclient will not handle the extra fields, so we need to have our own version both to correctly parse the extra data and also in order to make sure that the client will produce the correct hashes of the blocks that it does retrieve. Using this version of ethclient fixes #46

plugin/evm/vm.go
aaronbuchwald: Updated

node/node.go
aaronbuchwald: I'm in favor of commenting out as much of the stuff we aren't actually using as possible. I think for a future update, having the code commented out instead of deleted will make upgrading to a newer version of geth much easier.

coreth.go
aaronbuchwald: No idea. Going to remove the commented out imports.

plugin/evm/static_service.go
aaronbuchwald: Updated to return a BuildGenesisReply struct.

Determinant (approved):
LGTM.

PR #72, opened 2020-12-15
Dev

aaronbuchwald:


.ci/lint.sh
```diff
@@ -0,0 +1,12 @@
...
+# or install it into ./bin/
+curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.30.0
+# In alpine linux (as it does not come with curl by default)
+wget -O- -nv https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.30.0
+
+export PATH=$PATH:$(go env GOPATH)/bin
+
+golangci-lint --version
+
+golangci-lint run --max-same-issues 0
```
danlaine, line 12: Can we add a newline here so Github doesn't complain?

.golangci.yml
```diff
@@ -0,0 +1,50 @@
...
+  disable-all: true
+  enable:
+    - deadcode
+    - goconst
+    - goimports
+    - gosimple
+    - govet
+    - ineffassign
+    - misspell
+    # - staticcheck
```
danlaine, line 22: Should we remove this one since it's commented out? Same for `unused`

.ci/lint.sh
```diff
@@ -0,0 +1,12 @@
+# binary will be $(go env GOPATH)/bin/golangci-lint
```
danlaine, line 1: Am I correct in thinking that this will try to install `golangci-lint` three times? If so, not a huge deal, but wouldn't it be better to try to run `golangci-lint`, then if that fails because it's not installed, try to install the first way and try to run, then if that doesn't work, try the second way, and so on?

eth/backend.go
```diff
@@ -498,13 +507,13 @@ func (s *Ethereum) Start() error {
...
 
 	// Figure out a max peers count based on the server limits
-	maxPeers := s.p2pServer.MaxPeers
-	if s.config.LightServ > 0 {
-		if s.config.LightPeers >= s.p2pServer.MaxPeers {
-			return fmt.Errorf("invalid peer config: light peer count (%d) >= total peer count (%d)", s.config.LightPeers, s.p2pServer.MaxPeers)
-		}
-		maxPeers -= s.config.LightPeers
-	}
+	// maxPeers := s.p2pServer.MaxPeers
```
danlaine, line 510: Should we remove this commented out code?

ethclient/ethclient.go
```diff
@@ -324,13 +324,13 @@ func toBlockNumArg(number *big.Int) string {
...
 }
 
-type rpcProgress struct {
-	StartingBlock hexutil.Uint64
-	CurrentBlock  hexutil.Uint64
-	HighestBlock  hexutil.Uint64
-	PulledStates  hexutil.Uint64
-	KnownStates   hexutil.Uint64
-}
+// type rpcProgress struct {
```
danlaine, line 327: Same question here about removing commented out code

miner/worker.go
```diff
@@ -913,12 +913,15 @@ func (w *worker) commitNewWork(interrupt *int32, noempty bool, timestamp int64)
 		//timestamp = int64(parent.Time() + 1)
 		timestamp = int64(parent.Time())
 	}
-	// this will ensure we're not going off too far in the future
-	if now := time.Now().Unix(); timestamp > now+1 {
-		wait := time.Duration(timestamp-now) * time.Second
-		log.Info("Mining too far in the future", "wait", common.PrettyDuration(wait))
-		time.Sleep(wait)
-	}
+	// This block has been removed to prevent BuildBlock
```
danlaine, line 916: Same question here about removing commented out code

plugin/evm/vm.go
```diff
@@ -750,7 +761,7 @@ func (vm *VM) awaitTxPoolStabilized() {
 				vm.txPoolStabilizedLock.Unlock()
 			default:
 			}
-		case <-vm.txPoolStabilizedShutdownChan:
+		case <-vm.shutdownChan:
```
danlaine, line 764: Comment about should be updated to:
```
// Waits for signal to shutdown from vm.shutdownChan
```

.ci/lint.sh
aaronbuchwald: Added

.ci/lint.sh
aaronbuchwald: I just copied this script from avalanchego, but I'll check with Fabio.

plugin/evm/vm.go
aaronbuchwald: Good catch, just updated

plugin/evm/import_tx_test.go
```diff
@@ -178,7 +178,7 @@ func TestImportTxSemanticVerify(t *testing.T) {
 		Outs: []EVMOutput{evmOutput},
 	}
 
-	state, err := vm.chain.BlockState(vm.lastAccepted.ethBlock)
+	state, err := vm.chain.CurrentState()
```
Determinant, line 181: Are we sure this is the desired behavior? Unlike geth, we don't have time-consuming PoW, which means the "current" block may not be the block interested (accepted). Need to double check what CurrentState returns.

plugin/evm/vm.go
```diff
@@ -145,45 +152,62 @@ type VM struct {
...
+	// [mayBuildBlock] == [build] means that the next block may be built
+	// at any time.
+	mayBuildBlock mayBuildBlockStatus
+	// If true, try to notify the engine that a block should be built.
+	// Engine may not be notified because [mayBuildBlock] says to wait.
+	tryToBuildBlock bool
+	// If true, the engine has been notified that it should build a block
+	// but has not done so yet. If this is the case, wait until it has
+	// built a block before notifying it again.
+	awaitingBuildBlock bool
```
Determinant, line 197: This part looks much clearer than before. :)

eth/gasprice/gasprice.go
```diff
@@ -29,8 +29,6 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 )
 
-const sampleNumber = 3 // Number of transactions sampled in a block
```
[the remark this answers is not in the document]
aaronbuchwald: This wasn't used by anything. It was caught by the linter. I believe it was used in the old suggest gas price logic, which has been replaced to return the minimum gas price.

miner/worker.go
```diff
@@ -913,12 +913,15 @@ func (w *worker) commitNewWork(interrupt *int32, noempty bool, timestamp int64)
 		//timestamp = int64(parent.Time() + 1)
 		timestamp = int64(parent.Time())
 	}
-	// this will ensure we're not going off too far in the future
-	if now := time.Now().Unix(); timestamp > now+1 {
-		wait := time.Duration(timestamp-now) * time.Second
-		log.Info("Mining too far in the future", "wait", common.PrettyDuration(wait))
-		time.Sleep(wait)
-	}
```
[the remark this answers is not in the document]
aaronbuchwald: Yes. If a block is added as a future block, `blk.Verify()` will error.

PR #86, opened 2021-01-12
Fix get filter logs

aaronbuchwald:


StephenButtolph (approved):
This needs to be re-looked into for performance reasons... But as a patch this lgtm

PR #95, opened 2021-02-04
Add regression test for set preference race

aaronbuchwald:
Adds a regression test that catches the case where the consensus engine calls SetPreference on a block that may not become finalized resulting in the corruption of eth's notion of the canonical chain. Note: this is different from the canonical chain maintained by the VM, so it results in invalid responses from the API, but does not impact anything on-chain.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -900,6 +900,66 @@ func (bc *BlockChain) ValidateCanonicalChain() error {
 	return nil
 }
 
+// WriteCanonicalFromCurrentBlock writes the canonical chain from the
+// current block to the genesis.
+// WriteCanonicalFromCurrentBlock only grabs the lock as necessary to
+// write batches to disk, such that it can be run in a goroutine.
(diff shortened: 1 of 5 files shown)


StephenButtolph (approved):

core/blockchain.go
```diff
@@ -857,6 +857,109 @@ func (bc *BlockChain) GetBlockByNumber(number uint64) *types.Block {
...
+}
+
+// WriteCanonicalFromCurrentBlock writes the canonical chain from the
+// current block to the genesis.
+// WriteCanonicalFromCurrentBlock only grabs the lock as necessary to
+// write batches to disk, such that it can be run in a goroutine.
+func (bc *BlockChain) WriteCanonicalFromCurrentBlock() error {
+	current := bc.CurrentBlock()
+
+	batchSize := 10
```
patrick-ogrady, line 910: can we make this a `const`?

aaronbuchwald: Good idea

PR #109, opened 2021-02-09
Update block timestamp for mock call with pending block number

aaronbuchwald:


dasconnor (commented):
We might want to take the opportunity to increment the block number here too, but otherwise code looks good

PR #114, opened 2021-02-12
Dev

aaronbuchwald:


coreth.go
```diff
@@ -62,7 +63,7 @@ func NewETHChain(config *eth.Config, nodecfg *node.Config, etherBase *common.Add
 	cb := new(dummy.ConsensusCallbacks)
 	mcb := new(miner.MinerCallbacks)
 	bcb := new(eth.BackendCallbacks)
-	backend, _ := eth.New(node, config, cb, mcb, bcb, chainDB)
+	backend, _ := eth.New(node, config, cb, mcb, bcb, chainDB, settings)
```
danlaine, line 66: Perhaps not something to address at this very moment but should we ignore this error here?

plugin/evm/block.go
```diff
@@ -80,75 +159,107 @@ func (b *Block) Height() uint64 {
...
 	if tx != nil {
-		pState, err := b.vm.chain.BlockState(b.Parent().(*Block).ethBlock)
+		// If the ancestor is unknown, then the parent failed verification when
+		// it was called.
+		// If the ancestor is rejected, then this block shouldn't be inserted
+		// into the canonical chain because the parent is will be missing.
+		if blkStatus := ancestorIntf.Status(); blkStatus == choices.Unknown || blkStatus == choices.Rejected {
+			return errRejectedParent
+		}
+		ancestor := ancestorIntf.(*Block)
```
danlaine, line 199: Can we do a type check here to avoid a potential panic?

```
ancestor, ok := ancestorIntf.(*Block)
if !ok {
    return fmt.Errorf("expected %s, parent of %s, to be *Block but is %T", ancestor.ID(), b.ID(), ancestorIntf)
}
```

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.14
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.5.7
-	github.com/ava-labs/avalanchego v1.1.2-0.20210107163848-9e1d2963d429
+	github.com/ava-labs/avalanchego v1.1.6-0.20210211024119-1c20967d1d73
```
patrick-ogrady, line 7: Let's make this rely on a merged commit.

plugin/evm/block.go
```diff
@@ -80,75 +159,107 @@ func (b *Block) Height() uint64 {
...
+					// verification when it was called.
+					// If the ancestor is rejected, then this block shouldn't be
+					// inserted into the canonical chain because the parent is
+					// will be missing.
+					// If the ancestor is processing, then the block may have
+					// been verified.
+					if blkStatus := ancestorIntf.Status(); blkStatus == choices.Unknown || blkStatus == choices.Rejected {
+						return errRejectedParent
+					}
+					ancestor = ancestorIntf.(*Block)
```
danlaine, line 237: Same here about type assertion

StephenButtolph (approved):

PR #115, opened 2021-02-15
Db upgrade

aaronbuchwald:
This PR adds ChainState to provide an efficient caching layer for blocks that sits on top of the VM. This PR requires a database upgrade.

### DIFF
--- a/coreth.go
+++ b/coreth.go
@@ -41,7 +41,6 @@ type ETHChain struct {
 	backend *eth.Ethereum
 	cb      *dummy.ConsensusCallbacks
 	mcb     *miner.MinerCallbacks
-	bcb     *eth.BackendCallbacks
 }
 
 // NewETHChain creates an Ethereum blockchain with the given configs.
@@ -62,9 +61,8 @@ func NewETHChain(config *eth.Config, nodecfg *node.Config, etherBase *common.Add
 	//}
 	cb := new(dummy.ConsensusCallbacks)
 	mcb := new(miner.MinerCallbacks)
-	bcb := new(eth.BackendCallbacks)
-	backend, _ := eth.New(node, config, cb, mcb, bcb, chainDB, settings)
-	chain := &ETHChain{backend: backend, cb: cb, mcb: mcb, bcb: bcb}
+	backend, _ := eth.New(node, config, cb, mcb, chainDB, settings)
+	chain := &ETHChain{backend: backend, cb: cb, mcb: mcb}
 	if etherBase == nil {
 		etherBase = &BlackholeAddr
 	}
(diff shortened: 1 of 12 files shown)


plugin/evm/state/state.go
```diff
@@ -0,0 +1,370 @@
...
+		}
+		if err := c.acceptedBlocksByHeight.Put(heightKey(genesisBlock.Height()), genesisID[:]); err != nil {
+			return err
+		}
+		lastAcceptedBlock = &BlockWrapper{
+			Block:  genesisBlock,
+			status: choices.Accepted,
+		}
+	}
+	lastAcceptedBlock.state = c
```
StephenButtolph, line 140: Can we populate this field when we construct the `BlockWrapper`s?

plugin/evm/state/state.go
```diff
@@ -0,0 +1,370 @@
...
+		return blk, nil
+	}
+
+	if _, ok := c.missingBlocks.Get(blkID); ok {
+		return nil, errUnknownBlock
+	}
+
+	c.lock.Unlock()
+	blk, err := c.getBlock(blkID)
+	c.lock.Lock()
```
StephenButtolph, line 176: this looks really racy.

plugin/evm/state/state.go
```diff
@@ -0,0 +1,370 @@
...
+		return nil, err
+	}
+
+	return wrappedBlk.(*BlockWrapper).Block, nil
+}
+
+// getBlockIDAtHeight returns the blkID accepted at [height] if there is
+// one.
+// assumes the lock is held.
+func (c *ChainState) getBlockIDAtHeight(height uint64) (ids.ID, error) {
```
StephenButtolph, line 218: Why does this exist? Shouldn't this just get Geth's canonical hash?

plugin/evm/state/state_test.go
```diff
@@ -0,0 +1,439 @@
...
+	"github.com/ava-labs/avalanchego/database/memdb"
+	"github.com/ava-labs/avalanchego/database/prefixdb"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/snow/choices"
+	"github.com/ava-labs/avalanchego/snow/consensus/snowman"
+	"github.com/ava-labs/avalanchego/utils/hashing"
+	"github.com/stretchr/testify/assert"
+)
+
+type testBlock struct {
```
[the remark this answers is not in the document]
aaronbuchwald: Yes, just wrapped it to provide a SetStatus function as well.

plugin/evm/state/state.go
aaronbuchwald: Going to remove in favor of geth's canonical hash.

plugin/evm/state/state.go
aaronbuchwald: Removing all of the locks.

StephenButtolph (approved):
I think it needs to have a `go mod tidy` run on it, but aside from that, LGTM

coreth.go
```diff
@@ -62,9 +61,8 @@ func NewETHChain(config *eth.Config, nodecfg *node.Config, etherBase *common.Add
 	//}
 	cb := new(dummy.ConsensusCallbacks)
 	mcb := new(miner.MinerCallbacks)
-	bcb := new(eth.BackendCallbacks)
-	backend, _ := eth.New(node, config, cb, mcb, bcb, chainDB, settings)
-	chain := &ETHChain{backend: backend, cb: cb, mcb: mcb, bcb: bcb}
+	backend, _ := eth.New(node, config, cb, mcb, chainDB, settings)
```
patrick-ogrady, line 64: Why don't we check this error?

aaronbuchwald: Never has been before, but will update now to panic if there's an error here.

PR #116, opened 2021-02-16
Add  atomic tx mempool

aaronbuchwald:
This PR adds a mempool for atomic transactions. Since building a block may error at multiple stages and invokes multiple callbacks through the dummy consensus engine, the VM needs to signal to the mempool when a potential atomic tx has been issued into a block, fails verification, and whether or not block building was successful.

This PR also updates the getTx and getTxStatus endpoints to return transactions if they are in the mempool and adds a new Status type to include transactions that were "Dropped" from the mempool due to failing verification.

### DIFF
--- a/plugin/evm/mempool.go
+++ b/plugin/evm/mempool.go
@@ -1,0 +1,210 @@
+// (c) 2019-2020, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package evm
+
+import (
+	"sync"
+
+	"github.com/ava-labs/avalanchego/cache"
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+// Mempool is a simple mempool for atomic transactions
+type Mempool struct {
+	lock sync.Mutex
+	// maxSize is the maximum number of transactions allowed to be kept in mempool
+	maxSize int
+	// currentTx is the transaction about to be added to a block.
+	currentTx *Tx
+	// txs is the set of transactions that need to be issued into new blocks
+	txs map[ids.ID]*Tx
+	// issuedTxs is the set of transactions that have been issued into a new block
+	issuedTxs map[ids.ID]*Tx
+	// discardedTxs is an LRU Cache of recently discarded transactions
+	discardedTxs *cache.LRU
+	// Pending is a channel that signals when the mempool is ready to put an atomic transaction
+	// into a new block
+	Pending chan struct{}
+}
+
+// NewMempool returns a Mempool with [maxSize]
+func NewMempool(maxSize int) *Mempool {
+	return &Mempool{
+		txs:          make(map[ids.ID]*Tx),
+		issuedTxs:    make(map[ids.ID]*Tx),
+		discardedTxs: &cache.LRU{Size: 10},
+		Pending:      make(chan struct{}, 1),
+		maxSize:      maxSize,
+	}
+}
+
+// Len returns the number of transactions in the mempool
+func (m *Mempool) Len() int {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	return m.length()
+}
+
+// assumes the lock is held
+func (m *Mempool) length() int {
+	return len(m.txs) + len(m.issuedTxs)
+}
+
+// Add attempts to add [tx] to the mempool and returns true if the
+// mempool requires a new block to be built.
+func (m *Mempool) AddTx(tx *Tx) error {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	txID := tx.ID()
+	// If [txID] has already been issued, there is no need to
+	// build a new block.
+	if _, exists := m.issuedTxs[txID]; exists {
+		return nil
+	}
+
+	// Each of the following cases results in there being a
+	// pending transaction, so we add to the channel here.
+	m.addPending()
+
+	if m.length() >= m.maxSize {
+		return errTooManyAtomicTx
+	}
+
+	if _, exists := m.txs[txID]; exists {
+		return nil
+	}
+
+	m.txs[txID] = tx
+	return nil
+}
+
+// NextTx returns a transaction to be issued from the mempool.
+func (m *Mempool) NextTx() (*Tx, bool) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	for txID, tx := range m.txs {
+		delete(m.txs, txID)
+		m.currentTx = tx
+		// If there are still atomic transactions remaining
+		// ensure there is an item on the pending channel.
+		if len(m.txs) > 0 {
+			m.addPending()
+		}
+		return tx, true
+	}
+
+	return nil, false
+}
+
+// GetTx returns the transaction [txID] if it was issued
+// by this node and returns whether it was dropped and whether
+// it exists.
+func (m *Mempool) GetTx(txID ids.ID) (*Tx, bool, bool) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	if tx, ok := m.txs[txID]; ok {
+		return tx, false, true
+	}
+	if tx, ok := m.issuedTxs[txID]; ok {
+		return tx, false, true
+	}
+	if m.currentTx != nil && m.currentTx.ID() == txID {
+		return m.currentTx, false, true
+	}
+	if tx, exists := m.discardedTxs.Get(txID); exists {
+		return tx.(*Tx), true, true
+	}
+
+	return nil, false, false
+}
+
+// IssueCurrentTx marks [currentTx] as issued if there is one
+func (m *Mempool) IssueCurrentTx() {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	if m.currentTx != nil {
+		m.issuedTxs[m.currentTx.ID()] = m.currentTx
+		m.currentTx = nil
+	}
+
+	if len(m.txs) > 0 {
+		m.addPending()
+	}
+}
+
+// CancelCurrentTx marks the attempt to issue [currentTx]
+// as being aborted. If this is called after a buildBlock error
+// caused by the atomic transaction, then DiscardCurrentTx should have been called
+// such that this call will have no effect and should not re-issue the invalid tx.
+func (m *Mempool) CancelCurrentTx() {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	if m.currentTx != nil {
+		m.txs[m.currentTx.ID()] = m.currentTx
+		m.currentTx = nil
+	}
+
+	// If the VM failed to build a block for any reason
+	// make sure there is an item on the Pending channel if
+	// there are any more transactions to issue.
+	if len(m.txs) > 0 {
+		m.addPending()
+	}
+}
+
+// DiscardCurrentTx marks [currentTx] as invalid and aborts the attempt
+// to issue it since it failed verification.
+func (m *Mempool) DiscardCurrentTx() {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	if m.currentTx == nil {
+		return
+	}
+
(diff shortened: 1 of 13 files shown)


plugin/evm/mempool.go
```diff
@@ -0,0 +1,210 @@
...
+	txID := tx.ID()
+	// If [txID] has already been issued, there is no need to
+	// build a new block.
+	if _, exists := m.issuedTxs[txID]; exists {
+		return nil
+	}
+
+	// Each of the following cases results in there being a
+	// pending transaction, so we add to the channel here.
+	m.addPending()
```
patrick-ogrady, line 70: Would pending not already have been sent in the following places where we return early? Why send it again?

plugin/evm/status.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// UnmarshalJSON ...
+func (s *Status) UnmarshalJSON(b []byte) error {
+	str := string(b)
+	if str == "null" {
+		return nil
+	}
+	switch str {
+	case "\"Unknown\"":
```
patrick-ogrady, line 44: nit: `` `"Unknown"` `` ... may also want to make a const

plugin/evm/status.go
```diff
@@ -0,0 +1,81 @@
...
+	Processing
+	Accepted
+)
+
+// MarshalJSON ...
+func (s Status) MarshalJSON() ([]byte, error) {
+	if err := s.Valid(); err != nil {
+		return nil, err
+	}
+	return []byte("\"" + s.String() + "\""), nil
```
patrick-ogrady, line 34: nit: `` fmt.Sprintf(`"%s"`) ``

plugin/evm/vm.go
```diff
@@ -301,21 +299,29 @@ func (vm *VM) Initialize(
...
 	})
 	chain.SetOnFinalizeAndAssemble(func(state *state.StateDB, txs []*types.Transaction) ([]byte, error) {
-		select {
-		case atx := <-vm.pendingAtomicTxs:
-			if err := atx.UnsignedTx.(UnsignedAtomicTx).EVMStateTransfer(vm, state); err != nil {
-				vm.newBlockChan <- nil
+		if tx, exists := vm.mempool.NextTx(); exists {
+			if err := tx.UnsignedTx.(UnsignedAtomicTx).EVMStateTransfer(vm, state); err != nil {
+				log.Error("Atomic transaction failed verification", "txID", tx.ID(), "error", err)
+				// Discard the transaction from the mempool on failed verification.
```
patrick-ogrady, line 305: is there a reason we don't call `vm.newBlockChan <- nil` in these error cases anymore?

plugin/evm/vm.go
```diff
@@ -671,8 +647,8 @@ func (vm *VM) extractAtomicTx(block *types.Block) *Tx {
 	return atx
 }
 
-// getAtomicTx attempts to get [txID] from the database.
-func (vm *VM) getAtomicTx(txID ids.ID) (*Tx, uint64, error) {
+// getAcceptedAtomicTx attempts to get [txID] from the database.
+func (vm *VM) getAcceptedAtomicTx(txID ids.ID) (*Tx, uint64, error) {
```
patrick-ogrady, line 651: ❤️

plugin/evm/vm.go
```diff
@@ -496,9 +464,17 @@ func (vm *VM) internalBuildBlock() (chainState.Block, error) {
...
+		// transaction into the next block, block building failed and the
+		// transaction should be re-issued unless it was discarded during
+		// atomic tx verification.
+		vm.mempool.CancelCurrentTx()
 		return nil, errCreateBlock
 	}
 
+	// Marks the current tx from the mempool as being successfully issued
+	// into a block.
+	vm.mempool.IssueCurrentTx()
```
patrick-ogrady, line 477: I need to review more of the async code here to ensure there isn't a race here (where we build the block and then a mempool tx is added).

plugin/evm/vm.go
```diff
@@ -496,9 +464,17 @@ func (vm *VM) internalBuildBlock() (chainState.Block, error) {
 	vm.buildBlockLock.Unlock()
 
 	if block == nil {
+		// Signal the mempool that if it was attempting to issue an atomic
+		// transaction into the next block, block building failed and the
+		// transaction should be re-issued unless it was discarded during
+		// atomic tx verification.
+		vm.mempool.CancelCurrentTx()
```
patrick-ogrady, line 471: My following comment also applies here.

plugin/evm/mempool.go
```diff
@@ -0,0 +1,210 @@
...
+func (m *Mempool) RejectTx(txID ids.ID) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	tx, ok := m.issuedTxs[txID]
+	if !ok {
+		return
+	}
+	// If the transaction was issued by the mempool, add it back
+	// to transactions pending issuance.
```
patrick-ogrady, line 199: Would we also want to remove from `issuedTxs` here if we are adding back into the mempool?

plugin/evm/mempool.go
aaronbuchwald: Defense in depth, but you're right there should already be an item in each of the cases.

plugin/evm/vm.go
aaronbuchwald: Yes, if the atomic transaction fails verification, we should still try to build a block consisting of only normal transactions if we can.

plugin/evm/mempool.go
```diff
@@ -0,0 +1,210 @@
...
+	currentTx *Tx
+	// txs is the set of transactions that need to be issued into new blocks
+	txs map[ids.ID]*Tx
+	// issuedTxs is the set of transactions that have been issued into a new block
+	issuedTxs map[ids.ID]*Tx
+	// discardedTxs is an LRU Cache of recently discarded transactions
+	discardedTxs *cache.LRU
+	// Pending is a channel that signals when the mempool is ready to put an atomic transaction
+	// into a new block
+	Pending chan struct{}
```
StephenButtolph, line 28: What is the exact invariant relating to this channel?

plugin/evm/status.go
StephenButtolph: nit: nit: `fmt.Sprintf("%q", s)`

plugin/evm/status.go
StephenButtolph: We can probably use this cleanup in a couple places.

plugin/evm/vm.go
```diff
@@ -693,6 +669,24 @@ func (vm *VM) getAtomicTx(txID ids.ID) (*Tx, uint64, error) {
 	return tx, height, nil
 }
 
+// getAtomicTx returns the requested transaction, status, height if accepted,
+// and whether or not it was found.
+func (vm *VM) getAtomicTx(txID ids.ID) (*Tx, Status, uint64, bool) {
```
StephenButtolph, line 674: This should really be returning an error (Check for `database.NotFound` vs other errors)

plugin/evm/vm.go
aaronbuchwald: Good catch

plugin/evm/mempool.go
aaronbuchwald: Updated to: Pending is a channel of length one, which the mempool ensures has an item on it as long as there is an unissued transaction remaining in [txs]. This is a little jank because of how it interacts with the awaitSubmittedTxs goroutine. Gonna try to refactor that part to simplify this flow.

plugin/evm/mempool.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// assumes the lock is held
+func (m *Mempool) length() int {
+	return len(m.txs) + len(m.issuedTxs)
+}
+
+// Add attempts to add [tx] to the mempool and returns true if the
+// mempool requires a new block to be built.
+func (m *Mempool) AddTx(tx *Tx) error {
```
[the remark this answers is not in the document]
aaronbuchwald: Yes

plugin/evm/status.go
aaronbuchwald: Going to leave this as is for now, since if we add this nit, it should probably be done in AvalancheGo and imported to coreth.

plugin/evm/status.go
aaronbuchwald: Updated to the latter.

plugin/evm/vm.go
aaronbuchwald: Just pushed a refactor of the bd timer, which should make the logic a bit easier to reason about.

plugin/evm/mempool.go
```diff
@@ -0,0 +1,227 @@
...
+	if m.currentTx == nil {
+		return
+	}
+
+	m.discardedTxs.Put(m.currentTx.ID(), m.currentTx)
+	m.currentTx = nil
+}
+
+// RemoveTx removes [txID] from the mempool completely.
+func (m *Mempool) RemoveTx(txID ids.ID) {
```
patrick-ogrady, line 191: Should this also remove from discarded?

plugin/evm/vm.go
```diff
@@ -357,33 +346,14 @@ func (vm *VM) Initialize(
...
 	go ctx.Log.RecoverAndPanic(vm.buildBlockTimer.Dispatch)
 
-	vm.mayBuildBlock = mayBuild
-	vm.tryToBuildBlock = true
 	vm.txPoolStabilizedOk = make(chan struct{}, 1)
-	// TODO: read size from options
-	vm.pendingAtomicTxs = make(chan *Tx, 1024)
-	vm.atomicTxSubmitChan = make(chan struct{}, 1)
+	// TODO: read size from settings
+	vm.mempool = NewMempool(1024)
```
patrick-ogrady, line 356: nit: let's make this a constant

plugin/evm/vm.go
```diff
@@ -738,22 +660,92 @@ func (vm *VM) awaitTxPoolStabilized() {
...
+		// to BuildBlock resets the block status.
+		return 0, false
+	}
+
+	select {
+	case vm.notifyBuildBlockChan <- commonEng.PendingTxs:
+	default:
+		log.Error("Failed to push PendingTxs notification to the consensus engine.")
+	}
+	vm.buildStatus = building
```
patrick-ogrady, line 714: Should we only mark as building if we could push to `notifyBuildBlockChan`?

plugin/evm/vm.go
patrick-ogrady: Much better! Thanks!

plugin/evm/vm.go
```diff
@@ -738,22 +660,92 @@ func (vm *VM) awaitTxPoolStabilized() {
...
+
+// buildBlockTwoStageTimer is a two stage timer that sends a notification
+// to the engine when the VM is ready to build a block.
+// If it should be called back again, it returns the timeout duration at
+// which it should be called again.
+func (vm *VM) buildBlockTwoStageTimer() (time.Duration, bool) {
+	vm.buildBlockLock.Lock()
+	defer vm.buildBlockLock.Unlock()
+
+	switch vm.buildStatus {
```
patrick-ogrady, line 693: nit: throw error in the default case

patrick-ogrady (commented):
left a few comments/questions but this is very close!

plugin/evm/mempool.go
aaronbuchwald: Yes

plugin/evm/vm.go
aaronbuchwald: My thinking on this was that since this is the only message type, it should be in building already if the channel is full, but this is probably not a future proof assumption. Will update.

plugin/evm/vm.go
aaronbuchwald: Good call

plugin/evm/vm.go
aaronbuchwald: Good call

PR #131, opened 2021-03-25
Add e2e tests

aaronbuchwald:
This PR adds a Dockerfile and build image script to coreth, which builds an AvalancheGo image using a base AvalancheGo image and overwriting the evm binary using the current contents of this repo.

Additionally, this PR adds the e2e tests into Coreth's CI and reorganizes the github actions into two separate workflows for PRs and pushes.


.ci/run_e2e_tests.sh
```diff
@@ -0,0 +1,53 @@
...
+SUITE_EXECUTION_VOLUME="avalanche-test-suite_${AVALANCHE_IMAGE_TAG}_$(date +%s)"
+docker volume create "${SUITE_EXECUTION_VOLUME}"
+
+# Docker only allows you to have spaces in the variable if you escape them or use a Docker env file
+CUSTOM_ENV_VARS_JSON="CUSTOM_ENV_VARS_JSON={\"AVALANCHE_IMAGE\":\"${AVALANCHE_IMAGE}\",\"BYZANTINE_IMAGE\":\"${BYZANTINE_IMAGE}\"}"
+
+echo "${CUSTOM_ENV_VARS_JSON}"
+echo "${KURTOSIS_API_IMAGE}"
+echo "${INITIALIZER_IMAGE}"
+docker run \
```
tasinco, line 41: if the test fails does it actually fail the task?

aaronbuchwald: It does in travis, so I believe it would cause the same behavior in GitHub actions. Worth double checking though.

PR #134, opened 2021-03-29
Dev

aaronbuchwald:


danlaine:
One small nit I'd like to include in the next release. Can we change:

```
ApricotPhase1BlockTimestamp: big.NewInt(1617199200), // 10am EST 3/31/2021
```

to

```
ApricotPhase1BlockTimestamp: big.NewInt(1617199200), // 3/31/2021 14:00:00 UTC
```

I think using UTC is less ambiguous (and it's actually 10 AM EDT, not EST)

Issue #165, opened 2021-08-11
`no preimage found for hash` error when making `debug_getModifiedAccountsByNumber` call


stevekunk:
## Issue

I am unable to call `debug_getModifiedAccountsByNumber`. The response is always

```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32000,
        "message": "no preimage found for hash 5a9331cb1d584bc0703e5252f4fccf19b3887d5a12e7edad074d585640a7ff57"
    }
}
```

Likely related to commit 5095d5b45a74781949329e120b7a9fb934438bc0
```diff
+       Preimages bool // Whether to store preimage of trie key to the disk
```

## Version

```
# avalanchego --version
avalanche/1.4.12 [database=v1.4.5, commit=cae93d95c1bcdc02e1370d38ed1c9d87f1c8c814]
```

## Steps to Reproduce

See: https://github.com/stevekunk/avalanche-testnet#issue-no-preimage-found-for-hash

You should be able to `docker-compose up` and then `./curls/debug_getModifiedAccountsByNumber.sh` to reproduce the issue

stevekunk:
Bump. Just looking for any acknowledgement of this issue really. Are there plans to resolve it? Do you need an extra hand?

This worked previously, and I could probably go back and see exactly which change broke the endpoint, but I haven't been following `coreth` development directly. I'm not sure about the motivation behind these changes. I'm also not certain of the future roadmap.

I'm hesitant to start working on anything related to AVAX if api endpoints will be randomly dropped.

stevekunk:
Tagging the top 2 contributors hoping to get some visibility. @aaronbuchwald @patrick-ogrady

aaronbuchwald:
Thanks for creating this issue @stevekunk . It looks like this is caused by preimages being disabled on the nodes. The fix should be to add in a config option that will allow users to enable preimages via the C-Chain config file: https://docs.avax.network/build/references/command-line-interface#coreth-config.

The simple fix to this should be to add a `Preimages` value to the vm config here: https://github.com/ava-labs/coreth/blob/master/plugin/evm/config.go#L33

and then use that value to set `Preimages` to be enabled by setting `ethConfig.Preimages` here: https://github.com/ava-labs/coreth/blob/master/plugin/evm/vm.go#L335.

Would you like to make a PR for this?

PR #171, opened 2021-09-14
feat: Add Preimage option to C-Chain config

stevekunk:
As suggested by @aaronbuchwald here
https://github.com/ava-labs/coreth/issues/165

I just kinda blindly did what aaron suggested. I've been busy. Not sure what the best way to test it or what documentation needs to be updated as a result of this.

### DIFF
--- a/plugin/evm/config.go
+++ b/plugin/evm/config.go
@@ -15,7 +15,7 @@ const (
 	defaultEthApiEnabled               = true
 	defaultNetApiEnabled               = true
 	defaultWeb3ApiEnabled              = true
-	defaultPreimagesEnabled            = true
+	defaultPreimagesEnabled            = false
 	defaultPruningEnabled              = true
 	defaultSnapshotAsync               = true
 	defaultRpcGasCap                   = 2500000000 // 25000000 X 100


stevekunk:
I've never used golang heavily and it was years ago before "go modules" what was a thing. If anyone has any suggestions on how to update just coreth to my fork, I can update my example in to show if this works as expected or not.

plugin/evm/config.go
```diff
@@ -15,6 +15,7 @@ const (
 	defaultEthApiEnabled               = true
 	defaultNetApiEnabled               = true
 	defaultWeb3ApiEnabled              = true
+	defaultPreimagesEnabled            = true
```
aaronbuchwald, line 18: Can you remove this line? We should allow this to default to false.

aaronbuchwald (changes requested):
Can you change the behavior so that the default for `Preimages` is false? Then I think this will be good to merge.

plugin/evm/config.go
stevekunk: Done in 3bf563221

PR #175, opened 2021-12-13
eth/gasprice: Fix nextBaseFee

zensh:
nextBaseFee may be nil and it will panic


zensh:
@StephenButtolph Hi, I want to create a pull request but the master branch is out of date

PR #176, opened 2021-12-13
fix:evm.VM: use ChainID from the genesis block

lebdron:
Make `net_version` call return the correct network ID.


lebdron:
Just in case, ping @aaronbuchwald

PR #193, opened 2022-04-04
fixed typo

tacoo:
Hello, I noticed that there is a typo.


darioush:
Thank you for your contribution. This typo is already fixed in master.

PR #206, opened 2022-08-29
core/rawdb: should not import go-ethereum's rawdb

darioush:


ceyonur (approved):
Maybe we can consider adding a list of imported go-ethereum packages and a small script to crosscheck if we add any additional one.

PR #216, opened 2022-10-07
refactor: execute 'gofmt -s -w .'

kare:


ceyonur:
Unfortunately we won't be able to merge that in. These are mostly geth related changes. We also tried to merge that lint changes to there as well in https://github.com/ethereum/go-ethereum/pull/25670. However it's not merged so we don't apply these in Coreth in order to minimize the diff.
Sorry for the late response.

PR #260, opened 2023-06-14
Use avalanchego's sorting methods

danlaine:
## Why this should be merged

Removes duplicate code

## How this works

Use imported avalanchego sorting functions and add `Less` methods as needed

## How this was tested

New UT

### DIFF
--- a/plugin/evm/tx_test.go
+++ b/plugin/evm/tx_test.go
@@ -219,6 +219,18 @@ func TestEVMOutputLess(t *testing.T) {
 			expected: false,
 		},
 		{
+			name: "first address greater; assetIDs equal",
+			a: EVMOutput{
+				Address: common.BytesToAddress([]byte{0x02}),
+				AssetID: ids.ID{},
+			},
+			b: EVMOutput{
+				Address: common.BytesToAddress([]byte{0x01}),
+				AssetID: ids.ID{},
+			},
+			expected: false,
+		},
+		{
 			name: "addresses equal; first assetID less",
 			a: EVMOutput{
 				Address: common.BytesToAddress([]byte{0x01}),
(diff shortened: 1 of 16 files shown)


plugin/evm/tx_test.go
```diff
@@ -180,3 +185,145 @@ func executeTxTest(t *testing.T, test atomicTxTest) {
...
+			a: EVMInput{
+				Address: common.BytesToAddress([]byte{0x02}),
+				AssetID: ids.ID{0},
+			},
+			b: EVMInput{
+				Address: common.BytesToAddress([]byte{0x01}),
+				AssetID: ids.ID{1},
+			},
+			expected: false,
+		},
```
aaronbuchwald, line 291:
```suggestion
		{
			name: "first address less",
			a: EVMInput{
				Address: common.BytesToAddress([]byte{0x01}),
				AssetID: ids.ID{},
			},
			b: EVMInput{
				Address: common.BytesToAddress([]byte{0x02}),
				AssetID: ids.ID{},
			},
			expected: true,
		},
		{
			name: "first address greater",
			a: EVMInput{
				Address: common.BytesToAddress([]byte{0x02}),
				AssetID: ids.ID{},
			},
			b: EVMInput{
				Address: common.BytesToAddress([]byte{0x01}),
				AssetID: ids.ID{},
			},
			expected: false,
		},
```

Could we make the assetID's equal in these test cases so that it's clear the only difference is the address?

aaronbuchwald (changes requested):
Requested one change to the test cases, thanks for putting this up 🙌

plugin/evm/tx_test.go
danlaine: I think it makes sense to keep it as is to show that the address is the first thing checked for when comparing inputs/outputs. If you want I could add another case where the assetIDs are equal.

aaronbuchwald: Sure let's add one more case for this

danlaine: added

aaronbuchwald (approved):

PR #263, opened 2023-06-14
update golangci-lint-action

danlaine:
## Why this should be merged

Warning about using `save-state`

https://github.com/ava-labs/coreth/actions/runs/5268399245/jobs/9530685492?pr=260#step:7:63

## How this works

2 --> 3

## How this was tested

lint action no longer complains


hexfusion (approved):

abi87 (approved):
Is it aligned with avalanchego?

danlaine:
@abi87 avalanchego actually doesn't use this action. It invokes `lint.sh`, which does use golangci-lint, but not the github action

PR #264, opened 2023-06-15
Bloom filter gossip

joshua-kim:
## Why this should be merged

Performance optimization to lower end-to-end issuance to acceptance time for a transaction.

## How this works

Uses a pull-based approach to gossip instead of a push-based approach. See testing for more details.

## How this was tested
Ran some simulations, results are as follows:

With the existing push-based gossip, we see that as more and more nodes learn the new transaction, the rate at which nodes learn the transaction slows down exponentially.

At first, few nodes know of the transactions, but only one node knows about it so there's not many people gossiping it. Once it hits the 50% mark, the transaction is being gossiped at maximum velocity, since we're at a middle-point where both half the network is willing to gossip the transaction, and the other half is willing to continue forwarding it.

As a majority of the network learns of the transaction, it starts slowing down again. This is because nodes sample peers to gossip to, but the peers will only forward the gossiped transaction if it hasn't seen it already. Over time, peers effectively "absorb" the gossip and don't forward it, so the last few nodes are very unlikely to ever hear of the transaction until a re-gossip cycle kicks in.
image

With the pull-based approach, we see that the transaction is gossiped exponentially quickly as more and more nodes learn the transaction. This is because we now poll peers, and as more peers learn the transaction, it's more likely that you're going to poll a set of peers that has it.
image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -38,6 +38,7 @@ require (
 	github.com/tyler-smith/go-bip39 v1.1.0
 	github.com/urfave/cli/v2 v2.17.2-0.20221006022127-8f469abc00aa
 	golang.org/x/crypto v0.1.0
+	golang.org/x/exp v0.0.0-20230206171751-46f607a40771
 	golang.org/x/sync v0.1.0
 	golang.org/x/sys v0.8.0
 	golang.org/x/text v0.7.0
@@ -120,7 +121,6 @@ require (
 	go.uber.org/atomic v1.10.0 // indirect
 	go.uber.org/multierr v1.8.0 // indirect
 	go.uber.org/zap v1.24.0 // indirect
-	golang.org/x/exp v0.0.0-20230206171751-46f607a40771 // indirect
 	golang.org/x/net v0.7.0 // indirect
 	golang.org/x/term v0.7.0 // indirect
 	gonum.org/v1/gonum v0.11.0 // indirect
--- a/gossip/bloom.go
+++ b/gossip/bloom.go
@@ -1,7 +1,7 @@
 // Copyright (C) 2019-2023, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
-package mempool
+package gossip
 
 import (
 	"encoding/binary"
@@ -10,18 +10,6 @@ import (
 
 	bloomfilter "github.com/holiman/bloomfilter/v2"
 )
-
-type Tx interface {
-	Hash() []byte
-	Marshal() ([]byte, error)
-	Unmarshal(b []byte) error
-}
-
-type Mempool[T Tx] interface {
-	AddTx(tx T) error
-	GetPendingTxs() []T
-	GetPendingTxsBloomFilter() *BloomFilter
-}
 
 func NewBloomFilterFromBytes(bytes []byte) (*BloomFilter, error) {
 	b, err := NewBloomFilter()
@@ -67,7 +55,7 @@ func (b *BloomFilter) Add(hash []byte) {
 	b.resetIfNeeded()
 
 	binary.BigEndian.Uint64(hash)
-	b.bloomFilter.Add(BloomHasher{Hash: hash})
+	b.bloomFilter.Add(bloomHasher{Hash: hash})
 }
 
 // assumes [b.lock] is held
@@ -82,7 +70,7 @@ func (b *BloomFilter) Contains(hash []byte) bool {
 	b.lock.RLock()
 	defer b.lock.RUnlock()
 
-	return b.bloomFilter.Contains(BloomHasher{Hash: hash})
+	return b.bloomFilter.Contains(bloomHasher{Hash: hash})
 }
 
 func (b *BloomFilter) MarshalBinary() ([]byte, error) {
@@ -99,13 +87,13 @@ func (b *BloomFilter) UnmarshalBinary(bytes []byte) error {
 	return b.bloomFilter.UnmarshalBinary(bytes)
 }
 
-var _ hash.Hash64 = (*BloomHasher)(nil)
+var _ hash.Hash64 = (*bloomHasher)(nil)
 
-type BloomHasher struct {
+type bloomHasher struct {
 	hash.Hash64
 	Hash []byte
 }
 
-func (b BloomHasher) Sum64() uint64 {
+func (b bloomHasher) Sum64() uint64 {
 	return binary.BigEndian.Uint64(b.Hash)
 }
--- a/gossip/gossip.go
+++ b/gossip/gossip.go
@@ -68,22 +68,22 @@ func (p *PullGossiper[T]) Start() {
 				return
 			}
 
-			bloom := p.gossiper.Mempool().GetPendingTxsBloomFilter()
-			bloomBytes, err := bloom.MarshalBinary()
-			if err != nil {
-				log.Error("failed to marshal bloom filter", "error", err)
-				continue
-			}
-
-			request := p.gossiper.GossipRequest()
-			request.SetBloomFilter(bloomBytes)
-			requestBytes, err := message.RequestToBytes(p.codec, request)
-			if err != nil {
-				log.Error("failed to marshal gossip request", "error", err)
-				continue
-			}
-
 			for nodeID := range peers {
+				bloom := p.gossiper.Mempool().GetPendingTxsBloomFilter()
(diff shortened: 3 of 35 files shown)


go.mod
```diff
@@ -120,7 +121,6 @@ require (
 	go.uber.org/atomic v1.10.0 // indirect
 	go.uber.org/multierr v1.8.0 // indirect
 	go.uber.org/zap v1.24.0 // indirect
-	golang.org/x/exp v0.0.0-20230206171751-46f607a40771 // indirect
```
joshua-kim, line 123: Undo this diff

gossip/gossip.go
```diff
@@ -0,0 +1,117 @@
...
+	"time"
+
+	"github.com/ava-labs/avalanchego/codec"
+	"github.com/ethereum/go-ethereum/log"
+
+	"github.com/ava-labs/coreth/peer"
+	"github.com/ava-labs/coreth/plugin/evm/message"
+)
+
+type Gossiper[T Tx] interface {
```
joshua-kim, line 17: Trying to figure out if we can get rid of this interface and instead instantiate the request/response from the generic type. I think it's possible but it requires some generics black magic ([ref](https://stackoverflow.com/questions/71791637/cannot-use-variable-of-type-t-as-type-in-argument/71793077#71793077)).

gossip/gossip.go
```diff
@@ -0,0 +1,117 @@
...
+	gossiper        Gossiper[T]
+	client          peer.NetworkClient
+	codec           codec.Manager
+	gossipFrequency time.Duration
+	gossipSize      int
+	shutdown        chan struct{}
+	shutdownWg      *sync.WaitGroup
+}
+
+// TODO propagate fatal events instead of just logging them
```
joshua-kim, line 53: I think we can do this as a separate pr since a lot of spots in coreth that are in a vm goroutine just log events that could definitely be treated as fatal.

gossip/gossip_test.go
```diff
@@ -0,0 +1,49 @@
...
+func (t *testTx) Marshal() ([]byte, error) {
+	return t.hash, nil
+}
+
+func (t *testTx) Unmarshal(b []byte) error {
+	t.hash = b
+	return nil
+}
+
+func TestGossiper_Shutdown(t *testing.T) {
```
joshua-kim, line 31: Add more tests

plugin/evm/block_builder.go
```diff
@@ -8,13 +8,15 @@ import (
 	"time"
 
 	"github.com/ava-labs/avalanchego/utils/timer"
+
 	"github.com/ava-labs/coreth/core/txpool"
 	"github.com/ava-labs/coreth/params"
 
 	"github.com/ava-labs/avalanchego/snow"
 	commonEng "github.com/ava-labs/avalanchego/snow/engine/common"
-	"github.com/ava-labs/coreth/core"
```
joshua-kim, line 16: Undo diff

plugin/evm/codec.go
```diff
@@ -5,7 +5,6 @@ package evm
 
 import (
 	"fmt"
-
```
joshua-kim, line 8: Undo diff

plugin/evm/import_tx_test.go
```diff
@@ -7,9 +7,10 @@ import (
 	"math/big"
 	"testing"
 
-	"github.com/ava-labs/coreth/params"
```
joshua-kim, line 10: Undo diff

plugin/evm/mempool.go
```diff
@@ -10,15 +10,20 @@ import (
 
 	"github.com/ava-labs/avalanchego/cache"
 	"github.com/ava-labs/avalanchego/ids"
-	"github.com/ava-labs/coreth/metrics"
```
joshua-kim, line 13: Undo diff

plugin/evm/mempool.go
```diff
@@ -333,7 +366,7 @@ func (m *Mempool) IssueCurrentTxs() {
 	m.metrics.currentTxs.Update(int64(len(m.currentTxs)))
 
 	// If there are more transactions to be issued, add an item
-	// to Pending.
+	// to MempoolTx.
```
joshua-kim, line 369: Revert this

plugin/evm/mempool.go
```diff
@@ -491,3 +524,25 @@ func (m *Mempool) GetNewTxs() []*Tx {
 	m.metrics.newTxsReturned.Inc(int64(len(cpy))) // Increment the number of newTxs
 	return cpy
 }
+
+var _ gossip.Tx = (*MempoolTx)(nil)
```
joshua-kim, line 528: Maybe we should just make `evm.Tx` and `core.Transaction` implement `gossip.Tx` instead of this wrapper pattern

sync/handlers/handler.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/ava-labs/avalanchego/codec"
 	"github.com/ava-labs/avalanchego/ids"
+
```
joshua-kim, line 11: undo diff

plugin/evm/message/mempool_txs.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+var _ MempoolTxsRequest = (*MempoolEthTxsRequest)(nil)
+var _ MempoolTxsRequest = (*MempoolAtomicTxsRequest)(nil)
+
+// MempoolTxsRequest is a request to initiate a tx gossip session between two
+// nodes.
+type MempoolTxsRequest interface {
```
joshua-kim, line 18: Maybe this should live in the gossip package

gossip/gossip.go
```diff
@@ -0,0 +1,116 @@
...
+
+			bloom := p.gossiper.Mempool().GetPendingTxsBloomFilter()
+			bloomBytes, err := bloom.MarshalBinary()
+			if err != nil {
+				log.Error("failed to marshal bloom filter", "error", err)
+				continue
+			}
+
+			request := p.gossiper.GossipRequest()
+			request.SetBloomFilter(bloomBytes)
```
joshua-kim, line 79: We can recalculate this bloom filter instead of re-using it per gossip session to avoid redundant txs being gossiped to us

gossip/bloom.go
```diff
@@ -0,0 +1,99 @@
...
+	b.bloomFilter.Add(bloomHasher{Hash: hash})
+}
+
+// assumes [b.lock] is held
+func (b *BloomFilter) resetIfNeeded() {
+	filled := b.bloomFilter.PreciseFilledRatio()
+	if filled >= b.refreshRatio {
+		b.bloomFilter, _ = bloomfilter.New(b.m, b.k)
+	}
+}
```
aaronbuchwald, line 67: Hmm if we do this here, then this bloom filter implementation seems to lose the main feature of a bloom filter which is probabilistic yes or definitely not set membership.

Ie if you add A, then exceed the cap, A gets wiped, and if you ask if A is there, it may actually tell you no.

Would it make more sense to expose a function that allows the caller to explicitly wipe the bloom filter and then add all of the items that it still wants in the set back to the bloom filter?

gossip/mempool.go
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) 2019-2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package gossip
+
+type Tx interface {
+	Hash() []byte
```
aaronbuchwald, line 7: Should we restrict this to a 32 byte hash or leave it as `[]byte` ?

peer/network.go
```diff
@@ -564,3 +569,21 @@ func (n *network) TrackBandwidth(nodeID ids.NodeID, bandwidth float64) {
 
 	n.peers.TrackBandwidth(nodeID, bandwidth)
 }
+
+func (n *network) Sample(numPeers int) (set.Set[ids.NodeID], error) {
+	peers := maps.Keys(n.peers.peers)
+	s := sampler.NewUniform()
+	// TODO err
```
aaronbuchwald, line 576: TODO

gossip/bloom.go
joshua-kim: Yeah i think we can just expose number of elements and caller can wipe as needed.

joshua-kim:
Moved to https://github.com/ava-labs/coreth/pull/266

PR #273, opened 2023-07-05
Fix goroutine leaks in core/ tests

darioush:
## Why this should be merged
Fixes goroutine leaks in tests
Accompanying upstream PR https://github.com/ethereum/go-ethereum/pull/27660/files

## How this works
Fixes goroutine leaks in tests

## How this was tested
UT + uber's goleak checker


core/txpool/txpool_test.go
```diff
@@ -123,7 +123,7 @@ func (bc *testBlockChain) SubscribeChainHeadEvent(ch chan<- core.ChainHeadEvent)
 }
 
 func (bc *testBlockChain) SenderCacher() *core.TxSenderCacher {
-	return core.NewTxSenderCacher(1)
+	return &core.TxSenderCacher{}
```
darioush, line 126: This avoids starting goroutines.
Not a very nice way to shut down this without lots of textual modifications.

core/blockchain_test.go
```diff
@@ -355,11 +355,12 @@ func TestBlockChainOfflinePruningUngracefulShutdown(t *testing.T) {
 			return blockchain, nil
 		}
 
-		tempDir := t.TempDir()
 		if err := blockchain.CleanBlockRootsAboveLastAccepted(); err != nil {
 			return nil, err
 		}
+		blockchain.Stop()
```
aaronbuchwald, line 361: is this the right place to call `Stop()` if we are going to use `blockchain.LastAcceptedBlock().Root()` below?

Seems like this works because it returns an in-memory value

core/txpool/txpool_test.go
aaronbuchwald: Could you add a comment here explaining why it's ok to leave this uninitalized?

darioush: actually I'll just fix this since it's a modification to our code to have 
`	SenderCacher() *core.TxSenderCacher`
as part of the `blockChain` interface in `txpool.go`

core/main_test.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+// TestMain uses goleak to verify tests in this package do not leak unexpected
+// goroutines.
+func TestMain(m *testing.M) {
+	opts := []goleak.Option{
+		// No good way to shut down these goroutines:
+		goleak.IgnoreTopFunction("github.com/ava-labs/coreth/core/state/snapshot.(*diskLayer).generate"),
+		goleak.IgnoreTopFunction("github.com/ava-labs/coreth/metrics.(*meterArbiter).tick"),
+		goleak.IgnoreTopFunction("github.com/syndtr/goleveldb/leveldb.(*DB).mpoolDrain"),
```
abi87, line 19: q: why not?

core/blockchain_test.go
```diff
@@ -465,6 +467,7 @@ func testRepopulateMissingTriesParallel(t *testing.T, parallelism int) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	blockchain.Stop()
```
aaronbuchwald, line 470: should this be deferred instead?

darioush, line 470:
```suggestion
	defer blockchain.Stop()
```

anusha-ctrl (approved):
lgtm! were these not copied over from geth?

darioush:
@anusha-ctrl these ones are some of our deviations from upstream. I made an upstream pr (merged) that covers the few missing cases there.

core/main_test.go
darioush: There is no good way to wait for finishing of a goroutine that is started by code that doesn't track it with a wg or expose the wg to the public interface.
Feel free to open a PR and fix if you find a way :)

PR #293, opened 2023-08-01
Bump avalanchego to v1.10.6-rc.4

aaronbuchwald:
This PR bumps AvalancheGo dependency to use v1.10.6-rc.4 and updates the cache sizes used by the chain state cache to use the new sized LRU cache to limit the total memory allocated to the cache based off of bytes size (limited to a total of 20MB) as opposed to a specific number of blocks.

Since AvalancheGo imports Coreth, it will replace the version of AvalancheGo imported by Coreth with the actual current code when it's compiled. As a result, since AvalancheGo has switched to inputting sizes based off of the total memory taken up by these caches instead of the number of elements, we need to update the sizes for the AvalancheGo release.

Otherwise, the chain state cache will go from allocating 100/50/50 elements to the respective caches to allocating that number of bytes to each of those caches, which will end up being zero total blocks.

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -122,9 +122,9 @@ const (
 	codecVersion         = uint16(0)
 	secpFactoryCacheSize = 1024
 
-	decidedCacheSize    = 100
-	missingCacheSize    = 50
-	unverifiedCacheSize = 50
+	decidedCacheSize    = 10 * units.MiB
+	missingCacheSize    = 5 * units.MiB
+	unverifiedCacheSize = 5 * units.MiB
 
 	targetAtomicTxsSize = 40 * units.KiB
 )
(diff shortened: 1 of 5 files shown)


plugin/evm/vm.go
```diff
@@ -122,9 +122,9 @@ const (
 	codecVersion         = uint16(0)
 	secpFactoryCacheSize = 1024
 
-	decidedCacheSize    = 100
-	missingCacheSize    = 50
-	unverifiedCacheSize = 50
+	decidedCacheSize    = 10 * units.MiB
+	missingCacheSize    = 5 * units.MiB
```
StephenButtolph, line 126: the missing cache is not byte based - as it does not contain any blocks

aaronbuchwald: Will update

StephenButtolph (approved):

PR #298, opened 2023-08-09
renamed flags

ceyonur:
We have different flag names in Subnet-EVM as [here](https://github.com/ava-labs/subnet-evm/blob/master/plugin/evm/config.go#L157-L159).  Renamed those to be same with Subnet-EVM.

Also added a deprecation function which would set new configs to old ones if old ones has non-default values, and emits a warn log.

- [ ] document changes

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -1156,8 +1156,8 @@ func (vm *VM) CreateHandlers(context.Context) (map[string]*commonEng.HTTPHandler
 	enabledAPIs = append(enabledAPIs, "avax")
 	apis[avaxEndpoint] = avaxAPI
 
-	if vm.config.CorethAdminAPIEnabled {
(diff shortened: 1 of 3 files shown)


darioush (approved):
LGTM, should we make a corresponding docs PR?

anusha-ctrl (approved):

plugin/evm/vm.go
```diff
@@ -355,6 +359,10 @@ func (vm *VM) Initialize(
 
 	log.Info("Initializing Coreth VM", "Version", Version, "Config", vm.config)
 
+	if deprecateMsg != "" {
```
aaronbuchwald, line 362: Can we put the declaration of `deprecateMsg` inside of the if statement so that it's all in one place?

ceyonur: see the comment:
```
// We should deprecate config flags as the first thing, before we do anything else
// because this can set old flags to new flags. log the message after we have
// initialized the logger.
```

PR #300, opened 2023-08-09
add debug log for dropped pending txs notifications

joshua-kim:
## Why this should be merged

Add a debug log for blocks that we're not building due to the previous block still pending

## How this works

Adds a debug log

## How this was tested

Did not test

### DIFF
--- a/plugin/evm/block_builder.go
+++ b/plugin/evm/block_builder.go
@@ -111,6 +111,7 @@ func (b *blockBuilder) needToBuild() bool {
 func (b *blockBuilder) markBuilding() {
 	// If the engine has not called BuildBlock, no need to send another message.
 	if b.buildSent {
+		log.Debug("previous PendingTxs notification still pending in consensus")
 		return
 	}
 	b.buildBlockTimer.Cancel() // Cancel any future attempt from the timer to send a PendingTxs message


darioush (approved):

plugin/evm/block_builder.go
```diff
@@ -111,6 +111,7 @@ func (b *blockBuilder) needToBuild() bool {
 func (b *blockBuilder) markBuilding() {
 	// If the engine has not called BuildBlock, no need to send another message.
 	if b.buildSent {
```
patrick-ogrady, line 113: Do we want to bound this somehow as a failsafe?

joshua-kim: Hmm that's an interesting idea. I feel like the invariant should be that if we send `PendingTxs` to consensus, we are guaranteed that `BuildBlock` should be called on the VM.

If we do want to bound this by making this something like an atomic int we can as part of a separate PR but it's not something I'm personally inclined towards since I don't feel like the VM should have to worry about this from its perspective. I feel like I'd rather have either extreme of us only sending the notification once until the `BuildBlock` is called, or just not even track this `builtSent` flag at all.

joshua-kim:
Corresponding PR in subnet-evm: https://github.com/ava-labs/subnet-evm/pull/783

plugin/evm/block_builder.go
ceyonur: IMO the engine should signal this to the VM (probably with `BuildBlock` taking an extra err or in the block.context) so that VMs can know that proposerVM has failed creating the block. and they can reset `buildSent`.

Or we can set a safe timeout for this (>30 sec) and reset the `buildSent` once we timeout on `BuildBlock`. 

cc @darioush @aaronbuchwald

aaronbuchwald: This will currently emit a debug log every time we signal there are new transactions, so each time a new transactions event is issued, which seems more often than we would want imo.

aaronbuchwald:
Closing since we added logs to cover this edge case in AvalancheGo instead.

PR #305, opened 2023-08-21
e2e: Add avalanchego e2e job

marun:
## Why this should be merged

Adds a CI job to run the pinned version of the avalanchego e2e tests against the latest version of coreth. Once the PRs for #1547 are merged to avalanchego to ensure equivalent coverage to the current kurtosis e2e, the current coreth kurtosis e2e job can be disabled.

## How this works

A new script in the coreth repo - `./scripts/tests.e2e.sh`:
 - checks out the target version of AvalancheGo
 - configures it to use the local working directory as its coreth dependency
 - builds avalanchego binary
 - runs the AvalancheGo e2e test script

## How this was tested

Checked that the new `AvalancheGo E2E Tests` job exists and is passing.

### DIFF
--- a/scripts/tests.e2e.sh
+++ b/scripts/tests.e2e.sh
@@ -7,30 +7,45 @@ set -euo pipefail
 # e.g.,
 # ./scripts/tests.e2e.sh
 # AVALANCHE_VERSION=v1.10.x ./scripts/tests.e2e.sh
+if ! [[ "$0" =~ scripts/tests.e2e.sh ]]; then
+  echo "must be run from repository root"
+  exit 255
+fi
 
 # Coreth root directory
 CORETH_PATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )"; cd .. && pwd )
 
+# Allow configuring the clone path to point to an existing clone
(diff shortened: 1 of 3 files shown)


scripts/tests.e2e.sh
```diff
@@ -0,0 +1,37 @@
...
+if [[ -d avalanchego ]]; then
+  echo "updating existing clone"
+  cd avalanchego
+  git pull
+  git checkout -B "${avalanche_version}"
+else
+  echo "creating new clone"
+  git clone -b "${avalanche_version}" --single-branch https://github.com/ava-labs/avalanchego.git
+  cd avalanchego
+fi
```
aaronbuchwald, line 27: Could we break this functionality into a function or its own script: https://github.com/ava-labs/subnet-evm/blob/master/scripts/install_avalanchego_release.sh#L61 and re-use it here?

aaronbuchwald: Decided against this offline.

aaronbuchwald:
As discussed offline, it would be great to add an `AVALANCHEGO_CLONE_PATH` variable that we can override if desired. After that, this LGTM

aaronbuchwald (approved):

darioush (approved):

PR #306, opened 2023-08-21
Bump golang version to 1.19.12 for CI

marun:
Also bump targeted version of avalanchego to the v1.10.8 (which also uses golang 1.19.12).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
           token: ${{ secrets.AVALANCHE_PAT }}
       - uses: actions/setup-go@v3
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           token: ${{ secrets.AVALANCHE_PAT }}
       - uses: actions/setup-go@v3
         with:
-          go-version: "1.19"
+          go-version: "1.19.12"
```
darioush, line 34: could we use `~1.19.12` for consistency with the other repositories?

aaronbuchwald (approved):

PR #307, opened 2023-08-23
add upstream test: TestClientBatchRequest_len

darioush:
## Why this should be merged
Test is present in upstream: https://github.com/ethereum/go-ethereum/blob/v1.12.0/rpc/client_test.go#L172

## How this works
Copies the test

## How this was tested
CI


aaronbuchwald (approved):

anusha-ctrl (approved):
lgtm

PR #266, opened 2023-06-17
Bloom filter gossip

joshua-kim:
## Why this should be merged

Performance optimization to lower end-to-end issuance to acceptance time for a transaction.

## How this works

Uses a pull-based approach to gossip instead of a push-based approach. See testing for more details.

## How this was tested
Ran some simulations, results are as follows:

With the existing push-based gossip, we see that as more and more nodes learn the new transaction, the rate at which nodes learn the transaction slows down exponentially.

At first, few nodes know of the transactions, but only one node knows about it so there's not many people gossiping it. Once it hits the 50% mark, the transaction is being gossiped at maximum velocity, since we're at a middle-point where both half the network is willing to gossip the transaction, and the other half is willing to continue forwarding it.

As a majority of the network learns of the transaction, it starts slowing down again. This is because nodes sample peers to gossip to, but the peers will only forward the gossiped transaction if it hasn't seen it already. Over time, peers effectively "absorb" the gossip and don't forward it, so the last few nodes are very unlikely to ever hear of the transaction until a re-gossip cycle kicks in.
image

With the pull-based approach, we see that the transaction is gossiped exponentially quickly as more and more nodes learn the transaction. This is because we now poll peers, and as more peers learn the transaction, it's more likely that you're going to poll a set of peers that has it.
image

PR #316, opened 2023-09-05
Add SDK Router message handling

joshua-kim:
## Why this should be merged

Adds the P2P SDK's router to handle incoming sdk messages

## How this works

Forwards messages that fail unmarshaling against the codec into the SDK router

## How this was tested

Added a unit test

### DIFF
--- a/peer/network.go
+++ b/peer/network.go
@@ -88,16 +88,16 @@ type network struct {
 	outstandingRequestHandlers map[uint32]message.ResponseHandler // maps avalanchego requestID => message.ResponseHandler
 	activeAppRequests          *semaphore.Weighted                // controls maximum number of active outbound requests
 	activeCrossChainRequests   *semaphore.Weighted                // controls maximum number of active outbound cross chain requests
-	router                     *p2p.Router
-	appSender                  common.AppSender                 // avalanchego AppSender for sending messages
-	codec                      codec.Manager                    // Codec used for parsing messages
-	crossChainCodec            codec.Manager                    // Codec used for parsing cross chain messages
-	appRequestHandler          message.RequestHandler           // maps request type => handler
-	crossChainRequestHandler   message.CrossChainRequestHandler // maps cross chain request type => handler
-	gossipHandler              message.GossipHandler            // maps gossip type => handler
-	peers                      *peerTracker                     // tracking of peers & bandwidth
-	appStats                   stats.RequestHandlerStats        // Provide request handler metrics
-	crossChainStats            stats.RequestHandlerStats        // Provide cross chain request handler metrics
+	router                     *p2p.Router                        // handles messages being sent to the generic networking SDK
+	appSender                  common.AppSender                   // avalanchego AppSender for sending messages
+	codec                      codec.Manager                      // Codec used for parsing messages
+	crossChainCodec            codec.Manager                      // Codec used for parsing cross chain messages
+	appRequestHandler          message.RequestHandler             // maps request type => handler
+	crossChainRequestHandler   message.CrossChainRequestHandler   // maps cross chain request type => handler
+	gossipHandler              message.GossipHandler              // maps gossip type => handler
+	peers                      *peerTracker                       // tracking of peers & bandwidth
+	appStats                   stats.RequestHandlerStats          // Provide request handler metrics
+	crossChainStats            stats.RequestHandlerStats          // Provide cross chain request handler metrics
 
 	// Set to true when Shutdown is called, after which all operations on this
 	// struct are no-ops.
--- a/peer/network.go
+++ b/peer/network.go
@@ -103,12 +103,10 @@ type network struct {
 	// struct are no-ops.
 	//
 	// Invariant: Even though `closed` is an atomic, `lock` is required to be
-	// held when sending requests and receiving responses to guarantee that the
-	// network isn't closed during these calls. This is because closing the
-	// network cancels all outstanding requests, which means we must guarantee
-	// never to fulfill, or cancel again, after having cancelled the request.
-	// Similarly, we must ensure we don't register a request that will never be
-	// fulfilled or cancelled.
+	// held when sending requests to guarantee that the network isn't closed
+	// during these calls. This is because closing the network cancels all
+	// outstanding requests, which means we must guarantee never to register a
+	// request that will never be fulfilled or cancelled.
 	closed utils.Atomic[bool]
 }
 
--- a/peer/network.go
+++ b/peer/network.go
@@ -564,6 +564,9 @@ func (n *network) TrackBandwidth(nodeID ids.NodeID, bandwidth float64) {
 
 // invariant: peer/network must use explicitly even request ids.
 // for this reason, [n.requestID] is initialized as zero and incremented by 2.
+// This is for backwards-compatibility while the SDK router exists with the
+// legacy coreth handlers to avoid a (very) narrow edge case where request ids
+// can overlap, resulting in a dropped timeout.
 func (n *network) nextRequestID() uint32 {
 	next := n.requestIDGen
 	n.requestIDGen += 2
--- a/peer/network.go
+++ b/peer/network.go
@@ -332,7 +332,7 @@ func (n *network) AppRequest(ctx context.Context, nodeID ids.NodeID, requestID u
 
 	var req message.Request
 	if _, err := n.codec.Unmarshal(request, &req); err != nil {
-		log.Debug("forwarding unregistered AppRequest", "nodeID", nodeID, "requestID", requestID, "requestLen", len(request), "err", err)
+		log.Debug("forwarding AppRequest to SDK router", "nodeID", nodeID, "requestID", requestID, "requestLen", len(request), "err", err)
 		return n.router.AppRequest(ctx, nodeID, requestID, deadline, request)
 	}
 
--- a/peer/network.go
+++ b/peer/network.go
@@ -269,6 +269,9 @@ func (n *network) CrossChainAppRequest(ctx context.Context, requestingChainID id
 // If [requestID] is not known, this function will emit a log and return a nil error.
 // If the response handler returns an error it is propagated as a fatal error.
 func (n *network) CrossChainAppRequestFailed(ctx context.Context, respondingChainID ids.ID, requestID uint32) error {
+	n.lock.Lock()
+	defer n.lock.Unlock()
+
 	if n.closed.Get() {
 		return nil
 	}
@@ -293,6 +296,9 @@ func (n *network) CrossChainAppRequestFailed(ctx context.Context, respondingChai
 // If [requestID] is not known, this function will emit a log and return a nil error.
 // If the response handler returns an error it is propagated as a fatal error.
 func (n *network) CrossChainAppResponse(ctx context.Context, respondingChainID ids.ID, requestID uint32, response []byte) error {
+	n.lock.Lock()
+	defer n.lock.Unlock()
+
 	if n.closed.Get() {
 		return nil
 	}
@@ -358,13 +364,17 @@ func (n *network) AppRequest(ctx context.Context, nodeID ids.NodeID, requestID u
 // If [requestID] is not known, this function will emit a log and return a nil error.
 // If the response handler returns an error it is propagated as a fatal error.
 func (n *network) AppResponse(ctx context.Context, nodeID ids.NodeID, requestID uint32, response []byte) error {
+	n.lock.Lock()
+
 	if n.closed.Get() {
+		n.lock.Unlock()
 		return nil
 	}
 
 	log.Debug("received AppResponse from peer", "nodeID", nodeID, "requestID", requestID)
 
 	handler, exists := n.markRequestFulfilled(requestID)
+	n.lock.Unlock()
 	if !exists {
 		log.Debug("forwarding AppResponse to SDK router", "nodeID", nodeID, "requestID", requestID, "responseLen", len(response))
 		return n.router.AppResponse(ctx, nodeID, requestID, response)
@@ -383,13 +393,17 @@ func (n *network) AppResponse(ctx context.Context, nodeID ids.NodeID, requestID 
 // error returned by this function is expected to be treated as fatal by the engine
 // returns error only when the response handler returns an error
 func (n *network) AppRequestFailed(ctx context.Context, nodeID ids.NodeID, requestID uint32) error {
+	n.lock.Lock()
+
 	if n.closed.Get() {
+		n.lock.Unlock()
 		return nil
 	}
 
 	log.Debug("received AppRequestFailed from peer", "nodeID", nodeID, "requestID", requestID)
 
 	handler, exists := n.markRequestFulfilled(requestID)
+	n.lock.Unlock()
 	if !exists {
 		log.Debug("forwarding AppRequestFailed to SDK router", "nodeID", nodeID, "requestID", requestID)
 		return n.router.AppRequestFailed(ctx, nodeID, requestID)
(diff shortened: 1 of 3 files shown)


peer/network.go
```diff
@@ -378,9 +384,10 @@ func (n *network) AppResponse(_ context.Context, nodeID ids.NodeID, requestID ui
 
 	handler, exists := n.markRequestFulfilled(requestID)
 	if !exists {
-		// Should never happen since the engine should be managing outstanding requests
-		log.Error("received AppResponse to unknown request", "nodeID", nodeID, "requestID", requestID, "responseLen", len(response))
-		return nil
+		log.Debug("received AppResponse to unknown request", "nodeID", nodeID, "requestID", requestID, "responseLen", len(response))
```
darioush, line 387: could we update this log (and the one in AppRequest and the one in AppRequestFailed) or only emit them after checking the err returned from the call to the SDK?

darioush (approved):

peer/network.go
```diff
@@ -335,8 +339,8 @@ func (n *network) AppRequest(ctx context.Context, nodeID ids.NodeID, requestID u
 
 	var req message.Request
 	if _, err := n.codec.Unmarshal(request, &req); err != nil {
-		log.Debug("failed to unmarshal app request", "nodeID", nodeID, "requestID", requestID, "requestLen", len(request), "err", err)
-		return nil
+		log.Debug("forwarding unknown AppRequest to sdk", "nodeID", nodeID, "requestID", requestID, "requestLen", len(request), "err", err)
```
patrick-ogrady, line 342: nit: remove `to sdk`

patrick-ogrady, line 342: Shouldn't this log be an unmarshal error unlike the other changes (where it is "unknown")

joshua-kim: In this instance the SDK requests are prepended with a varint identifier, so they won't marshal correctly into the EVM's codec. It's possible that some opaque bytes we can't unmarshal with the EVM's codec came from the SDK, since the SDK handlers will define their own way of interpreting the bytes.

peer/network.go
```diff
@@ -335,8 +339,8 @@ func (n *network) AppRequest(ctx context.Context, nodeID ids.NodeID, requestID u
 
 	var req message.Request
 	if _, err := n.codec.Unmarshal(request, &req); err != nil {
-		log.Debug("failed to unmarshal app request", "nodeID", nodeID, "requestID", requestID, "requestLen", len(request), "err", err)
-		return nil
+		log.Debug("forwarding unknown AppRequest", "nodeID", nodeID, "requestID", requestID, "requestLen", len(request), "err", err)
```
aaronbuchwald, line 342: Can we change all of these debug logs to be more specific to what's happening? `unknown request` sounds like an error message, but this is expected behavior during the transition period.

Can we change to `forwarding request unknown to codec to SDK handler` or similar?

peer/network.go
```diff
@@ -87,23 +89,25 @@ type network struct {
...
+	appStats                   stats.RequestHandlerStats        // Provide request handler metrics
+	crossChainStats            stats.RequestHandlerStats        // Provide cross chain request handler metrics
 
 	// Set to true when Shutdown is called, after which all operations on this
 	// struct are no-ops.
 	closed utils.Atomic[bool]
 }
 
-func NewNetwork(appSender common.AppSender, codec codec.Manager, crossChainCodec codec.Manager, self ids.NodeID, maxActiveAppRequests int64, maxActiveCrossChainRequests int64) Network {
+func NewNetwork(router *p2p.Router, appSender common.AppSender, codec codec.Manager, crossChainCodec codec.Manager, self ids.NodeID, maxActiveAppRequests int64, maxActiveCrossChainRequests int64) Network {
```
aaronbuchwald, line 108: Following up on https://github.com/ava-labs/coreth/pull/266#discussion_r1313622140, how do we ensure that there are never overlaps between the requestID used by the SDK and the monotonically increasing requestID used by the current networking code?

I see you marked this comment as resolved on the old PR, how we avoid the following situation?

```
1. Issue a request via the SDK with requestID 1
2. Issue an unrelated request with the current networking code with requestID 1
3. Receive a response to the SDK request with requestID 1 and incorrectly clear the handler registered to the current networking code

Comment (aaronbuchwald) [reply peer/network.go:108]: It seems like we treat an un-requested `AppResponse` as a fatal error in the p2p networking code assuming the engine ensures we don't receive unrequested responses, so I assume if we mess this up in the opposite order that it becomes an unexpected fatal error.

File: peer/network_test.go [hunk 5]
```diff
@@ -457,12 +461,11 @@ func TestHandleInvalidMessages(t *testing.T) {
...
-	assert.NoError(t, clientNetwork.AppResponse(context.Background(), nodeID, requestID, requestMessage))
-	assert.NoError(t, clientNetwork.AppResponse(context.Background(), nodeID, requestID, garbageResponse))
-	assert.NoError(t, clientNetwork.AppResponse(context.Background(), nodeID, requestID, emptyResponse))
-	assert.NoError(t, clientNetwork.AppResponse(context.Background(), nodeID, requestID, nilResponse))
-	assert.NoError(t, clientNetwork.AppRequestFailed(context.Background(), nodeID, requestID))
+	assert.ErrorIs(t, p2p.ErrUnrequestedResponse, clientNetwork.AppResponse(context.Background(), nodeID, requestID, gossipMsg))
+	assert.ErrorIs(t, p2p.ErrUnrequestedResponse, clientNetwork.AppResponse(context.Background(), nodeID, requestID, requestMessage))
+	assert.ErrorIs(t, p2p.ErrUnrequestedResponse, clientNetwork.AppResponse(context.Background(), nodeID, requestID, garbageResponse))
+	assert.ErrorIs(t, p2p.ErrUnrequestedResponse, clientNetwork.AppResponse(context.Background(), nodeID, requestID, emptyResponse))
+	assert.ErrorIs(t, p2p.ErrUnrequestedResponse, clientNetwork.AppResponse(context.Background(), nodeID, requestID, nilResponse))
```
aaronbuchwald, line 468: This test was written to ensure that an invalid message NEVER triggers an unintentional fatal error, so it seems a bit weird to change it in this way.

peer/network.go
joshua-kim: I believe if we make two concurrent requests with the same request id, the last one to be sent out overwrites the previous one's request id ([ref](https://github.com/ava-labs/avalanchego/blob/14b8c98ce479d1c6f9c0fdb32f3fa60e6e137eb9/snow/networking/router/chain_router.go#L174)).

This means that the incoming `AppResponse` will only go to either the coreth handler, or the sdk handler.

In the case that the sdk request is taking over the codec request, this means that there will be a handler present from the original `AppRequest` flow, so we'll send the sdk response to the codec handler which will fail unmarshalling and we'll emit a debug log.

In the case that sdk goes first and codec goes second, then the codec handler will get its response as it expected and sdk will have its request time out correctly.

peer/network_test.go
joshua-kim: Yeah I was unsure if it made more sense to just remove these test cases or just test for the sdk error. I can remove these if that makes more sense.

peer/network.go
StephenButtolph: > sdk will have its request time out correctly.

The SDK will never receive the timeout - as the request was cancelled.

StephenButtolph, line 108: imo - we should initialize `requestIDGen` to `MaxUint32` here and then decrement the counter per requestID in `sendAppRequest`.

This guarantees to avoid any overwriting unless we send 4 billion AppRequest messages.

In the case we _do_ send 4 billion requests, we could encounter unexpected behavior if _both_ message handlers have outstanding requests (with the same requestID + nodeID). This wouldn't panic - but the request would never be delivered to one of the handlers.

Currently in coreth we only use `AppRequest` messages for state syncing afaict. Assuming we will never send any mempool bloom filter requests during state sync, we should be fine regardless of the number of requests we send.

StephenButtolph: A separate alternative would be to explicitly separate the `requestID` space by sending odd requests from the coreth client and even requests from the avalanchego sdk.

joshua-kim: Ah yes. We might have to make an upstream change to fire AppRequestFailed on eviction then if there's a collision in the request id.

joshua-kim: Discussed offline and it seems cleanest to just use even/odd to avoid the slim possibility of a collision

PR to avalanchego: https://github.com/ava-labs/avalanchego/pull/1975/files

joshua-kim: See https://github.com/ava-labs/coreth/pull/316#discussion_r1317645330, we opted to break up the request id space into odd/even to avoid collisions. Turns out we do drop this and we don't fire the timeout.

peer/network.go
joshua-kim: Patrick disliked the `... to sdk` wording from the first revision so I'll use `forwarding unregistered AppRequest`

peer/network.go
```diff
@@ -335,8 +332,8 @@ func (n *network) AppRequest(ctx context.Context, nodeID ids.NodeID, requestID u
 
 	var req message.Request
 	if _, err := n.codec.Unmarshal(request, &req); err != nil {
-		log.Debug("failed to unmarshal app request", "nodeID", nodeID, "requestID", requestID, "requestLen", len(request), "err", err)
-		return nil
+		log.Debug("forwarding unregistered AppRequest", "nodeID", nodeID, "requestID", requestID, "requestLen", len(request), "err", err)
```
StephenButtolph, line 335:
```suggestion
		log.Debug("forwarding AppRequest to SDK router", "nodeID", nodeID, "requestID", requestID, "requestLen", len(request), "err", err)
```

StephenButtolph (approved):

peer/network.go
```diff
@@ -366,7 +363,7 @@ func (n *network) AppRequest(ctx context.Context, nodeID ids.NodeID, requestID u
 // Error returned by this function is expected to be treated as fatal by the engine
 // If [requestID] is not known, this function will emit a log and return a nil error.
 // If the response handler returns an error it is propagated as a fatal error.
-func (n *network) AppResponse(_ context.Context, nodeID ids.NodeID, requestID uint32, response []byte) error {
+func (n *network) AppResponse(ctx context.Context, nodeID ids.NodeID, requestID uint32, response []byte) error {
 	n.lock.Lock()
 	defer n.lock.Unlock()
```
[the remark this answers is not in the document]
joshua-kim: Yeah the router invokes a callback which can be potentially long running so we don't want to hold it for this entire duration.

peer/network.go
```diff
@@ -365,15 +365,16 @@ func (n *network) AppRequest(ctx context.Context, nodeID ids.NodeID, requestID u
 // If the response handler returns an error it is propagated as a fatal error.
 func (n *network) AppResponse(ctx context.Context, nodeID ids.NodeID, requestID uint32, response []byte) error {
 	n.lock.Lock()
```
darioush, line 367: since `closed` is atomic, could we move this lock to directly before
`	handler, exists := n.markRequestFulfilled(requestID)`?

Should we move it to inside `markRequestFulfilled` instead?

joshua-kim: Yeah that makes much more sense than what I was currently doing.

peer/network.go
```diff
@@ -564,3 +549,12 @@ func (n *network) TrackBandwidth(nodeID ids.NodeID, bandwidth float64) {
 
 	n.peers.TrackBandwidth(nodeID, bandwidth)
 }
+
+// invariant: peer/network must use explicitly even request ids.
```
patrick-ogrady, line 553: nit: may be nice to explain why this is

peer/network.go
```diff
@@ -366,10 +357,7 @@ func (n *network) AppRequest(ctx context.Context, nodeID ids.NodeID, requestID u
 // Error returned by this function is expected to be treated as fatal by the engine
 // If [requestID] is not known, this function will emit a log and return a nil error.
 // If the response handler returns an error it is propagated as a fatal error.
-func (n *network) AppResponse(_ context.Context, nodeID ids.NodeID, requestID uint32, response []byte) error {
-	n.lock.Lock()
-	defer n.lock.Unlock()
-
+func (n *network) AppResponse(ctx context.Context, nodeID ids.NodeID, requestID uint32, response []byte) error {
 	if n.closed.Get() {
```
StephenButtolph, line 361: This is a very subtle change. So we should be terrified to make it.

I believe we _must_ hold the lock when checking for `n.closed` on all of the inbound `response` + `requestFailed` flows to avoid a potential panic due to a write to a closed channel (or closing a channel twice).

joshua-kim: As discussed offline I'll see if I can make a separate regression test for this invariant as a follow-up to this PR.

peer/network.go
joshua-kim: Yeah good idea. Will comment here.

peer/network.go
StephenButtolph: I documented the invariant on `n.closed` as well 👍

peer/network.go
```diff
@@ -366,21 +371,21 @@ func (n *network) AppRequest(ctx context.Context, nodeID ids.NodeID, requestID u
...
 // If [requestID] is not known, this function will emit a log and return a nil error.
 // If the response handler returns an error it is propagated as a fatal error.
-func (n *network) AppResponse(_ context.Context, nodeID ids.NodeID, requestID uint32, response []byte) error {
+func (n *network) AppResponse(ctx context.Context, nodeID ids.NodeID, requestID uint32, response []byte) error {
 	n.lock.Lock()
-	defer n.lock.Unlock()
 
 	if n.closed.Get() {
+		n.lock.Unlock()
 		return nil
```
StephenButtolph, line 379: Are we okay with dropping responses (and timeouts) that should have been sent to the SDK router after we close the network?

StephenButtolph: If we are not okay with this - then I think we'll need to be fairly careful around what gets passed through to the router.

joshua-kim: We chatted offline but agreed it was cleanest to leave the `Router` code as-is, and just drop the `closed` check on the server-end because we're guaranteed that the outstanding request is empty on shutdown because we empty it on `Shutdown` and stop sending requests after the flag is set.

StephenButtolph (approved):

peer/network.go
```diff
@@ -87,23 +88,31 @@ type network struct {
...
-	appSender                  common.AppSender                   // avalanchego AppSender for sending messages
-	codec                      codec.Manager                      // Codec used for parsing messages
-	crossChainCodec            codec.Manager                      // Codec used for parsing cross chain messages
-	appRequestHandler          message.RequestHandler             // maps request type => handler
-	crossChainRequestHandler   message.CrossChainRequestHandler   // maps cross chain request type => handler
-	gossipHandler              message.GossipHandler              // maps gossip type => handler
-	peers                      *peerTracker                       // tracking of peers & bandwidth
-	appStats                   stats.RequestHandlerStats          // Provide request handler metrics
-	crossChainStats            stats.RequestHandlerStats          // Provide cross chain request handler metrics
+	router                     *p2p.Router
```
patrick-ogrady, line 91: nit: should this have a trailing comment like the others?

peer/network.go
```diff
@@ -87,23 +88,31 @@ type network struct {
...
+	crossChainRequestHandler   message.CrossChainRequestHandler // maps cross chain request type => handler
+	gossipHandler              message.GossipHandler            // maps gossip type => handler
+	peers                      *peerTracker                     // tracking of peers & bandwidth
+	appStats                   stats.RequestHandlerStats        // Provide request handler metrics
+	crossChainStats            stats.RequestHandlerStats        // Provide cross chain request handler metrics
 
 	// Set to true when Shutdown is called, after which all operations on this
 	// struct are no-ops.
+	//
+	// Invariant: Even though `closed` is an atomic, `lock` is required to be
```
patrick-ogrady, line 105: if we have this invariant, should it be an atomic?

StephenButtolph: We do not _always_ need to hold the lock... Only when sending a request do we need this lock held

patrick-ogrady: sorry, glanced over that

patrick-ogrady (approved):

peer/network.go
joshua-kim: Yeah this is definitely confusing nuance of the existing code that I definitely didn't even notice. It might make a lot of sense to switch this out with just a normal to make it more clear in a follow-up...

PR #318, opened 2023-09-05
Use SDK Gossip

joshua-kim:


plugin/evm/gossip_mempool.go
```diff
@@ -0,0 +1,140 @@
...
+				reset, err := gossip.ResetBloomFilterIfNeeded(g.bloom, txGossipBloomMaxFilledRatio)
+				if err != nil {
+					log.Error("failed to reset bloom filter", "err", err)
+					continue
+				}
+
+				if reset {
+					log.Debug("resetting bloom filter", "reason", "reached max filled ratio")
+
+					pending := g.mempool.Pending(false)
```
StephenButtolph, line 88: Can we document why we pass `true` or `false` into `Pending`? We pass both values in different situations without much explanation

plugin/evm/gossip_mempool.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// Add enqueues the transaction to the mempool. Subscribe should be called
+// to receive an event if tx is actually added to the mempool or not.
+func (g *GossipEthTxPool) Add(tx *GossipEthTx) error {
+	if err := g.mempool.AddRemotes([]*types.Transaction{tx.Tx})[0]; err != nil {
+		return err
+	}
+
+	return nil
```
StephenButtolph, line 108:
```suggestion
	return g.mempool.AddRemotes([]*types.Transaction{tx.Tx})[0]
```

plugin/evm/mempool.go
```diff
@@ -259,16 +274,51 @@ func (m *Mempool) addTx(tx *Tx, force bool) error {
...
+		}
+	}
+}
+
+func (m *Mempool) GetFilter() ([]byte, []byte, error) {
+	m.lock.RLock()
+	defer m.lock.RUnlock()
+
+	bloom, err := m.bloom.Bloom.MarshalBinary()
+	return bloom, m.bloom.Salt[:], err
```
StephenButtolph, line 319: I believe `ResetBloomFilterIfNeeded` reuses the `Salt` array - so wouldn't the returned slice be racy?

plugin/evm/gossip_mempool.go
```diff
@@ -0,0 +1,140 @@
...
+		return f(&GossipEthTx{Tx: tx})
+	})
+}
+
+func (g *GossipEthTxPool) GetFilter() ([]byte, []byte, error) {
+	g.lock.RLock()
+	defer g.lock.RUnlock()
+
+	bloom, err := g.bloom.Bloom.MarshalBinary()
+	return bloom, g.bloom.Salt[:], err
```
StephenButtolph, line 122: I believe `ResetBloomFilterIfNeeded` reuses the `Salt` array - so wouldn't the returned slice be racy?

plugin/evm/vm.go
```diff
@@ -316,7 +342,7 @@ func (vm *VM) GetActivationTime() time.Time {
 
 // Initialize implements the snowman.ChainVM interface
 func (vm *VM) Initialize(
-	_ context.Context,
+	ctx context.Context,
```
StephenButtolph, line 345: This context is the context of the `Initialize` call - not the context of the VM. It should be assumed that `ctx` could be cancelled (which would mark `vm.backgroundCtx` as cancelled).

joshua-kim: Ah right. I'll replace this with a `context.TODO()`.

plugin/evm/mempool.go
joshua-kim: In my head I for whatever reason thought I was doing a copy here. Updated to use a copy.

plugin/evm/gossip_mempool.go
joshua-kim: In my head I for whatever reason thought I was doing a copy here. Updated to use a copy.

plugin/evm/gossip_mempool.go
joshua-kim: We can remove this call with a call to `IteratePending` I think.

plugin/evm/gossip_mempool.go
```diff
@@ -0,0 +1,140 @@
...
+
+type GossipEthTxPool struct {
+	mempool    *txpool.TxPool
+	pendingTxs chan core.NewTxsEvent
+
+	bloom *gossip.BloomFilter
+	lock  sync.RWMutex
+}
+
+func (g *GossipEthTxPool) Subscribe(shutdownChan chan struct{}, shutdownWg *sync.WaitGroup) {
```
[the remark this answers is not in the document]
joshua-kim: +1, chatted offline about wrapping w/ goroutine.

core/txpool/txpool.go
```diff
@@ -630,6 +630,19 @@ func (pool *TxPool) PendingSize() int {
 	return count
 }
 
+// IteratePending iterates over the [pool.pending] until [f] returns false.
+func (pool *TxPool) IteratePending(f func(tx *types.Transaction) bool) {
+	pending := pool.Pending(true)
```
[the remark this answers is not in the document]
joshua-kim: I think it's actually reasonable to execute w/ lock held actually given that our only current use-case is reading the values to copy into a struct.

plugin/evm/gossip_mempool.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func (tx *GossipAtomicTx) Marshal() ([]byte, error) {
+	return Codec.Marshal(message.Version, tx)
+}
+
+func (tx *GossipAtomicTx) Unmarshal(bytes []byte) error {
+	_, err := Codec.Unmarshal(bytes, tx)
+	return err
+}
```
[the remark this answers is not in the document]
joshua-kim: Updated, working on a regression test.

plugin/evm/gossip_mempool.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (tx *GossipAtomicTx) Marshal() ([]byte, error) {
+	return tx.Tx.SignedBytes(), nil
+}
+
+func (tx *GossipAtomicTx) Unmarshal(bytes []byte) error {
+	result := &Tx{}
+
+	_, err := Codec.Unmarshal(bytes, result)
```
StephenButtolph, line 42:
```suggestion
	_, err := Codec.Unmarshal(bytes, result)
	if err != nil {
		return err
	}
```

plugin/evm/gossip_mempool.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (g *GossipEthTxPool) GetFilter() ([]byte, []byte, error) {
+	g.lock.RLock()
+	defer g.lock.RUnlock()
+
+	bloom, err := g.bloom.Bloom.MarshalBinary()
+	salt := make([]byte, len(g.bloom.Salt))
+	copy(salt, g.bloom.Salt[:])
+
+	return bloom, salt, err
```
StephenButtolph, line 128:
```suggestion
	salt := g.bloom.Salt

	return bloom, salt[:], err
```

plugin/evm/mempool.go
```diff
@@ -259,16 +274,54 @@ func (m *Mempool) addTx(tx *Tx, force bool) error {
...
+
+func (m *Mempool) GetFilter() ([]byte, []byte, error) {
+	m.lock.RLock()
+	defer m.lock.RUnlock()
+
+	bloom, err := m.bloom.Bloom.MarshalBinary()
+	salt := make([]byte, len(m.bloom.Salt))
+	copy(salt, m.bloom.Salt[:])
+
+	return bloom, salt, err
```
StephenButtolph, line 322:
```suggestion
	salt := m.bloom.Salt

	return bloom, salt[:], err
```

plugin/evm/gossip_mempool.go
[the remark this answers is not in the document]
joshua-kim: Added regression test. I was both not populating `Tx` as well as not correctly grabbing both the signed + unsigned byetes.

plugin/evm/gossip_mempool.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (tx *GossipAtomicTx) Unmarshal(bytes []byte) error {
+	result := &Tx{}
+
+	_, err := Codec.Unmarshal(bytes, result)
+	if err != nil {
+		return err
+	}
+	unsignedBytes, err := Codec.Marshal(codecVersion, result.UnsignedAtomicTx)
```
StephenButtolph, line 46: This should have the typeID
```suggestion
	unsignedBytes, err := Codec.Marshal(codecVersion, &result.UnsignedAtomicTx)
```

plugin/evm/mempool.go
```diff
@@ -8,10 +8,13 @@ import (
...
+
 	"github.com/ava-labs/avalanchego/cache"
 	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/network/p2p/gossip"
+
 	"github.com/ava-labs/coreth/metrics"
-	"github.com/ethereum/go-ethereum/log"
 )
 
 const (
```
patrick-ogrady, line 20: Move `discardedTxsCacheSize` to `vm.go`?

plugin/evm/vm.go
```diff
@@ -128,8 +129,26 @@ const (
 	bytesToIDCacheSize  = 5 * units.MiB
 
 	targetAtomicTxsSize = 40 * units.KiB
+
+	// p2p app protocols
+	ethTxGossipProtocol    = 0x0
+	atomicTxGossipProtocol = 0x1
+
+	// gossip constants
+	txGossipMaxResponseSize        = 20 * units.KiB
```
patrick-ogrady, line 138: What happens if there is a single transaction that is this large (like a contract deploy)? I'm assuming it sends the first item, even if it exceeds?

patrick-ogrady: Nonetheless, I think we may want to at least set this larger than the max contract deploy size.

plugin/evm/gossip_mempool.go
```diff
@@ -0,0 +1,148 @@
...
+		select {
+		case <-ctx.Done():
+			log.Debug("shutting down subscription")
+			return
+		case pendingTxs := <-g.pendingTxs:
+			g.lock.Lock()
+			for _, pendingTx := range pendingTxs.Txs {
+				tx := &GossipEthTx{Tx: pendingTx}
+				g.bloom.Add(tx)
+				reset, err := gossip.ResetBloomFilterIfNeeded(g.bloom, txGossipBloomMaxFilledRatio)
```
patrick-ogrady, line 91: Seems like we'd want to reset far below a target false positive rate of 0.75?

plugin/evm/vm.go
joshua-kim: In this case we canter gossip it. I based this by seeing that most txs are between 300B to 1KB... but theoretically I think our max message size is a MB but that sounds too high. Ideally this is the p100 of our tx size but I don't know if we define one.

plugin/evm/vm.go
```diff
@@ -946,13 +982,67 @@ func (vm *VM) SetState(_ context.Context, state snow.State) error {
...
+	vm.shutdownWg.Add(1)
+	go func() {
+		ethTxPool.Subscribe(vm.backgroundCtx)
+		vm.shutdownWg.Done()
+	}()
+
+	ethTxGossipHandler := &p2p.ValidatorHandler{
+		ValidatorSet: vm.validators,
+		Handler: &p2p.ThrottlerHandler{
+			Throttler: p2p.NewSlidingWindowThrottler(throttlingPeriod, throttlingLimit),
```
patrick-ogrady, line 1006: nit: why doesn't this live in `throttling`?

plugin/evm/mempool.go
```diff
@@ -259,16 +274,53 @@ func (m *Mempool) addTx(tx *Tx, force bool) error {
 	for utxoID := range utxoSet {
 		m.utxoSpenders[utxoID] = tx
 	}
+
+	m.bloom.Add(&GossipAtomicTx{Tx: tx})
+	reset, err := gossip.ResetBloomFilterIfNeeded(m.bloom, txGossipBloomMaxFilledRatio)
```
patrick-ogrady, line 279: Do we have a metric for resets?

plugin/evm/vm.go
joshua-kim: This is because I wanted to keep all the sdk components in their own package instead of scattering them across the networking package. Eventually it might make sense to move them all into a sdk repo as well.

plugin/evm/gossip_mempool.go
joshua-kim: Yeah you're right. I forgot to update this when we changed this signature to be the false positive probability instead of the ratio of filled elements... I'll update this to 5%

plugin/evm/mempool.go
joshua-kim: I think this is unrelated to the current diff (correct me if wrong) so I'm down to do this as followup instead

plugin/evm/mempool.go
joshua-kim: We don't have metrics on this... but we should have metrics on all of the new gossip stuff and this is definitely a metric that we would want to track

plugin/evm/vm.go
patrick-ogrady: Seems pretty easy to block the mempool then by issuing 1 contract deploy larger than 20 KiB with the highest fee?

patrick-ogrady: Unless we want to avoid gossiping things above a certain size?

StephenButtolph:
Important ref: https://github.com/ava-labs/coreth/blob/master/core/txpool/txpool.go#L63-L69

aaronbuchwald:
Can we add metrics to show how many transactions are sent over the bloom filter mechanism, how many hits/misses we find while responding to them, and what portion of transactions we are able to iterate before hitting the max size of the message?

Are there any other metrics we should include cc @StephenButtolph @patrick-ogrady @joshua-kim  ?

plugin/evm/vm.go
```diff
@@ -210,7 +229,9 @@ func init() {
 
 // VM implements the snowman.ChainVM interface
 type VM struct {
-	ctx *snow.Context
+	ctx           *snow.Context   // TODO rename to snowCtx
+	backgroundCtx context.Context // TODO rename to ctx
+	cancel        context.CancelFunc
```
aaronbuchwald, line 234: Why are we embedding `context.Context` in the VM? This is explicitly recommended against in golang.

plugin/evm/gossip_mempool.go
```diff
@@ -0,0 +1,148 @@
...
+func (tx *GossipAtomicTx) GetID() ids.ID {
+	return tx.Tx.ID()
+}
+
+func (tx *GossipAtomicTx) Marshal() ([]byte, error) {
+	return tx.Tx.SignedBytes(), nil
+}
+
+func (tx *GossipAtomicTx) Unmarshal(bytes []byte) error {
+	result := &Tx{}
```
aaronbuchwald, line 40: Can we use this here: https://github.com/ava-labs/coreth/blob/master/plugin/evm/codec.go#L68 ?

plugin/evm/vm.go
joshua-kim: Replaced w/ target gossip size

plugin/evm/vm.go
joshua-kim: This is because the gossip package uses contexts to cancel (e.g `gossiper.Gossip(ctx)` so the caller can shut down the gossip thread by calling `cancel()`. The alternative is to use a channel

core/txpool/txpool.go
```diff
@@ -630,6 +630,21 @@ func (pool *TxPool) PendingSize() int {
...
+	pool.mu.RLock()
+	defer pool.mu.RUnlock()
+
+	for _, list := range pool.pending {
+		for _, tx := range list.txs.items {
+			if !f(tx) {
+				return
+			}
+		}
+	}
```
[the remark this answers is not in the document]
joshua-kim: `pending` and `items` are both maps which have random-ish iteration orders so I'm not super concerned about starvation but it could be a good follow-up

plugin/evm/mempool_atomic_gossiping_test.go
```diff
@@ -193,10 +195,12 @@ func TestMempoolPriorityDrop(t *testing.T) {
...
 
 	tx1 := createImportTx(t, vm, ids.ID{1}, params.AvalancheAtomicTxFee)
-	assert.NoError(mempool.AddTx(tx1))
+	err := mempool.AddTx(tx1)
+	assert.NoError(err)
 	assert.True(mempool.has(tx1.ID()))
 	tx2 := createImportTx(t, vm, ids.ID{2}, params.AvalancheAtomicTxFee)
-	assert.ErrorIs(mempool.AddTx(tx2), errInsufficientAtomicTxFee)
+	err = mempool.AddTx(tx2)
+	assert.ErrorIs(err, errInsufficientAtomicTxFee)
```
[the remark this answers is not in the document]
joshua-kim: Unstaged

plugin/evm/vm.go
```diff
@@ -502,13 +532,17 @@ func (vm *VM) Initialize(
...
+	if err != nil {
+		return fmt.Errorf("failed to initialize mempool: %w", err)
+	}
 
 	if err := vm.initializeMetrics(); err != nil {
 		return err
 	}
 
 	// initialize peer network
+	vm.validators = p2p.NewValidators(vm.ctx.Log, vm.ctx.SubnetID, vm.ctx.ValidatorState, maxValidatorSetStaleness)
```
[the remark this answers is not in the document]
joshua-kim: https://github.com/ava-labs/avalanchego/issues/1998

core/txpool/txpool.go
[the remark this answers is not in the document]
aaronbuchwald: Sounds good, the other concern would be if we end up sending only transactions with future nonces and don't include the necessary transaction to make them executable. This is an edge case when we hit the max size, so I don't think this needs to block the PR.

plugin/evm/vm.go
aaronbuchwald: Imo we should not embed a context since there's a very direct recommendation against this.

Could we use the alternative?

joshua-kim: This should be safe since we're only using the contexts to to cancel the goroutine, we can definitely use a channel but since it requires us to keep going back-and-forth between avalanchego and coreth to cut more rcs I think it would be easier to do this as a follow-up to this PR

aaronbuchwald: Discussed offline, this is fine since we don't include key-value pairs in contexts here, so will need to be aware of how this is used going forward.

.github/workflows/ci.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           token: ${{ secrets.AVALANCHE_PAT }}
       - uses: actions/setup-go@v3
         with:
-          go-version: '~1.19.12'
+          go-version: '~1.20'
```
StephenButtolph, line 34: I think this should enforce the minimum patch version of `~1.20.8`

.github/workflows/ci.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           token: ${{ secrets.AVALANCHE_PAT }}
       - uses: actions/setup-go@v3
         with:
-          go-version: '~1.19.12'
+          go-version: '~1.20.8'
```
aaronbuchwald, line 34: In general, this should be a separate PR rather than including within a larger PR.

plugin/evm/tx_gossip_test.go
```diff
@@ -0,0 +1,254 @@
...
+	tx := types.NewTransaction(0, address, big.NewInt(10), 100_000, big.NewInt(params.LaunchMinGasPrice), nil)
+	signedTx, err := types.SignTx(tx, types.NewEIP155Signer(vm.chainID), key)
+	require.NoError(err)
+
+	errs := vm.txPool.AddLocals([]*types.Transaction{signedTx})
+	require.Len(errs, 1)
+	require.Nil(errs[0])
+
+	// wait so we aren't throttled by the vm
+	time.Sleep(throttlingPeriod / throttlingLimit)
```
aaronbuchwald, line 135: Can we replace this in the future with a direct function call that tells the throttler X amount of time has passed?

plugin/evm/vm.go
```diff
@@ -990,6 +1121,7 @@ func (vm *VM) Shutdown(context.Context) error {
 	if vm.ctx == nil {
 		return nil
 	}
+	vm.cancel()
```
aaronbuchwald, line 1124: Should we wait for everything that is canceled by this to finish cleaning up? Not necessary for this PR

plugin/evm/tx_gossip_test.go
joshua-kim: Updated it in the latest commit

plugin/evm/vm.go
joshua-kim: We already do here: https://github.com/ava-labs/coreth/blob/fbfa3726fda8ba53184cc4fef4fb82621f2328a7/plugin/evm/vm.go#L1015-L1019

aaronbuchwald (approved):

plugin/evm/vm.go
StephenButtolph: There is really no reason why this context is stored on the VM. We only need `cancel`. The context can be left as a local variable here.

plugin/evm/vm.go
```diff
@@ -990,6 +1119,7 @@ func (vm *VM) Shutdown(context.Context) error {
...
```
StephenButtolph, line 1122: I think this can be `nil` now if the VM never gets into normal operations.

darioush (approved):

aaronbuchwald:
Separate PR to bump golangci-lint timeout from 3m to 10m here: https://github.com/ava-labs/coreth/pull/330

aaronbuchwald (approved):

StephenButtolph (approved):

PR #326, opened 2023-09-08
Fix hanging requests after Shutdown

joshua-kim:
## Why this should be merged

We forget to release the requests semaphore after shutting down the p2p layer

## How this works

Adds a release if we drop an outbound request

## How this was tested

Added a regression test

### DIFF
--- a/peer/network.go
+++ b/peer/network.go
@@ -166,6 +166,7 @@ func (n *network) SendAppRequest(nodeID ids.NodeID, request []byte, responseHand
 // Assumes write lock is held
 func (n *network) sendAppRequest(nodeID ids.NodeID, request []byte, responseHandler message.ResponseHandler) error {
 	if n.closed.Get() {
+		n.activeAppRequests.Release(1)
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


darioush (approved):

StephenButtolph (approved):

peer/network.go
```diff
@@ -175,6 +175,7 @@ func (n *network) SendAppRequest(nodeID ids.NodeID, request []byte, responseHand
 // Assumes write lock is held
 func (n *network) sendAppRequest(nodeID ids.NodeID, request []byte, responseHandler message.ResponseHandler) error {
 	if n.closed.Get() {
+		n.activeAppRequests.Release(1)
```
StephenButtolph, line 178: I think we also need to release these during `Shutdown()`

StephenButtolph: Although I don't think we have a way of knowning if we should be releasing from `activeAppRequests` or `activeCrossChainRequests` there 🤔

joshua-kim: Yeah you're right we have to release on `Shutdown` as well... I think we can actually do this by just making a wrapper handler type. I would not be surprised if this was the reason we had all those weird time out EVM bugs on shutdown...

PR #328, opened 2023-09-12
Update to 1.10.10-rc.2

joshua-kim:
## Why this should be merged

Bumps go version and avalanchego dependency

## How this works

Updates go version and avalanchego version 

## How this was tested

Build passses

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -97,8 +97,6 @@ const (
 
 	// Prefixes for metrics gatherers
 	ethMetricsPrefix        = "eth"
-	sdkMetricsPrefix        = "sdk"
-	p2pMetricsPrefix        = "p2p"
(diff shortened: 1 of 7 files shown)


aaronbuchwald (approved):
LGTM once CI passes

plugin/evm/vm.go
```diff
@@ -509,7 +509,7 @@ func (vm *VM) Initialize(
 	}
 
 	// initialize peer network
-	vm.router = p2p.NewRouter(vm.ctx.Log, appSender)
+	vm.router = p2p.NewRouter(vm.ctx.Log, appSender, prometheus.NewRegistry(), "p2p")
```
StephenButtolph, line 512: I feel like we should be exposing these metrics in this PR?

StephenButtolph (approved):

PR #333, opened 2023-09-13
Add debug logs to atomic tx drops

anusha-ctrl:
## Why this should be merged
Added some debugging logs to find where/why the txs are dropping in a flaky test

https://github.com/ava-labs/avalanchego/issues/1993


aaronbuchwald (commented):
We shouldn't introduce using `[function name or other context]` to the start of log messages in this PR.

We could also reduce these logs a decent bit imo. For example, this adds a debug log before any case where `InsertTxs` would return an error, but only call `InsertTxs` from one place, so it would be more efficient to just add a single log where we call `InsertTxs`.

anusha-ctrl:
Thanks, Ill use this for debugging but cleaner version is here -> https://github.com/ava-labs/coreth/pull/335

PR #335, opened 2023-09-13
Add debug logs when we discard txs

anusha-ctrl:
## Why this should be merged
We should add debug logs whenever we discard txs
## How this was tested
Locally

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -775,7 +775,7 @@ func (vm *VM) preBatchOnFinalizeAndAssemble(header *types.Header, state *state.S
 		rules := vm.chainConfig.AvalancheRules(header.Number, header.Time)
 		if err := vm.verifyTx(tx, header.ParentHash, header.BaseFee, state, rules); err != nil {
 			// Discard the transaction from the mempool on failed verification.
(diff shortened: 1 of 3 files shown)


plugin/evm/vm.go
```diff
@@ -775,6 +775,7 @@ func (vm *VM) preBatchOnFinalizeAndAssemble(header *types.Header, state *state.S
 		rules := vm.chainConfig.AvalancheRules(header.Number, header.Time)
 		if err := vm.verifyTx(tx, header.ParentHash, header.BaseFee, state, rules); err != nil {
 			// Discard the transaction from the mempool on failed verification.
+			log.Debug("discarding tx from mempool on failed verification", "txID", tx.ID(), "error", err.Error())
```
aaronbuchwald, line 778:
```suggestion
			log.Debug("discarding tx from mempool on failed verification", "txID", tx.ID(), "err", err)
```

same suggestion for the others as well. I see the codebase is actually a bit inconsistent on "err" vs. "error" but we do not need to call `err.Error()` since the logger will stringify for us.

anusha-ctrl: thanks

darioush (approved):

aaronbuchwald (approved):

PR #346, opened 2023-09-22
Check Processing Blocks For Atomic Txs

anusha-ctrl:
## Why this should be merged

We have run into a scenario as follows: 

`NodeA` issues` blockA` with the tx
`NodeB `starts to build `blockB` on top of `blockA` with the tx
 tx fails verification due to nonce mismatch so it is discarded
`getAtomicTxStatus` returns dropped from `nodeB` for tx even though tx is processing in `blockA`

`getAtomicTx` should check for the existence of atomic txs in blocks that are not yet accepted and still in a processing state.  

## How this works

atomicBackend now tracks pendingTxs. We add to the map on `InsertTxs` and remove on `Accept` and `Reject`. Now when `getAtomicTx` is called we check the accepted atomic txs, the pending atomic txs, and then the mempool. 

## How this was tested

UTs


plugin/evm/mempool.go
```diff
@@ -189,19 +189,21 @@ func (m *Mempool) checkConflictTx(tx *Tx) (uint64, ids.ID, []*Tx, error) {
 }
 
 // addTx adds [tx] to the mempool. Assumes [m.lock] is held.
-// If [force], skips conflict checks within the mempool.
+// If [force], skips existence and conflict checks within the mempool.
```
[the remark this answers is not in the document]
anusha-ctrl: force is not used outside of testing 👍🏽 This function is used to simulate conflicting transactions. I expanded the use of `forceAddTx` to simulate the edge case we were seeing.

plugin/evm/vm.go
```diff
@@ -1427,6 +1428,10 @@ func (vm *VM) getAtomicTx(txID ids.ID) (*Tx, Status, uint64, error) {
 		return tx, Accepted, height, nil
 	} else if err != database.ErrNotFound {
 		return nil, Unknown, 0, err
+	} else if tx, height, err := vm.getAtomicTxFromProcessingBlocks(txID); err == nil {
```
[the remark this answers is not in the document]
anusha-ctrl: I can rewrite this for clarity and add comments.

plugin/evm/vm.go
```diff
@@ -1439,6 +1444,36 @@ func (vm *VM) getAtomicTx(txID ids.ID) (*Tx, Status, uint64, error) {
 	}
 }
 
+// getAtomicTxFromProcessingBlocks takes in a [txID] and searches for a matching atomic tx in processing blocks
+// Specifically we look from at blocks with lastAcceptedBlockHeight + 1  to preferredBlockHeight.
```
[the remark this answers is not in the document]
anusha-ctrl: thanks for catching that

plugin/evm/vm.go
```diff
@@ -1863,6 +1896,15 @@ func (vm *VM) getAtomicTxFromPreApricot5BlockByHeight(height uint64) (*Tx, error
 	return ExtractAtomicTx(blk.ExtData(), vm.codec)
 }
 
+func (vm *VM) getAtomicTxFromBlockByHeight(height uint64) ([]*Tx, error) {
+	blk := vm.blockChain.GetBlockByNumber(height)
+	if blk == nil {
+		return nil, nil
+	}
+	isApricotPhase5 := vm.chainConfig.IsApricotPhase5(blk.Time())
```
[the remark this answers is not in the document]
anusha-ctrl: Someone who is bootstrapping a node for example can call this function at a height prior to Apricot Phase 5.

plugin/evm/vm.go
```diff
@@ -1439,6 +1444,33 @@ func (vm *VM) getAtomicTx(txID ids.ID) (*Tx, Status, uint64, error) {
 	}
 }
 
+// getAtomicTxFromProcessingBlocks takes in a [txID] and searches for a matching atomic tx in processing blocks
+// Specifically we look from at blocks with lastAcceptedBlockHeight + 1  to preferredBlockHeight.
+// This function assumes that we have already searched accepted blocks for atomic txs.
+// We return an error if we cannot find a matching atomic tx in the processing blocks.
+func (vm *VM) getAtomicTxFromProcessingBlocks(txID ids.ID) (*Tx, uint64, error) {
```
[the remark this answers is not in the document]
anusha-ctrl: Let me look into this 👍🏽

plugin/evm/vm_test.go
```diff
@@ -810,6 +810,327 @@ func TestIssueAtomicTxs(t *testing.T) {
 	assert.Equal(t, indexedExportTx.ID(), exportTx.ID(), "expected ID of indexed import tx to match original txID")
 }
 
+func TestGetAtomicTxFromProcessingBlocksPreApricot5(t *testing.T) {
+	importAmount := uint64(1000000000)
```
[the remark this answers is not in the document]
anusha-ctrl: It's the easiest way to include an atomic tx in a block for testing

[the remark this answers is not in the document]
anusha-ctrl: Other ways include marshalling the atomic txs into bytes, creating a brand new block with all the components including header, txs, receipts, and extData. This was more error-prone in my opinion.

plugin/evm/vm_test.go
```diff
@@ -810,6 +810,490 @@ func TestIssueAtomicTxs(t *testing.T) {
...
+			test.preferredBlockHeight,
+			test.lastAcceptedBlockHeight,
+			func(height uint64) bool { return test.isApricotPhase5[height] },
+			func(height uint64) *types.Block { return test.blocks[height] },
+			vm.codec)
+
+		if test.expectedErr != nil {
+			require.Equal(t, test.expectedErr, err)
+			continue
+		}
```
darioush, line 960: this doesn't check `err == nil` in the case where `test.ExpectedErr == nil`.
Could we do:
```suggestion
		require.ErrorIs(t, test.expectedErr, err)
		if test.expectedErr != nil {
			require.Nil(t, tx)
			require.Zero(t, height)
			continue
		}
```

plugin/evm/vm.go
```diff
@@ -1439,6 +1454,50 @@ func (vm *VM) getAtomicTx(txID ids.ID) (*Tx, Status, uint64, error) {
...
+func getAtomicTxFromProcessingBlocks(
+	txID ids.ID,
+	preferredBlockHeight uint64,
+	lastAcceptedBlockHeight uint64,
+	isApricotPhase5 func(uint64) bool,
+	getBlockByNumber func(uint64) *types.Block,
+	codec codec.Manager) (*Tx, uint64, error) {
+	for processingBlockHeight := lastAcceptedBlockHeight + 1; processingBlockHeight <= preferredBlockHeight; processingBlockHeight++ {
+		// We know that we have some blocks processing that have not yet been accepted that might potentially have the atomic tx
+		blk := getBlockByNumber(processingBlockHeight)
```
darioush, line 1481: does this need a nil check?

plugin/evm/vm_test.go
```diff
@@ -810,6 +810,490 @@ func TestIssueAtomicTxs(t *testing.T) {
...
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(logs) != 0 {
+		t.Fatalf("Expected log length to be 0, but found %d", len(logs))
+	}
+	if logs == nil {
+		t.Fatal("Expected logs to be non-nil")
+	}
```
darioush, line 1239: what is this for?

plugin/evm/vm.go
```diff
@@ -1439,6 +1454,50 @@ func (vm *VM) getAtomicTx(txID ids.ID) (*Tx, Status, uint64, error) {
...
+		txID,
+		vm.blockChain.CurrentBlock().Number.Uint64(),
+		vm.blockChain.LastAcceptedBlock().Header().Number.Uint64(),
+		vm.chainConfig.IsApricotPhase5,
+		vm.blockChain.GetBlockByNumber,
+		vm.codec,
+	)
+}
+
+func getAtomicTxFromProcessingBlocks(
```
darioush, line 1472: an alternate implementation idea would be tracking the pending txs in the atomic backend (eg in a map), then removing them when the block is accepted/rejected.
https://github.com/ava-labs/coreth/blob/master/plugin/evm/atomic_backend.go#L364
https://github.com/ava-labs/coreth/blob/master/plugin/evm/atomic_state.go#L49
https://github.com/ava-labs/coreth/blob/master/plugin/evm/atomic_state.go#L91

not sure if it is simpler

plugin/evm/vm_test.go
```diff
@@ -810,6 +810,490 @@ func TestIssueAtomicTxs(t *testing.T) {
...
+			require.Equal(t, test.expectedErr, err)
+			continue
+		}
+
+		require.Equal(t, test.expectedTx.ID(), tx.ID())
+		require.Equal(t, test.expectedHeight, height)
+	}
+}
+
+func TestGetAtomicTxFromProcessingBlocksPostApricot5(t *testing.T) {
```
darioush, line 967: Could we merge this test with pre apricot phase 5? (in a different test case)

plugin/evm/vm_test.go
```diff
@@ -810,6 +810,490 @@ func TestIssueAtomicTxs(t *testing.T) {
...
+	if status := blk3.Status(); status != choices.Processing {
+		t.Fatalf("Expected status of built block to be %s, but found %s", choices.Processing, status)
+	}
+
+	if err := vm.SetPreference(context.Background(), blk2.ID()); err != nil {
+		t.Fatal(err)
+	}
+
+	// getAtomicTx should check the processing blocks for any atomic txs.
+	exportTxFetched, status, height, err := vm.getAtomicTx(exportTx.ID())
```
darioush, line 1290: Could we add this (integration) test into the test for TestIssueAtomicTxs instead of duplicating the setup?
I think it suffices alongside the other UT.
We can add a check before accepting the block.

plugin/evm/vm.go
```diff
@@ -1439,6 +1454,50 @@ func (vm *VM) getAtomicTx(txID ids.ID) (*Tx, Status, uint64, error) {
...
 
+// getAtomicTxFromProcessingBlocks takes in a [txID] and searches for a matching atomic tx in processing blocks
+// Specifically we look from blocks with lastAcceptedBlockHeight + 1  to preferredBlockHeight.
+// This function assumes that we have already searched accepted blocks for atomic txs.
+// We return an error if we cannot find a matching atomic tx in the processing blocks.
+func (vm *VM) getAtomicTxFromProcessingBlocks(txID ids.ID) (*Tx, uint64, error) {
+	return getAtomicTxFromProcessingBlocks(
+		txID,
+		vm.blockChain.CurrentBlock().Number.Uint64(),
+		vm.blockChain.LastAcceptedBlock().Header().Number.Uint64(),
```
darioush, line 1465: I think we need to use the vm (plugin/evm) LastAccepted block, since the chain accepts block asynchronously and the VM may have accepted a block the chain has not accepted yet.

plugin/evm/vm.go
anusha-ctrl: yes thanks

plugin/evm/vm_test.go
anusha-ctrl: thanks

plugin/evm/vm_test.go
anusha-ctrl: Removed thanks

plugin/evm/vm_test.go
anusha-ctrl: done

plugin/evm/vm_test.go
anusha-ctrl: Yes

plugin/evm/vm.go
anusha-ctrl: Thanks for the suggestion! I chose to iterate through the blocks [last accepted +1, preferred block] since it was obvious to me to look through those as the processing blocks and there were preexisting functions to check if those blocks have atomic txs. I'm looking more at the atomic backend since this looks good too, I see that `InsertTxs` is called in `onExtraStateChange` which is called in `Finalize`. Where is this being replicated in `FinalizeAndAssemble`?

plugin/evm/vm.go
anusha-ctrl: 👀

plugin/evm/vm.go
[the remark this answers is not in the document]
anusha-ctrl: thanks done

plugin/evm/vm_test.go
darioush: We can use the same genesis and activate the upgrade at a timestamp (if you want to build both pre- and post- upgrade blocks)

plugin/evm/vm.go
darioush: >  I suppose if acceptorQueue is backlogged.
yes this is correct. the vm last accepted height is >= the chain last accepted height, due to the possible lag in the async queue

anusha-ctrl: Thanks using the vm's

plugin/evm/vm_test.go
anusha-ctrl: Oh sweet that works. Let me merge these

plugin/evm/vm_test.go
```diff
@@ -810,6 +810,271 @@ func TestIssueAtomicTxs(t *testing.T) {
...
 
+func TestGetAtomicTxsInProcessingBlocks(t *testing.T) {
+	importAmount := uint64(50000000)
+	issuer, vm, _, _, _ := GenesisVMWithUTXOs(t, true, genesisJSONApricotPhase2, "", "", map[ids.ShortID]uint64{
+		testShortIDAddrs[0]: importAmount,
+	})
+
+	defer func() {
+		if err := vm.Shutdown(context.Background()); err != nil {
+			t.Fatal(err)
```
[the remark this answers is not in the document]
anusha-ctrl: done

plugin/evm/vm.go
```diff
@@ -1439,6 +1455,59 @@ func (vm *VM) getAtomicTx(txID ids.ID) (*Tx, Status, uint64, error) {
 	}
 }
 
+// getAtomicTxFromProcessingBlocks takes in a [txID] and searches for a matching atomic tx in processing blocks
+// Specifically we look from blocks with lastAcceptedBlockHeight + 1  to preferredBlockHeight.
```
darioush, line 1459:
```suggestion
// Specifically we look from blocks with lastAcceptedBlockHeight + 1 to preferredBlockHeight.
```

plugin/evm/vm.go
```diff
@@ -1439,6 +1455,59 @@ func (vm *VM) getAtomicTx(txID ids.ID) (*Tx, Status, uint64, error) {
...
+// This function assumes that we have already searched accepted blocks for atomic txs.
+// We return an error if we cannot find a matching atomic tx in the processing blocks.
```
darioush, line 1461:
```suggestion
// We return errNoAtomicTxsFound if we cannot find a matching atomic tx in the processing blocks.
```

plugin/evm/vm_test.go
```diff
@@ -748,20 +739,12 @@ func TestIssueAtomicTxs(t *testing.T) {
...
-	}
-	if len(logs) != 0 {
-		t.Fatalf("Expected log length to be 0, but found %d", len(logs))
-	}
-	if logs == nil {
-		t.Fatal("Expected logs to be non-nil")
-	}
+	require.NoError(err)
+	require.Empty(logs, "Expected logs to have length 0")
+	require.NotNil(logs, "Expected logs to be non-nil")
```
darioush, line 744: I don't think we need to verify the logs

plugin/evm/vm_test.go
```diff
@@ -770,44 +753,357 @@ func TestIssueAtomicTxs(t *testing.T) {
...
-	}
+	require.Equal(blk2.Status(), choices.Processing, "Expected status to match")
 
-	if err := blk2.Accept(context.Background()); err != nil {
-		t.Fatal(err)
-	}
+	err = blk2.Accept(context.Background())
+	require.NoError(err)
 
 	if status := blk2.Status(); status != choices.Accepted {
```
darioush, line 766: nit: can also be updated to `require.Equal`

plugin/evm/vm_test.go
```diff
@@ -770,44 +753,357 @@ func TestIssueAtomicTxs(t *testing.T) {
...
+	vm.clock.Set(vm.clock.Time().Add(2 * time.Second))
+	blk4, err := vm.BuildBlock(context.Background())
+	require.NoError(err)
+
+	noAtomicTxBlockPostApricot5 := blk4.(*chain.BlockWrapper).Block.(*Block)
+
+	// Create a post apricot5 block that has multiple atomic txs
+	kc := secp256k1fx.NewKeychain()
+	kc.Add(testKeys[0])
+	txID, err := ids.ToID(hashing.ComputeHash256(testShortIDAddrs[0][:]))
```
darioush, line 942: not sure I follow why this is called txID?

plugin/evm/vm_test.go
```diff
@@ -770,44 +753,357 @@ func TestIssueAtomicTxs(t *testing.T) {
...
+		expectedErr             error
+	}
+
+	testCases := map[string]testGetAtomicTx{
+		"happy path": {
+			lastAcceptedBlockHeight: 1,
+			preferredBlockHeight:    2,
+			blocks:                  map[uint64]*types.Block{2: atomicTxBlock.ethBlock},
+			isApricotPhase5:         map[uint64]bool{2: false},
+			inputTxID:               importTx.ID(),
```
darioush, line 981: seems we could slightly simplify by passing the inputTx and verifying it is returned as `expectedTx` if the `expectedErr` is nil. as is also fine, more explicit.

darioush (approved):
Approved with some nits. I think this works since:
Tx will be removed in mempool on accept. Before that, the block should be added as verified.

There can be some test cleanup but I think we can also do that more broadly in the vm_test file instead of being too picky now.

plugin/evm/vm.go
darioush: https://github.com/ava-labs/coreth/blob/37c9cc40b2bc8a4bc72906c4091a8839a888d4f7/plugin/evm/vm.go#L941

darioush: It's not, since the block will be verified again after it's built.
We used to do this when we were planning to include the root of the atomic trie in the header https://github.com/ava-labs/coreth/commit/7b260fcf708cbf98c772dfa5575c67aef975f6d2

plugin/evm/vm.go
anusha-ctrl: thanks

plugin/evm/vm.go
anusha-ctrl: thanks

plugin/evm/vm_test.go
anusha-ctrl: yeah, these were copied from the existing test, removing.

plugin/evm/vm_test.go
anusha-ctrl: copied from another test, changing name

anusha-ctrl: I believe this is called txID to match the input name string for `addUtxo` . I'll keep this name.

plugin/evm/vm_test.go
anusha-ctrl: hey thanks. this was cleaner. changed it

plugin/evm/vm.go
aaronbuchwald: Can we switch to @darioush 's suggestion of tracking the processing transactions in the atomic backend?

This will borrow from an existing abstraction where tracking the processing transactions is very easy as opposed to adding an additional getter to maintain for this one-off case.

plugin/evm/vm_test.go
```diff
@@ -688,80 +687,53 @@ func TestImportMissingUTXOs(t *testing.T) {
...
-	})
-	if err != nil {
-		t.Fatal(err)
-	}
-	if len(logs) != 0 {
-		t.Fatalf("Expected log length to be 0, but found %d", len(logs))
-	}
-	if logs == nil {
-		t.Fatal("Expected logs to be non-nil")
-	}
```
aaronbuchwald, line 759: Why is this removed in this PR?

plugin/evm/vm_test.go
```diff
@@ -770,44 +742,343 @@ func TestIssueAtomicTxs(t *testing.T) {
...
+
+	err = vm.issueTx(exportTx2, true /*=local*/)
+	require.NoError(err)
+
+	<-issuer
+
+	blk3, err := vm.BuildBlock(context.Background())
+	require.NoError(err)
+
+	err = vm.mempool.ForceAddTx(exportTx2)
```
aaronbuchwald, line 774: Could we create a separate test for this instead of adding to the existing one? Alternatively, can we get rid of the modification to this test in favor of just adding the regression test below?

plugin/evm/vm_test.go
```diff
@@ -770,44 +742,343 @@ func TestIssueAtomicTxs(t *testing.T) {
...
-		t.Fatal(err)
+	<-issuer
+
+	blk2, err := vm.BuildBlock(context.Background())
+	require.NoError(err)
+
+	noAtomicTxBlock := blk2.(*chain.BlockWrapper).Block.(*Block)
+
+	// All new blocks will be post apricot5
+	vm.chainConfig.ApricotPhase5BlockTimestamp = utils.NewUint64(0)
```
aaronbuchwald, line 884: It's invalid to modify the rules after we've already started accepting blocks. Can we switch to adding a single test that checks whether after attempting to issue a transaction that is already included in a processing block, the status of that transaction shows up as processing correctly?

plugin/evm/vm_test.go
aaronbuchwald: Looks like this is copied from addUTXO at the start of a test where it could actually just be a randomly generated UTXOID instead. Can we switch to using `ids.GenerateTestID()` here and fix it above as well instead of copying my old mistake 😅

aaronbuchwald (changes requested):
Can we switch from the current implementation strategy of:

1. Adding an extra getter function to iterate blocks for processing atomic transactions
2. Adding an extra check when retrieving the transaction's status to check if it's in processing before checking the status as reported by the mempool and returning `Processing` early if found

To

1. Track the set of processing transactions using the atomic backend (there are already insert/accept/reject functions we can add to here)
2. Modify `onFinalizeAndAssemble` to check if a transaction is already included in processing blocks before verification fails and it is marked as dropped.
3. If during `onFinalizeAndAssemble` a transaction is seen as a duplicate of a processing transaction, mark it as issued.

This fixes the root cause of the problem, which is that we are marking a transaction that is currently in processing as dropped in the mempool, whereas the current implementation would still have the processing transaction included in the dropped transactions cache and make sure to skip it when fetching a transaction's status.

plugin/evm/vm_test.go
anusha-ctrl: Wasn't sure why we were testing for logs in an `atomictx` test.

plugin/evm/vm_test.go
anusha-ctrl: I originally kept this as a separate test but merged it based on other's feedback as it tests the same function. It should be able to test the accepted case as well as the processing case. Is there a reason that the processing test should be kept separate? I see this as a more complete e2e test whereas the other test just tests different input cases.

plugin/evm/vm_test.go
anusha-ctrl: I see. Should I have two separate tests for pre/post apricot 5? I know you mentioned that we don't need to necessarily support pre apricot 5. I think its worth to test all different input cases here.

anusha-ctrl:
> Can we switch from the current implementation strategy of:
> 
> 1. Adding an extra getter function to iterate blocks for processing atomic transactions
> 2. Adding an extra check when retrieving the transaction's status to check if it's in processing before checking the status as reported by the mempool and returning `Processing` early if found
> 
> To
> 
> 1. Track the set of processing transactions using the atomic backend (there are already insert/accept/reject functions we can add to here)
> 2. Modify `onFinalizeAndAssemble` to check if a transaction is already included in processing blocks before verification fails and it is marked as dropped.
> 3. If during `onFinalizeAndAssemble` a transaction is seen as a duplicate of a processing transaction, mark it as issued.
> 
> This fixes the root cause of the problem, which is that we are marking a transaction that is currently in processing as dropped in the mempool, whereas the current implementation would still have the processing transaction included in the dropped transactions cache and make sure to skip it when fetching a transaction's status.

Replied offline- I think it's ultimately better to go with this approach as the above implementation will still report a  dropped transaction in the edge case.

plugin/evm/vm.go
```diff
@@ -1423,11 +1424,23 @@ func (vm *VM) conflicts(inputs set.Set[ids.ID], ancestor *Block) error {
...
 		return tx, Accepted, height, nil
 	} else if err != database.ErrNotFound {
 		return nil, Unknown, 0, err
 	}
+
+	// Check processing blocks for atomic txs
+	// We should do this check here because blocks can pick up these atomic txs
+	// prior to block acceptance and should have status as processing.
+	// These txs will be available in the mempool as well but could give misleading
+	// results if dropped by another processing block.
```
darioush, line 1438:
```suggestion
	// We check the atomic backend here because processing blocks can
	// contain atomic txs and their status report as processing.
	// This takes precedence over the tx's mempool status, where
	// the status may be dropped (if it was included in a processing block).
```

plugin/evm/vm_test.go
darioush: This was added here https://github.com/ava-labs/coreth/pull/270/files (best to consult the git blame) in cases like this.
Could we add a comment, or a helper function like `verifyNoNilLogRegression(...)` here and adopt the require style?

plugin/evm/vm_test.go
```diff
@@ -770,44 +742,60 @@ func TestIssueAtomicTxs(t *testing.T) {
...
+	require.NoError(err)
+
+	require.Equal(blk3.Status(), choices.Processing, "Expected status to match")
+
+	// getAtomicTx should check the processing blocks for any atomic txs.
+	exportTxFetched, status, height, err := vm.getAtomicTx(exportTx2.ID())
+	require.NoError(err)
+	require.Equal(Processing, status)
+	require.Equal(uint64(3), height, "expected height of export tx to be 3")
+	require.Equal(exportTxFetched.ID(), exportTx2.ID(), "expected ID of fetched export tx to match original txID")
```
darioush, line 798: should we further simplify the test by checking the status of `exportTx` before blk2 is accepted instead of creating blk3?

plugin/evm/vm.go
anusha-ctrl: thanks

anusha-ctrl: thanks

darioush (approved):

aaronbuchwald (changes requested):
This is not a full review since we need to fix a bug in this PR to reference count transactions across processing blocks.

plugin/evm/atomic_backend.go
```diff
@@ -38,6 +38,11 @@ type AtomicBackend interface {
 	// but not Accepted or Rejected yet.
 	GetVerifiedAtomicState(blockHash common.Hash) (AtomicState, error)
 
+	// GetPendingTx retrieves an atomic transaction and its corresponding block height for a
+	// given txID that has been inserted but has not yet been accepted or rejected.
+	// This transaction is found in a 'processing' or 'undecided' block.
+	GetPendingTx(txID ids.ID) (*Tx, uint64, error)
```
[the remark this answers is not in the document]
darioush: good catch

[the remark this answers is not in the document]
anusha-ctrl: can you repeat the above example with proposervm in consideration (including setting a block as preferred). I'm still a bit confused.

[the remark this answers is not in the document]
aaronbuchwald: `BuildBlock` is defined by this codebase, but does not stop a different node from building a block in a different way. If we do not enforce a condition when verifying the block, we cannot rely on other nodes building blocks according to the logic defined in `BuildBlock`.

The ProposerVM and setting the preference does not change anything here. The ProposerVM is a soft proposer mechanism that determines who should build the next block, but that proposer could still issue multiple different blocks at that slot that we need to handle. Additionally, after the proposer issues a block, the next proposer could issue a competing block after the proposer window expands and before the first proposer's block is accepted. The result is that the ProposerVM does not change in any way the insertion orders that the VM needs to handle correctly.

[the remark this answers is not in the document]
aaronbuchwald: You cannot assume anything about how a different node will behave in a byzantine environment.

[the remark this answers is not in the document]
anusha-ctrl: Ok, pendingTxs now references all the blocks its associated with.

plugin/evm/atomic_state.go
```diff
@@ -91,6 +95,10 @@ func (a *atomicState) Accept(commitBatch database.Batch) error {
 func (a *atomicState) Reject() error {
 	// Remove the block from the map of undecided blocks.
 	delete(a.backend.verifiedRoots, a.blockHash)
+	// Removes pendingTx from the pendingTx map
+	for _, tx := range a.txs {
+		delete(a.backend.pendingTxs, tx.ID())
```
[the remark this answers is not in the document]
anusha-ctrl: noted!

plugin/evm/vm_test.go
```diff
@@ -808,6 +815,137 @@ func TestIssueAtomicTxs(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	<-issuer
+
+	// blk4 also picks up the same atomic tx, fails block building, and should discard the atomic tx
+	blk4, _ := vm.BuildBlock(context.Background())
+	if blk4 != nil {
+		t.Fatalf("blk4 should have failed verification since it holds the same tx as the preferred block")
+	}
```
aaronbuchwald, line 859: We should check that the error is non-nil instead of checking that the block is nil

plugin/evm/vm_test.go
```diff
@@ -808,6 +815,137 @@ func TestIssueAtomicTxs(t *testing.T) {
...
 	assert.Equal(t, uint64(2), height, "expected height of indexed export tx to be 2")
 	assert.Equal(t, indexedExportTx.ID(), exportTx.ID(), "expected ID of indexed import tx to match original txID")
+
+	// The same atomic transaction is included in multiple processing blocks.
+	// It gets dropped in one block and is still processing in the other.
+	// getAtomicTx should return status Processing
+	exportTx2, err := vm.newExportTx(vm.ctx.AVAXAssetID, 1, vm.ctx.XChainID, testShortIDAddrs[0], initialBaseFee, []*secp256k1.PrivateKey{testKeys[0]})
+	if err != nil {
+		t.Fatal(err)
+	}
```
aaronbuchwald, line 825: Can we separate the flow from this point on into its own test? This test is already pretty big, and I'd prefer to separate the basic flow test from the regression test for this exact behavior.

aaronbuchwald (changes requested):

plugin/evm/atomic_state.go
```diff
@@ -91,6 +95,28 @@ func (a *atomicState) Accept(commitBatch database.Batch) error {
...
+	}
+
+	if len(pendingTx.blockMap) == 1 {
+		// There is only one processing block that holds this atomic tx so we can delete the entry from the map
+		delete(a.backend.pendingTxs, txID)
+		return
+	}
+	// There are multiple processing blocks that hold this atomic tx so we can delete specific block from the blockMap.
+	delete(pendingTx.blockMap, blockHash)
+}
```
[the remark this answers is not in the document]
anusha-ctrl: done

plugin/evm/vm_test.go
anusha-ctrl: done

plugin/evm/vm_test.go
anusha-ctrl: 👍🏽

plugin/evm/vm_test.go
```diff
@@ -810,6 +817,192 @@ func TestIssueAtomicTxs(t *testing.T) {
...
+
+	if lastAcceptedID, err := vm.LastAccepted(context.Background()); err != nil {
+		t.Fatal(err)
+	} else if lastAcceptedID != blk.ID() {
+		t.Fatalf("Expected last accepted blockID to be the accepted block: %s, but found %s", blk.ID(), lastAcceptedID)
+	}
+
+	// The same atomic transaction is included in multiple processing blocks.
+	// It gets dropped in one block and is still processing in the other.
+	// getAtomicTx should return status Processing
```
aaronbuchwald, line 876:
```suggestion
	// An atomic transaction already included in a processing block is dropped during
	// block building. It should still report status processing after BuildBlock fails.
```

plugin/evm/vm_test.go
```diff
@@ -810,6 +817,192 @@ func TestIssueAtomicTxs(t *testing.T) {
...
+	}
+
+	if status := blk2.Status(); status != choices.Accepted {
+		t.Fatalf("Expected status of accepted block to be %s, but found %s", choices.Accepted, status)
+	}
+
+	if lastAcceptedID, err := vm.LastAccepted(context.Background()); err != nil {
+		t.Fatal(err)
+	} else if lastAcceptedID != blk2.ID() {
+		t.Fatalf("Expected last accepted blockID to be the accepted block: %s, but found %s", blk2.ID(), lastAcceptedID)
```
aaronbuchwald, line 939: can we add a check that `exportTx` is marked as accepted with a correctly indexed height here?

plugin/evm/vm_test.go
```diff
@@ -810,6 +817,192 @@ func TestIssueAtomicTxs(t *testing.T) {
...
+	}
+
+	// The same atomic transaction is included in multiple processing blocks that were built at around the same time.
+	// It gets rejected in one block and is still processing in the other.
+	// getAtomicTx should return status Processing
```
aaronbuchwald, line 944:
```suggestion
	// The same atomic transaction is included in multiple processing blocks. After
	// rejecting one processing block, the atomic tx should still report status processing
	// while there is still a processing block containing the atomic transaction.
```

plugin/evm/vm_test.go
```diff
@@ -810,6 +817,192 @@ func TestIssueAtomicTxs(t *testing.T) {
...
+	if err := blk3.Reject(context.Background()); err != nil {
+		t.Fatal(err)
+	}
+
+	// getAtomicTx should still return status Processing, even though blk3 got rejected with exportTx2, since blk4 is still processing with exportTx2.
+	exportTxFetched, status, height, err = vm.getAtomicTx(exportTx2.ID())
+	assert.NoError(t, err)
+	assert.Equal(t, Processing, status)
+	assert.Equal(t, uint64(3), height, "expected height of export tx to be 3")
+	assert.Equal(t, exportTxFetched.ID(), exportTx2.ID(), "expected ID of fetched export tx to match original txID")
```
aaronbuchwald, line 1003: Can we add a final check here to accept `blk4` and check the status of `exportTx2` is correctly indexed and marked as accepted?

aaronbuchwald (changes requested):

plugin/evm/vm_test.go
anusha-ctrl: done

plugin/evm/vm_test.go
anusha-ctrl: yup

plugin/evm/vm_test.go
anusha-ctrl: thanks

plugin/evm/vm_test.go
anusha-ctrl: done

PR #353, opened 2023-10-05
State sync 1 - Blocks backfilling option

abi87:
## Why this should be merged
Counterpart of State sync - Blocks backfilling option

## How this works
Missing stuff:

- [X] UTs
- [X] Add ETA to complete backfilling
- [X] What if state sync is re-run again before block backfilling completes? Current implementation would redownload a bunch of already downloaded blocks. Avoid that

## How this was tested
Extra UTs  + Fuji run (ongoing)


plugin/evm/syncervm_client.go
```diff
@@ -317,6 +333,130 @@ func (client *stateSyncerClient) syncStateTrie(ctx context.Context) error {
...
+			log.Warn(
+				"Unexpected backfilled block. Reissuing request",
+				"expected block ID", topBlk.Parent(),
+				"received block ID", blk.ID(),
+				"Requested block ID", topBlk.Parent(),
+			)
+			break
+		}
+
+		if err := blk.Accept(ctx); err != nil {
```
abi87, line 426: This is most likely wrong. Accepting a past block seems to mess up a whole bunch of metadata.
@darioush  I'd love a feedback on how the indexing of past block should be handled (brand new methods? Upgrade block.Accept to handle past blocks? something else?).

PR #354, opened 2023-10-09
Remove lock options from http handlers

StephenButtolph:
## Why this should be merged

See: [#2148](https://github.com/ava-labs/avalanchego/pull/2148)

## How this works

Explicitly grabs the context lock where it would have implicitly been grabbed before. There were a _couple_ APIs that I felt obviously didn't require a lock. But this should be a ~noop change.

## How this was tested

CI

### DIFF
--- a/plugin/evm/service.go
+++ b/plugin/evm/service.go
@@ -106,15 +106,15 @@ type ExportKeyReply struct {
 
 // ExportKey returns a private key from the provided user
 func (service *AvaxAPI) ExportKey(r *http.Request, args *ExportKeyArgs, reply *ExportKeyReply) error {
-	service.vm.ctx.Lock.Lock()
-	defer service.vm.ctx.Lock.Unlock()
-
 	log.Info("EVM: ExportKey called")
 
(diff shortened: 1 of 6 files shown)


aaronbuchwald (dismissed):

ceyonur (approved):

aaronbuchwald (approved):

plugin/evm/service.go
```diff
@@ -86,7 +86,7 @@ type VersionReply struct {
 }
 
 // ClientVersion returns the version of the VM running
-func (service *AvaxAPI) Version(r *http.Request, args *struct{}, reply *VersionReply) error {
+func (service *AvaxAPI) Version(r *http.Request, _ *struct{}, reply *VersionReply) error {
```
abi87, line 89: q: why didn't linter catch this?

abi87 (approved):

plugin/evm/service.go
StephenButtolph: I don't think coreth uses `revive` (which is what would catch this in avalanchego)

Issue #355, opened 2023-10-11
Couldn't build abigen binary


christiankakesa:
**Describe the bug**
When building `cmd/abigen/main.go`, 2 errors occurs: `undefined: nameFilter` and `undefined: newNamefilter`.

**To Reproduce**
Steps to reproduce the behavior.

**Expected behavior**
No error and `abigen` binary created.

**Screenshots**

    christian@MSI:~/workspace/ava-labs-coreth/cmd/abigen$ go mod tidy
    christian@MSI:~/workspace/ava-labs-coreth/cmd/abigen$ go build -o abigen-avax main.go
    # command-line-arguments
    ./main.go:162:16: undefined: nameFilter
    ./main.go:165:22: undefined: newNameFilter

**Operating System**
GNU/linux Ubuntu 20.04.

Avalanche Bug Bounty program can be found [here](https://hackenproof.com/avalanche/avalanche-protocol).

ceyonur:
It should be built from the root directory (where the `go.mod` file resides):
`christian@MSI:~/workspace/ava-labs-coreth$ go build -o ./abigen-avax cmd/abigen/*.go`

christiankakesa:
Yes of course but many documentation mention only the way to do that by compiling directly main.go.
Take a look at this outdated documention: https://docs.avax.network/build/dapp/smart-contracts/abigen.

By the way it's OK for me.

ceyonur:
Opened PR https://github.com/ava-labs/avalanche-docs/pull/1549/
Thanks for reporting this @christiankakesa.

PR #356, opened 2023-10-11
fix Couldn't build abigen binary #355

christiankakesa:
This change fixes errors occured when compiling cmd/abigen/main.go.

Now devs could continue generate go code from Solidity code or ABI file.


ceyonur:
This should not require any change in the pkg structure. 
See: https://github.com/ava-labs/coreth/issues/355#issuecomment-1757271974

PR #363, opened 2023-10-25
Migrating Push Gossip to SDK handlers

joshua-kim:
## Why this should be merged

Continuation on our work to standardize our p2p logic, this moves all of our push gossip workflow into SDK handlers.

## How this works

Adds handlers for push gossip into the SDK's router. Following the next network upgrade, we'll deprecate the old gossip gossip sent without the SDK.

## How this was tested

Added a unit test


plugin/evm/tx_gossip_test.go
```diff
@@ -5,250 +5,308 @@ package evm
...
-	"go.uber.org/mock/gomock"
-
-	"google.golang.org/protobuf/proto"
-
 	"github.com/ava-labs/coreth/core"
 	"github.com/ava-labs/coreth/core/types"
 	"github.com/ava-labs/coreth/params"
 )
 
 func TestEthTxGossip(t *testing.T) {
```
joshua-kim, line 30: The old tests were unbelievably confusing to follow. The new tests just set up two vms, issue a tx to one of them, and wait until we see it.

plugin/evm/vm_test.go
```diff
@@ -124,6 +124,69 @@ func init() {
...
+		}
+	}
+
+	genesis := newGenesis()
+	genesis.Alloc = alloc
+
+	return genesis
+}
+
+func newGenesis() *core.Genesis {
```
joshua-kim, line 144: I feel like the way we're testing the VM by passing in a string-ed json is confusing.... added some code to just pass around a genesis struct instead + initialize a funded genesis instead of imports w/ shared memory.

plugin/evm/gossiper_atomic_gossiping_test.go
```diff
@@ -11,75 +11,12 @@ import (
...
 	"github.com/ava-labs/avalanchego/ids"
-	"github.com/ava-labs/avalanchego/snow"
 	"github.com/ava-labs/avalanchego/utils/set"
 	"github.com/stretchr/testify/assert"
 
 	"github.com/ava-labs/coreth/plugin/evm/message"
 )
 
-// locally issued txs should be gossiped
-func TestMempoolAtmTxsIssueTxAndGossiping(t *testing.T) {
```
joshua-kim, line 22: This seems to test the same behavior as the new atomic tx test so it's deleted.

plugin/evm/vm.go
```diff
@@ -1,4 +1,3 @@
-// (c) 2019-2020, Ava Labs, Inc. All rights reserved.
```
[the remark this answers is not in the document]
joshua-kim: oops :<

plugin/evm/vm_test.go
```diff
@@ -124,6 +118,69 @@ func init() {
...
+			ApricotPhase4BlockTimestamp:     new(uint64),
+			ApricotPhase5BlockTimestamp:     new(uint64),
+			ApricotPhasePre6BlockTimestamp:  new(uint64),
+			ApricotPhase6BlockTimestamp:     new(uint64),
+			ApricotPhasePost6BlockTimestamp: new(uint64),
+			BanffBlockTimestamp:             new(uint64),
+			CortinaBlockTimestamp:           new(uint64),
+			DUpgradeBlockTimestamp:          new(uint64),
+			CancunTime:                      new(uint64),
+		},
```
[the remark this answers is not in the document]
joshua-kim: Hmm I didn't know this package existed, I think that makes sense to use. I think it would still be a good idea to have a function that creates these configs per test as opposed to having a global variable in case tests want to override a default without affecting the state of that config for other tests.

plugin/evm/vm.go
```diff
@@ -1023,92 +1019,127 @@ func (vm *VM) initBlockBuilding() error {
...
+	gossipStats := NewGossipStats()
+	vm.gossipHandler = NewGossipHandler(vm, gossipStats)
+
+	ethTxPushGossipHandler := ethTxPushGossipHandler{
+		NoOpHandler:   p2p.NoOpHandler{},
+		GossipHandler: vm.gossipHandler,
+		codec:         vm.networkCodec,
+	}
+
+	ethTxPushGossipClient, err := vm.router.RegisterAppProtocol(ethTxPushGossipProtocol, ethTxPushGossipHandler, nil)
```
aaronbuchwald, line 1122: Can we pass in a non-nil sampler here? Looking at the p2p SDK it looks like this will panic if we ever send an `AppRequestAny` on this type.

We should also improve the documentation around this if the client is going to pass in a sampler, since the documentation for `AppRequestAny` and `RegisterAppProtocol` do not mention that this combination could cause a panic.

plugin/evm/vm_test.go
```diff
@@ -4092,3 +4104,282 @@ func TestSkipChainConfigCheckCompatible(t *testing.T) {
...
+			require.False(vm2.txPool.Has(signedTx.Hash()))
+
+			require.NoError(vm1.txPool.AddRemote(signedTx))
+			require.True(vm1.txPool.Has(signedTx.Hash()))
+
+			update := make(chan core.NewTxsEvent)
+			vm2.txPool.SubscribeNewTxsEvent(update)
+			<-update
+
+			require.True(vm2.txPool.Has(signedTx.Hash()))
```
aaronbuchwald, line 4219: ❤️ this is great

aaronbuchwald: Not necessary for this PR, but it would be great to factor out this structure from these two tests so that we can easily re-use this to test vm2vm networking behaviors.

aaronbuchwald (dismissed):

plugin/evm/vm.go
```diff
@@ -1023,92 +1019,127 @@ func (vm *VM) initBlockBuilding() error {
...
+	if err != nil {
+		return err
+	}
+
+	atomicTxPushGossipHandler := atomicTxPushGossipHandler{
+		NoOpHandler:   p2p.NoOpHandler{},
+		GossipHandler: vm.gossipHandler,
+		codec:         vm.networkCodec,
+	}
```
ceyonur, line 1131: Seems there are bunch of duplicated code. I wonder if we can combine them into a single function that works with interfaces and take handlers as parameter to return client. I.e initGossiper(mempool gossip.gossippable, protocol, gossip.config...)

plugin/evm/vm.go
```diff
@@ -1023,92 +1019,127 @@ func (vm *VM) initBlockBuilding() error {
...
+
+	atomicTxPushGossipClient, err := vm.router.RegisterAppProtocol(atomicTxPushGossipProtocol, atomicTxPushGossipHandler, nil)
+	if err != nil {
+		return err
+	}
+
+	vm.gossiper = vm.createGossiper(gossipStats, ethTxPushGossipClient, atomicTxPushGossipClient)
+	vm.builder = vm.NewBlockBuilder(vm.toEngine)
+	vm.builder.awaitSubmittedTxs()
+	vm.Network.SetGossipHandler(vm.gossipHandler)
```
ceyonur, line 1141: @aaronbuchwald do you think we should first set the handler then call `awaitSubmittedTxs`?
This is unrelated with this PR.

plugin/evm/vm.go
joshua-kim: Fixed w/ the new refactor

plugin/evm/vm.go
```diff
@@ -1023,92 +1019,127 @@ func (vm *VM) initBlockBuilding() error {
...
-		gossip.Every(ctx, vm.ctx.Log, atomicTxGossiper, gossipFrequency)
+		gossip.Every(ctx, vm.ctx.Log, atomicTxPullGossiper, gossipFrequency)
 		vm.shutdownWg.Done()
 	}()
 
+	gossipStats := NewGossipStats()
+	vm.gossipHandler = NewGossipHandler(vm, gossipStats)
+
+	ethTxPushGossipHandler := ethTxPushGossipHandler{
+		NoOpHandler:   p2p.NoOpHandler{},
```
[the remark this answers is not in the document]
joshua-kim: Added

plugin/evm/gossiper.go
```diff
@@ -43,6 +45,8 @@ const (
 	minGossipBatchInterval = 50 * time.Millisecond
 )
 
+var _ p2p.Handler = (*ethTxPushGossipHandler)(nil)
```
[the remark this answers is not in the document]
joshua-kim: Removed this type

plugin/evm/gossiper.go
```diff
@@ -300,8 +308,19 @@ func (n *pushGossiper) gossipAtomicTx(tx *Tx) error {
 		"gossiping atomic tx",
 		"txID", txID,
 	)
+
+	if err := n.atomicTxClient.AppGossip(context.TODO(), msgBytes); err != nil {
+		return err
+	}
+
 	n.stats.IncAtomicGossipSent()
```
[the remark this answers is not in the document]
joshua-kim: Added client-side metrics for the sdk gossipers

peer/network.go
```diff
@@ -444,11 +444,11 @@ func (n *network) Gossip(gossip []byte) error {
 // AppGossip is called by avalanchego -> VM when there is an incoming AppGossip from a peer
 // error returned by this function is expected to be treated as fatal by the engine
 // returns error if request could not be parsed as message.Request or when the requestHandler returns an error
```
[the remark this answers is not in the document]
joshua-kim: Updated comment

plugin/evm/vm.go
joshua-kim: I think there should be less duplicated code now that there's only one gossip handler per gossip protocol... previously there was 2x2=4 which was definitely too much...

peer/network.go
```diff
@@ -444,11 +444,11 @@ func (n *network) Gossip(gossip []byte) error {
...
-func (n *network) AppGossip(_ context.Context, nodeID ids.NodeID, gossipBytes []byte) error {
+func (n *network) AppGossip(ctx context.Context, nodeID ids.NodeID, gossipBytes []byte) error {
 	var gossipMsg message.GossipMessage
 	if _, err := n.codec.Unmarshal(gossipBytes, &gossipMsg); err != nil {
-		log.Debug("could not parse app gossip", "nodeID", nodeID, "gossipLen", len(gossipBytes), "err", err)
-		return nil
+		log.Debug("forwarding AppGossip to SDK router", "nodeID", nodeID, "gossipLen", len(gossipBytes), "err", err)
```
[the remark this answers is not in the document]
joshua-kim: This is consistent w/ the current logs... going to leave this as-is for now

plugin/evm/vm.go
```diff
@@ -1084,7 +1090,13 @@ func (vm *VM) initBlockBuilding() error {
 
 	if vm.ethTxGossipHandler == nil {
 		var ethTxGossipHandler p2p.Handler
-		ethTxGossipHandler, err = gossip.NewHandler[*GossipEthTx](ethTxPool, ethTxGossipHandlerConfig, vm.sdkMetrics)
+		ethTxGossipHandler = gossip.NewHandler[GossipEthTx, *GossipEthTx](
+			vm.ctx.Log,
+			vm.ethTxPushGossiper,
```
StephenButtolph, line 1095: We need to allow inbound `AppGossip` traffic from non-validators for both eth and atomic txs. Otherwise non-validators would have no way of issuing a tx.

plugin/evm/gossiper.go
```diff
@@ -259,7 +270,24 @@ func (n *pushGossiper) awaitEthTxGossip() {
...
+					continue
+				}
+
+				gossipTxs := make([]*GossipEthTx, 0, len(txs))
+				for _, tx := range txs {
+					gossipTxs = append(gossipTxs, &GossipEthTx{Tx: tx})
+				}
+
+				n.ethTxGossiper.Add(gossipTxs...)
+				if err := n.ethTxGossiper.Gossip(context.TODO()); err != nil {
```
StephenButtolph, line 282: No sure if this is a valid concern - but it's possible that we added more transactions than we sent here. Could this ever result in a memory buildup?

plugin/evm/mempool.go
```diff
@@ -152,6 +157,10 @@ func (m *Mempool) AddTx(tx *Tx) error {
 	defer m.lock.Unlock()
```
StephenButtolph, line 157: addeed -> added? (couple lines above this)

joshua-kim: Re-spelled

plugin/evm/gossiper.go
joshua-kim: Hmm I guess this can happen if txs are coming in at a rate quicker than they're being gossiped.... but I feel like this is not too concerning since currently we're pretty much running a gossip cycle every time we see a tx I don't really expect this the txs waiting to be gossiped to build up to a noticeable size

plugin/evm/vm.go
joshua-kim: Should be fixed in the latest commit: https://github.com/ava-labs/coreth/pull/363/commits/d28cea07c33d5a299d13bf4e07649e500d767e56

plugin/evm/vm.go
joshua-kim: Added function to do this

plugin/evm/gossip.go
```diff
@@ -19,11 +22,65 @@ import (
...
+	throttlingPeriod time.Duration,
+	throttlingLimit int,
+	validators *p2p.Validators,
+) txGossipHandler {
+	// push gossip messages can be handled from any peer
+	handler := gossip.NewHandler[T, U](
+		log,
+		// Don't forward gossip to avoid double-forwarding while legacy
+		// gossip is still running.
+		gossip.NoOpAccumulator[U]{},
```
joshua-kim, line 46: So I think for now we can't actually use the push gossiper to forward transactions. The mempool gets notified of a new tx by looking at a channel which is updated on add, but since this handler adds the new txs to the mempool this will result in a double gossip if we don't skip the push gossip step after adding to the mempool.

plugin/evm/gossiper.go
```diff
@@ -259,7 +270,24 @@ func (n *pushGossiper) awaitEthTxGossip() {
 						"len(txs)", attempted,
 						"err", err,
 					)
+					continue
```
StephenButtolph, line 273: I don't think we actually need to `continue` here. But this `warn` shouldn't be firing anyways.. So it seems fine.

plugin/evm/gossiper.go
```diff
@@ -259,7 +270,24 @@ func (n *pushGossiper) awaitEthTxGossip() {
...
+				}
+
+				n.ethTxGossiper.Add(gossipTxs...)
+				if err := n.ethTxGossiper.Gossip(context.TODO()); err != nil {
+					log.Warn(
+						"failed to send eth transactions",
+						"len(txs)", len(txs),
+						"err", err,
+					)
+					continue
```
StephenButtolph, line 288: This `continue` seems useless right?
```suggestion
```

plugin/evm/gossiper.go
StephenButtolph: We could add a call to `Gossip` inside of the `gossipTicker` call if we wanted to further protect against this.

plugin/evm/mempool.go
```diff
@@ -145,13 +150,17 @@ func (m *Mempool) Add(tx *GossipAtomicTx) error {
...
+// AddTx attempts to add [tx] to the mempool and returns an error if
+// it could not be added to the mempool.
 func (m *Mempool) AddTx(tx *Tx) error {
 	m.lock.Lock()
 	defer m.lock.Unlock()
 
 	err := m.addTx(tx, false)
+	if errors.Is(err, errTxAlreadyKnown) {
+		return nil
+	}
```
StephenButtolph, line 162: We call `addTx` in `AddLocalTx` and `ForceAddTx` - don't we need to check for this there as well?

plugin/evm/vm.go
```diff
@@ -144,39 +144,22 @@ const (
 	txGossipBloomMaxItems          = 8 * 1024
 	txGossipBloomFalsePositiveRate = 0.01
 	txGossipMaxFalsePositiveRate   = 0.05
-	txGossipTargetResponseSize     = 20 * units.KiB
+	txGossipMaxSize                = 20 * units.KiB
```
StephenButtolph, line 147: I don't think `maxSize` is correct - we allow larger sizes.

plugin/evm/gossiper.go
joshua-kim: It is a no-op, it's just there because I feel like we might add code after this line in the future and unintentionally run it if this errors.

joshua-kim: It's a no-op.... I left it in because i was worried that there might be a world where we add more code and an error would result in stuff running that we didn't intend to run. I am fine removing this though

plugin/evm/mempool.go
joshua-kim: Yeah we do...

plugin/evm/vm.go
joshua-kim: renamed to txGossipTargetMessageSize

plugin/evm/gossiper.go
joshua-kim: I'll add this in.

plugin/evm/gossip.go
```diff
@@ -19,11 +22,65 @@ import (
...
+	maxMessageSize int,
+	throttlingPeriod time.Duration,
+	throttlingLimit int,
+	validators *p2p.Validators,
+) txGossipHandler {
+	// push gossip messages can be handled from any peer
+	handler := gossip.NewHandler[T, U](
+		log,
+		// Don't forward gossip to avoid double-forwarding while legacy
+		// gossip is still running.
```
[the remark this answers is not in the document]
joshua-kim: Yeah i can reword this to indicate that we're already forwarding so we don't need to double-forward here.

StephenButtolph (approved):

plugin/evm/mempool.go
```diff
@@ -21,7 +21,12 @@ const (
 	discardedTxsCacheSize = 50
 )
 
-var errNoGasUsed = errors.New("no gas used")
+var (
+	errTxAlreadyKnown = errors.New("tx already known")
```
darioush, line 25: why do we add this error then bypass it at all the callsites?

darioush (approved):

plugin/evm/mempool.go
StephenButtolph: You're totally right - this should be returned from `Add`.

StephenButtolph: 🦅 👁️

joshua-kim: Yeah this is supposed to be returned from `Add` to indicate that a duplicate existed...

plugin/evm/mempool.go
```diff
@@ -142,16 +147,20 @@ func (m *Mempool) Add(tx *GossipAtomicTx) error {
 	m.ctx.Lock.RLock()
 	defer m.ctx.Lock.RUnlock()
 
-	return m.AddTx(tx.Tx)
+	return m.addTx(tx.Tx, false)
```
StephenButtolph, line 150: This needs to hold the mempool's lock

StephenButtolph, line 150: I think we should be marking the tx as discarded if it returns an error (other than `errTxAlreadyKnown`) right?

plugin/evm/mempool.go
```diff
@@ -139,19 +144,40 @@ func (m *Mempool) atomicTxGasPrice(tx *Tx) (uint64, error) {
 }
 
 func (m *Mempool) Add(tx *GossipAtomicTx) error {
+	m.lock.Lock()
+	defer m.lock.Unlock()
```
StephenButtolph, line 148: This grabs the locks in the wrong order. The context lock must be grabbed first

joshua-kim: I'm still trying to figure out why we acquire the context lock here

plugin/evm/mempool.go
```diff
@@ -139,19 +144,40 @@ func (m *Mempool) atomicTxGasPrice(tx *Tx) (uint64, error) {
...
 
-	return m.AddTx(tx.Tx)
+	err := m.addTx(tx.Tx, false)
+	if errors.Is(err, errTxAlreadyKnown) {
+		return err
+	}
+
+	if err != nil {
+		txID := tx.Tx.ID()
+		m.discardedTxs.Put(tx.Tx.ID(), tx.Tx)
```
StephenButtolph, line 160:
```suggestion
		txID := tx.Tx.ID()
		m.discardedTxs.Put(txID, tx.Tx)
```

plugin/evm/mempool.go
joshua-kim: for future readers this is to lock state while `vm.verifyAtTip` is being run.

StephenButtolph (approved):
take 2

PR #364, opened 2023-10-25
Trim down size of secp256k1 `Factory` struct

dhrubabasu:
Mirror of https://github.com/ava-labs/avalanchego/pull/2223


aaronbuchwald:
This LGTM once the e2e tests are fixed.

To fix it, you need to update the version of AvalancheGo that is used by the e2e tests to the same version that is used in the go.mod file. ref https://github.com/ava-labs/coreth/blob/secp256k1-factory-cleanup/scripts/versions.sh#L4. The e2e tests will check out the targeted version of AvalancheGo and replace its coreth dependency with this branch to run the e2e tests here: https://github.com/ava-labs/coreth/blob/secp256k1-factory-cleanup/scripts/tests.e2e.sh#L30.

scripts/versions.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
 # Don't export them as they're used in the context of other calls
-avalanche_version=${AVALANCHE_VERSION:-'v1.10.14'}
+avalanche_version=${AVALANCHE_VERSION:-'826f9415'}
```
ceyonur, line 4: can we tag avalanchego with an rc and use it here?

ceyonur (changes requested):

aaronbuchwald (approved):

scripts/versions.sh
dhrubabasu: We didn't use an `rc` here since there were none tagged. We should be able to test against commits without rcs (@marun shipped this in #366)

ceyonur: Does this work? https://github.com/ava-labs/avalanchego/releases/tag/v1.10.15-rc.1

ceyonur: Does this work? https://github.com/ava-labs/avalanchego/releases/tag/v1.10.15-rc.1

aaronbuchwald (approved):

PR #366, opened 2023-10-30
Update e2e test script to support hash pins

marun:
I don't see another way to do this except doing a full clone instead of the previous single branch clone.


aaronbuchwald (approved):
LGTM

PR #370, opened 2023-10-31
Fix secp dependency

joshua-kim:
## Why this should be merged

https://github.com/ava-labs/avalanchego/pull/2223 Made breaking changes to `secp.Factory` so this PR updates code to match the new upstream api. This will need to be updated to support the new upstream version of avalanchego.

## How this works

See above

## How this was tested

Build passes


dhrubabasu:
This is a duplicate of #364

PR #372, opened 2023-11-01
[core/state] Prefetch Trie Nodes Concurrently

patrick-ogrady:
## Why this should be merged

Context: https://github.com/ethereum/go-ethereum/issues/28266

## How this works

This change modifies `core/state/trie_prefetcher.go` to prefetch trie nodes in parallel. In addition to this tweak, prefetching is no longer aborted (wait until all keys on a trie are prefetched) and we now prefetch tries during block building.

## How this was tested

- [x] Existing unit tests
- [x] Run on the Avalanche Network

## Open Questions

- [x] What should the default max prefetch concurrency be? ~(`32` seems low given the size of this data, see example below): https://github.com/ava-labs/coreth/pull/375~ -> start with `16`
- [x] Should we scale up copies faster? -> use the same scale up approach as bounded workers
- [x] ~Why are there any skipped prefetch tasks?~: https://github.com/ava-labs/coreth/pull/372/commits/947cad860049df5447e42beec28432625b053550
- [x] How many trie copies are created per block -> add metric
- [x] Should we reserve a trie just for copying? -> a fetch on an unpopulated path could take 7ms (7 levels), and likely one copy per 7ms
- [x] Should we reduce tasks per copy? -> could take 50ms per group (if broke into sections of 3, would only take 16-20ms), also add a metric for this
- [x] does a lock get grabbed when multiple threads fetch the same node? -> no
- [ ] ~update the base trie to one with the most populated keys from prefetching (approx): https://github.com/ava-labs/coreth/pull/396~
- [ ] ~Limit copies per `subfetcher` (could ensure no single trie is copied too many times, while still being less than disk concurrency)?~ 
- [x] Nits: https://github.com/ava-labs/coreth/pull/398

PR #376, opened 2023-11-03
core/state: add parallel trie prefetching across tr copies

aaronbuchwald:
- Adds a bounded worker implementation that will allocate at most N goroutines to execute work on demand. Execute will either execute synchronously or on one of the worker routines (must be handled correctly by the caller)
- Update trie prefetcher code to prefetch trie nodes using independent copies of the trie
- remove copy trie support from the prefetcher in favor of leaving the original trie untouched (changes from pulling trie nodes into the actual trie struct to ONLY attempting to pre-cache them)


patrick-ogrady:
Closing for #372

PR #384, opened 2023-11-08
Use `p2p.Network`

joshua-kim:
## Why this should be merged

Uses the new `p2p.Network` component

## How this works

Uses the new `p2p.Network` component which provides default values for the client. `p2p.Network` manages an internal peer set which is used as a default if no sampling strategy is provided.

## How this was tested

CI


peer/network.go
```diff
@@ -38,6 +38,7 @@ var (
 	_                     common.AppHandler    = &network{}
 )
 
+// TODO remove this interface
```
joshua-kim, line 41: If we need this interface it should be defined in the consuming package `plugin`

peer/network.go
```diff
@@ -327,7 +333,7 @@ func (n *network) AppRequest(ctx context.Context, nodeID ids.NodeID, requestID u
 	var req message.Request
 	if _, err := n.codec.Unmarshal(request, &req); err != nil {
 		log.Debug("forwarding AppRequest to SDK router", "nodeID", nodeID, "requestID", requestID, "requestLen", len(request), "err", err)
```
joshua-kim, line 335: We use the SDK's networking layer here which is just the router.... so I'm indifferent on whether we should update this log or not but will defer to others' opinions.

darioush (approved):

PR #387, opened 2023-11-13
Migrate warp backend from Subnet-EVM

aaronbuchwald:
## Why this should be merged

This PR migrates the `warp/` package and required accompanying changes from Subnet-EVM into Coreth: https://github.com/ava-labs/subnet-evm/tree/v0.5.9/warp

## How this was tested

Takes test coverage from the migrated package.


darioush (dismissed):

warp/validators/state.go
```diff
@@ -0,0 +1,51 @@
...
+) (map[ids.NodeID]*validators.GetValidatorOutput, error) {
+	// If the subnetID is anything other than the Primary Network, this is a direct
+	// passthrough
+	if subnetID != constants.PrimaryNetworkID {
+		return s.State.GetValidatorSet(ctx, height, subnetID)
+	}
+
+	// If the requested subnet is the primary network, then we return the validator
+	// set for the Subnet that is receiving the message instead.
+	return s.State.GetValidatorSet(ctx, height, s.chainContext.SubnetID)
```
ceyonur, line 50: Isn't this also `PrimaryNetworkID`? Do we need this `State` struct?

warp/handlers/signature_request_test.go
```diff
@@ -0,0 +1,202 @@
...
+			setup: func() (request message.BlockSignatureRequest, expectedResponse []byte) {
+				return message.BlockSignatureRequest{
+					BlockID: unknownMessageID,
+				}, emptySignature[:]
+			},
+			verifyStats: func(t *testing.T, stats *handlerStats) {
+				require.EqualValues(t, 0, stats.messageSignatureRequest.Count())
+				require.EqualValues(t, 0, stats.messageSignatureHit.Count())
+				require.EqualValues(t, 0, stats.messageSignatureMiss.Count())
+				require.EqualValues(t, 1, stats.blockSignatureRequest.Count())
```
ceyonur, line 172: should we add `require.EqualValues(t, stats.messageSignatureRequestDuration.Value(), int64(0))`  and `require.Greater(t, stats.blockSignatureRequestDuration.Value(), int64(0))` ? 
https://github.com/ava-labs/subnet-evm/blob/32514d6c7c6523d08a07eadda9919ef2253232de/warp/handlers/signature_request_test.go#L161

aaronbuchwald: This was explicitly removed because it fails on windows (it doesn't register the amount of elapsed time as non-zero). Made the same change on Subnet-EVM to keep them in sync.

Good 👁️

peer/client.go
```diff
@@ -23,15 +24,15 @@ type NetworkClient interface {
 	// node version greater than or equal to minVersion.
 	// Returns response bytes, the ID of the chosen peer, and ErrRequestFailed if
 	// the request should be retried.
-	SendAppRequestAny(minVersion *version.Application, request []byte) ([]byte, ids.NodeID, error)
+	SendAppRequestAny(ctx context.Context, minVersion *version.Application, request []byte) ([]byte, ids.NodeID, error)
```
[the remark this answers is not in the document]
aaronbuchwald: Merged that PR and will update this one

plugin/evm/vm.go
```diff
@@ -122,11 +122,12 @@ const (
...
-	decidedCacheSize    = 10 * units.MiB
-	missingCacheSize    = 50
-	unverifiedCacheSize = 5 * units.MiB
-	bytesToIDCacheSize  = 5 * units.MiB
+	secpCacheSize          = 1024
+	decidedCacheSize       = 10 * units.MiB
+	missingCacheSize       = 50
+	unverifiedCacheSize    = 5 * units.MiB
+	bytesToIDCacheSize     = 5 * units.MiB
+	warpSignatureCacheSize = 500
```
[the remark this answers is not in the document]
aaronbuchwald: I don't think that's necessary at this time.

plugin/evm/config.go
```diff
@@ -116,6 +117,7 @@ type Config struct {
 	AllowMissingTries               bool    `json:"allow-missing-tries"`                // If enabled, warnings preventing an incomplete trie index are suppressed
 	PopulateMissingTries            *uint64 `json:"populate-missing-tries,omitempty"`   // Sets the starting point for re-populating missing tries. Disables re-generation if nil.
 	PopulateMissingTriesParallelism int     `json:"populate-missing-tries-parallelism"` // Number of concurrent readers to use when re-populating missing tries on startup.
+	PruneWarpDB                     bool    `json:"prune-warp-db-enabled"`              // Determines if the warpDB should be cleared on startup
```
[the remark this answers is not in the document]
aaronbuchwald: Adding a commit to remove this from Subnet-EVM in https://github.com/ava-labs/subnet-evm/pull/995

warp/validators/state.go
aaronbuchwald: This will be a no-op for the C-Chain since if it is the primary network ID, then it will use its own subnetID, which is already the primary network.

I think it's easier to keep the code the same in both repos since it's a no-op for the C-Chain rather than having a discrepancy between the two.

warp/aggregator/aggregator.go
```diff
@@ -0,0 +1,173 @@
...
+// Aggregator requests signatures from validators and
+// aggregates them into a single signature.
+type Aggregator struct {
+	validators  []*avalancheWarp.Validator
+	totalWeight uint64
+	client      SignatureGetter
+}
+
+// New returns a signature aggregator for the chain with the given [state] on the
+// given [subnetID], and where [client] can be used to fetch signatures from validators.
```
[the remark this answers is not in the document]
aaronbuchwald: Updated here and in https://github.com/ava-labs/subnet-evm/pull/995

warp/aggregator/aggregator.go
```diff
@@ -0,0 +1,175 @@
...
+
```
ceyonur, line 44:
```suggestion
```

darioush (approved):

PR #396, opened 2023-11-20
[core/state] Return Best Trie Copy

patrick-ogrady:
## TODO
- [ ] fix race on updating `operations` while stored as `best`


patrick-ogrady:
In testing, found that this only leads to ~0.5 operations being saved by using the best copy (likely not worth the complexity): 
image

Issue #397, opened 2023-11-20
Support omitting writing tx indices


ceyonur:
Currently we support removing tx indices with `TxLookupLimit`, but that only removes them after indices are actually written. Providing `1` for `TxLookupLimit` would mean "index them and remove indices in the next block". We should support "not" writing those indices at all.

PR #399, opened 2023-11-20
add skip index flag

ceyonur:
This adds a new flag and skip writing tx indices along with keep running the tx undinexer

PR #400, opened 2023-11-21
Tx lookup skip

ceyonur:
It modifies type of `TxLookupLimit` flag from `uint64` type to `int64` with support for negative numbers which indicates to skip tx indexing.

Using negative number also means stop the tx unindexer from removing old indices. This is because unindexer still writes tail to DB and tries to keep up with the latest tip (even though the actual tx indexing is not performed on DB). Stopping (

Closes: #397

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -489,8 +489,8 @@ func (bc *BlockChain) dispatchTxUnindexer() {
 // - updating the acceptor tip index
 func (bc *BlockChain) writeBlockAcceptedIndices(b *types.Block) error {
 	batch := bc.db.NewBatch()
-	// negative means skip writing tx lookup entries
-	if bc.cacheConfig.TxLookupLimit >= 0 {
+	// -1 means skip writing tx lookup entries
+	if bc.cacheConfig.TxLookupLimit != -1 {
 		rawdb.WriteTxLookupEntriesByBlock(batch, b)
 	}
 	if err := rawdb.WriteAcceptorTip(batch, b.Hash()); err != nil {
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -411,11 +411,9 @@ func NewBlockChain(
 }
 
 // unindexBlocks unindexes transactions depending on user configuration
+// Invariant: This should not be called if TxLookupLimit is negative.
 func (bc *BlockChain) unindexBlocks(tail uint64, head uint64, done chan struct{}) {
 	start := time.Now()
-	if bc.cacheConfig.TxLookupLimit < 0 {
-		return
-	}
 	txLookupLimit := uint64(bc.cacheConfig.TxLookupLimit)
 	defer func() {
 		txUnindexTimer.Inc(time.Since(start).Milliseconds())
@@ -431,12 +429,10 @@ func (bc *BlockChain) unindexBlocks(tail uint64, head uint64, done chan struct{}
 // dispatchTxUnindexer is responsible for the deletion of the
 // transaction index.
 // Invariant: If TxLookupLimit is 0, it means all tx indices will be preserved.
+// Invariant: This should not be called if TxLookupLimit is negative.
 // Meaning that this function should never be called.
 func (bc *BlockChain) dispatchTxUnindexer() {
 	defer bc.wg.Done()
-	if bc.cacheConfig.TxLookupLimit < 0 {
-		return
-	}
 	txLookupLimit := uint64(bc.cacheConfig.TxLookupLimit)
 
 	// If the user just upgraded to a new version which supports transaction
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -159,7 +159,7 @@ type CacheConfig struct {
 	SnapshotVerify                  bool          // Verify generated snapshots
 	Preimages                       bool          // Whether to store preimage of trie key to the disk
 	AcceptedCacheSize               int           // Depth of accepted headers cache and accepted logs cache at the accepted tip
-	TxLookupLimit                   uint64        // Number of recent blocks for which to maintain transaction lookup indices
+	TxLookupLimit                   int64         // Number of recent blocks for which to maintain transaction lookup indices
 
 	SnapshotNoBuild bool // Whether the background generation is allowed
 	SnapshotWait    bool // Wait for snapshot construction on startup. TODO(karalabe): This is a dirty hack for testing, nuke it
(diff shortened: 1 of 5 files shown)


plugin/evm/config.go
```diff
@@ -193,9 +196,10 @@ type Config struct {
 
 	// TxLookupLimit is the maximum number of blocks from head whose tx indices
 	// are reserved:
+	// 	* -1:  means skip tx indexing
```
patrick-ogrady, line 199: Should also clarify that no txs will be cleaned up in this mode here.

eth/ethconfig/config.go
```diff
@@ -153,7 +153,8 @@ type Config struct {
 
 	// TxLookupLimit is the maximum number of blocks from head whose tx indices
 	// are reserved:
+	// 	* -1:  means skip tx indexing
```
aaronbuchwald, line 156: nit: align formatting with the values below

aaronbuchwald (approved):

core/blockchain.go
```diff
@@ -413,7 +413,10 @@ func NewBlockChain(
 // unindexBlocks unindexes transactions depending on user configuration
 func (bc *BlockChain) unindexBlocks(tail uint64, head uint64, done chan struct{}) {
 	start := time.Now()
-	txLookupLimit := bc.cacheConfig.TxLookupLimit
+	if bc.cacheConfig.TxLookupLimit < 0 {
```
patrick-ogrady, line 416: Because `unindexBlocks` is spawned by `dispatchTxUnindexer`, do we need to worry about this?

core/blockchain.go
```diff
@@ -431,7 +434,10 @@ func (bc *BlockChain) unindexBlocks(tail uint64, head uint64, done chan struct{}
 // Meaning that this function should never be called.
 func (bc *BlockChain) dispatchTxUnindexer() {
 	defer bc.wg.Done()
-	txLookupLimit := bc.cacheConfig.TxLookupLimit
+	if bc.cacheConfig.TxLookupLimit < 0 {
```
patrick-ogrady, line 437: Because we check `TxLookupLimit` before spawning `dispatchTxUnindexer`, do we need to worry about this?

patrick-ogrady:
This is very close. I think it is functionally correct right now (just has unnecessary checks).

core/blockchain.go
ceyonur: yup correct. I put it to be extra safe (in case this is directly called by tests), but I guess it's fine to just add it as a comment.

core/blockchain.go
ceyonur: responded: https://github.com/ava-labs/coreth/pull/400#discussion_r1400405969

core/blockchain_test.go
```diff
@@ -706,14 +729,16 @@ func TestTransactionIndices(t *testing.T) {
...
 	lastAcceptedHash := chain.CurrentHeader().Hash()
 
 	// Reconstruct a block chain which only reserves limited tx indices
 	// 128 blocks were previously indexed. Now we add a new block at each test step.
-	limit := []uint64{130 /* 129 + 1 reserve all */, 64 /* drop stale */, 32 /* shorten history */}
-	tails := []uint64{0 /* reserve all */, 67 /* 130 - 64 + 1 */, 100 /* 131 - 32 + 1 */}
+	limit := []int64{130 /* 129 + 1 reserve all */, 64 /* drop stale */, 32 /* shorten history */, -1 /* skip indexing */, -1 /* skip indexing */}
+	tails := []uint64{0 /* reserve all */, 67 /* 130 - 64 + 1 */, 100 /* 131 - 32 + 1 */, 100 /* skip indexing */, 100 /* skip indexing */}
+	// lastIndexed should stop with limit=-1 otherwise it should be the current block number
+	lastIndexed := []uint64{129 /* current block number */, 130 /* current block number */, 131 /* current block number */, 131 /* skip indexing */, 131 /* skip indexing */}
```
aaronbuchwald, line 741: we should convert this to a struct in the future. It was a bit hard to follow before and going from 2 to 3 values makes it a bit harder to follow.

core/blockchain.go
```diff
@@ -160,7 +160,7 @@ type CacheConfig struct {
 	SnapshotVerify                  bool          // Verify generated snapshots
 	Preimages                       bool          // Whether to store preimage of trie key to the disk
 	AcceptedCacheSize               int           // Depth of accepted headers cache and accepted logs cache at the accepted tip
-	TxLookupLimit                   uint64        // Number of recent blocks for which to maintain transaction lookup indices
+	TxLookupLimit                   int64         // Number of recent blocks for which to maintain transaction lookup indices
```
darioush, line 163: Should we add a new parameter like`tx-indexing-disabled` instead of adding -1?
Seems this would simplify validation

core/blockchain.go
```diff
@@ -405,17 +405,18 @@ func NewBlockChain(
...
-	if bc.cacheConfig.TxLookupLimit != 0 {
+	if bc.cacheConfig.TxLookupLimit > 0 {
 		bc.wg.Add(1)
 		go bc.dispatchTxUnindexer()
 	}
 	return bc, nil
 }
 
 // unindexBlocks unindexes transactions depending on user configuration
+// Invariant: This should not be called if TxLookupLimit is negative.
```
darioush, line 416: would it be clearer to take a uint param here & to dispatchTxUnindexer?

core/blockchain.go
```diff
@@ -485,7 +487,10 @@ func (bc *BlockChain) dispatchTxUnindexer() {
 // - updating the acceptor tip index
 func (bc *BlockChain) writeBlockAcceptedIndices(b *types.Block) error {
 	batch := bc.db.NewBatch()
-	rawdb.WriteTxLookupEntriesByBlock(batch, b)
+	// -1 means skip writing tx lookup entries
+	if bc.cacheConfig.TxLookupLimit != -1 {
```
darioush, line 491: Let's be consistent with using < 0 here

core/blockchain_test.go
```diff
@@ -749,7 +775,7 @@ func TestTransactionIndices(t *testing.T) {
 // TestCanonicalHashMarker tests all the canonical hash markers are updated/deleted
 // correctly in case reorg is called.
 func TestCanonicalHashMarker(t *testing.T) {
-	var cases = []struct {
+	cases := []struct {
```
darioush, line 778: could we revert the unnecessary changes?

core/blockchain_test.go
darioush: might be nice to do this as part of this PR (or immediate follow-up), doesn't seem like a big change.

core/blockchain.go
ceyonur: In that case we should validate it against TxLookupLimit and it might be confusing to use both of them at the same time.

core/blockchain.go
ceyonur: I specifically kept it -1 to make it consistent with docs/comments.

ceyonur:
closing in favor of #399

PR #406, opened 2023-11-30
Simplify atomic root acceptance

StephenButtolph:
## Why this should be merged

Improves code readability.

## How this works

1. Uses a standard packer rather than the bespoke height format that is currently being used
2. I found the usage of `nearestCommitHeight` inside of `AcceptTrie` really confusing.
  a) This re-writes the comment to explain why the logic exists rather than repeating the code.
  b) This removes the dependency on `commit` implicitly updating `lastCommittedHeight`.

## How this was tested

CI

### DIFF
--- a/plugin/evm/atomic_trie.go
+++ b/plugin/evm/atomic_trie.go
@@ -4,7 +4,6 @@
 package evm
 
 import (
-	"encoding/binary"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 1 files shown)


plugin/evm/atomic_trie.go
```diff
@@ -182,10 +181,13 @@ func lastCommittedRootIfExists(db database.Database) (common.Hash, uint64, error
...
 		return common.Hash{}, 0, err
-	case len(lastCommittedHeightBytes) != wrappers.LongLen:
-		return common.Hash{}, 0, fmt.Errorf("expected value of lastCommittedKey to be %d but was %d", wrappers.LongLen, len(lastCommittedHeightBytes))
 	}
-	height := binary.BigEndian.Uint64(lastCommittedHeightBytes)
+
+	height, err := database.ParseUInt64(lastCommittedHeightBytes)
+	if err != nil {
+		return common.Hash{}, 0, fmt.Errorf("expected value length of lastCommittedKey to be %d but was %d: %w", wrappers.LongLen, len(lastCommittedHeightBytes), err)
+	}
```
aaronbuchwald, line 189:
```suggestion
		return common.Hash{}, 0, fmt.Errorf("expected value at lastCommittedKey to be a valid uint64: %w", err)
	}
```

it seems odd to add a formatted error message that infers exactly what went wrong inside of `database.ParseUint64` instead of including that in the error message returned by `database.ParseUint64`.

darioush (approved):

PR #409, opened 2023-12-01
Skip bonus blocks applying atomic ops from  trie to shared memory

darioush:
## Why this should be merged
The bonus block atomic ops were not meant to be indexed in the trie, however due to a regression their inclusion in the trie is blocking state sync. This workaround unblocks state sync while we canonicalize the trie representation.

## How this works
Skips the bonus block heights in applying the operations from the trie to the shared memory.

## How this was tested
CI, testing with mainnet in progress

### DIFF
--- a/plugin/evm/atomic_backend.go
+++ b/plugin/evm/atomic_backend.go
@@ -270,7 +270,7 @@ func (a *atomicBackend) ApplyToSharedMemory(lastAcceptedBlock uint64) error {
 
 		// If [height] is a bonus block, do not apply the atomic operations to shared memory
 		if _, found := a.bonusBlocks[height]; found {
-			log.Info(
+			log.Debug(
 				"skipping bonus block in applying atomic ops from atomic trie to shared memory",
 				"height", height,
 			)


StephenButtolph (commented):
lgtm

StephenButtolph (dismissed):

patrick-ogrady (dismissed):

plugin/evm/atomic_backend.go
```diff
@@ -268,8 +268,16 @@ func (a *atomicBackend) ApplyToSharedMemory(lastAcceptedBlock uint64) error {
 			break
 		}
 
-		atomicOps := it.AtomicOps()
+		// If [height] is a bonus block, do not apply the atomic operations to shared memory
+		if _, found := a.bonusBlocks[height]; found {
+			log.Info(
```
darioush, line 273:
```suggestion
			log.Debug(
```

patrick-ogrady, line 273: `log.Debug`

StephenButtolph (approved):

PR #411, opened 2023-12-01
Fix trie reuse after Commit in atomic sync

darioush:
## Why this should be merged
plugin/evm code should honor the invariant specified for committing tries here:
https://github.com/ava-labs/coreth/blob/0eef20a5aca777b5177961bc349fdb0da99c2d6a/trie/trie.go#L41-L48

## How this works
Re-opens the trie from the database after commit.

## How this was tested
CI

### DIFF
--- a/plugin/evm/atomic_backend.go
+++ b/plugin/evm/atomic_backend.go
@@ -181,6 +181,11 @@ func (a *atomicBackend) initialize(lastAcceptedHeight uint64) error {
 				return err
 			}
 		}
+		// Trie should be re-opened after committing.
+		tr, err = a.atomicTrie.OpenTrie(root)
+		if err != nil {
+			return err
+		}
 
 		heightsIndexed++
 		if time.Since(lastUpdate) > progressLogFrequency {
--- a/plugin/evm/atomic_backend.go
+++ b/plugin/evm/atomic_backend.go
@@ -181,7 +181,7 @@ func (a *atomicBackend) initialize(lastAcceptedHeight uint64) error {
 				return err
 			}
 		}
-		// Trie should be re-opened after committing.
+		// Trie must be re-opened after committing (not safe for re-use after commit)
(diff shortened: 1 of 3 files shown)


plugin/evm/atomic_syncer.go
```diff
@@ -107,6 +107,12 @@ func (s *atomicSyncer) onLeafs(keys [][]byte, values [][]byte) error {
 					return err
 				}
 			}
+			// Trie should be re-opened after committing.
+			trie, err := s.atomicTrie.OpenTrie(root)
```
patrick-ogrady, line 111: Do we only need to do this if `isCommit`?

darioush: The trie is always committed, `isCommit` determines whether the `trieDB` was committed hence requiring a commit to the versiondb.

patrick-ogrady (dismissed):

plugin/evm/atomic_syncer.go
```diff
@@ -107,6 +107,12 @@ func (s *atomicSyncer) onLeafs(keys [][]byte, values [][]byte) error {
 					return err
 				}
 			}
+			// Trie should be re-opened after committing.
```
aaronbuchwald, line 110:
```suggestion
			// Trie must be re-opened after committing (not safe for re-use after commit)
```

plugin/evm/atomic_backend.go
```diff
@@ -181,6 +181,11 @@ func (a *atomicBackend) initialize(lastAcceptedHeight uint64) error {
 				return err
 			}
 		}
+		// Trie should be re-opened after committing.
```
aaronbuchwald, line 184:
```suggestion
		// Trie must be re-opened after committing (not safe for re-use after commit)
```

aaronbuchwald (dismissed):
optional suggestion to make comments more explicit - LGTM

aaronbuchwald (approved):

PR #413, opened 2023-12-01
Atomic trie height map repair

darioush:
## Why this should be merged
Some nodes may have a non-canonical mapping of block heights to corresponding atomic trie roots.
This can be caused by either:
- a bug in state sync client which incorrectly included the next heights in creating the root,
- changing the definition of the canonical mapping to include bonus blocks on nodes that did not index these.

## How this works
Rebuilds the index by iterating the trie in the background

## How this was tested
CI, pending fuji/mainnet testing.


plugin/evm/atomic_trie_height_map_repair.go
```diff
@@ -0,0 +1,113 @@
...
+		return fmt.Errorf("error iterating atomic trie: %w", err)
+	}
+	if height%a.commitInterval == 0 {
+		if err := commitRepairedHeight(height); err != nil {
+			return err
+		}
+	}
+
+	// mark height map as repaired
+	return database.PutUInt64(a.metadataDB, heightMapRepairKey, repairDone)
```
darioush, line 112: need to commit to vdb

plugin/evm/atomic_trie_height_map_repair.go
```diff
@@ -0,0 +1,117 @@
...
+			return err
+		}
+		log.Info("repaired atomic trie height map", "height", commitHeight, "root", root)
+		hasher, err = a.OpenTrie(root)
+		return err
+	}
+
+	// iterate over all leaves in the current atomic trie
+	root, _ := a.LastCommitted()
+	it, err := a.Iterator(root, database.PackUInt64(from+1))
```
patrick-ogrady, line 78: Does this iterator already start at +1?

plugin/evm/atomic_trie_height_map_repair.go
```diff
@@ -0,0 +1,117 @@
...
+	switch {
+	case errors.Is(err, database.ErrNotFound):
+		// height map not repaired yet, proceed
+	case err != nil:
+		return false, err
+	case repairFrom == repairDone:
+		// height map already repaired, nothing to do
+		return false, nil
+	}
+	return true, a.doRepairHeightMap(vdb, repairFrom, to)
```
patrick-ogrady, line 31: Do we want to start at 4096 instead of 0?

plugin/evm/atomic_trie_height_map_repair.go
```diff
@@ -0,0 +1,117 @@
...
+		for next := lastCommit + a.commitInterval; next < height; next += a.commitInterval {
+			if err := commitRepairedHeight(next); err != nil {
+				return err
+			}
+			lastCommit = next
+		}
+
+		if err := hasher.Update(it.Key(), it.Value()); err != nil {
+			return fmt.Errorf("could not update atomic trie at root %s: %w", root, err)
+		}
```
patrick-ogrady, line 100: We may want to put a "pacer" here to avoid putting a spike of load on the disk.

plugin/evm/atomic_trie_height_map_repair.go
```diff
@@ -0,0 +1,117 @@
...
+		root, nodes := hasher.Commit(false)
+		err := a.trieDB.Update(root, types.EmptyRootHash, trienode.NewWithNodeSet(nodes))
+		if err != nil {
+			return err
+		}
+		err = a.trieDB.Commit(root, false)
+		if err != nil {
+			return err
+		}
+		err = a.metadataDB.Put(database.PackUInt64(commitHeight), root[:])
```
patrick-ogrady, line 60: If we call this concurrently (during normal block processing), will the vdb we are using here get messed up?

darioush: yes this is right, perhaps we should allow the migration to commit to the versiondb along with everything else pending in the batch on the next block accept? seems janky but correct

plugin/evm/atomic_trie_height_map_repair.go
```diff
@@ -0,0 +1,121 @@
...
+	hasher, err := a.OpenTrie(fromRoot)
+	if err != nil {
+		return fmt.Errorf("could not open atomic trie at root %s: %w", fromRoot, err)
+	}
+
+	// hashes values inserted in [hasher], and stores the result in the height
+	// map at [commitHeight]. Additionally, it updates the resume marker and
+	// re-opens [hasher] to respect the trie's no use after commit invariant.
+	lastLog := from
+	logEach := uint64(100_000)
```
darioush, line 53: probably should make this quieter or time based, and add a log at the end.

plugin/evm/atomic_trie_height_map_repair_test.go
```diff
@@ -0,0 +1,119 @@
...
+			skipAtomicTxs: func(height uint64) bool { return false },
+		},
+		{
+			lastAccepted:  3 * testCommitInterval,
+			skipAtomicTxs: func(height uint64) bool { return height > testCommitInterval && height <= 2*testCommitInterval },
+		},
+		{
+			lastAccepted:  3 * testCommitInterval,
+			skipAtomicTxs: func(height uint64) bool { return height > testCommitInterval+1 },
+		},
```
aaronbuchwald, line 35: could we add names to the test cases?

aaronbuchwald (approved):
LGTM - optional nits / questions

plugin/evm/vm.go
```diff
@@ -639,6 +639,14 @@ func (vm *VM) Initialize(
 	}
 	vm.atomicTrie = vm.atomicBackend.AtomicTrie()
 
+	// Run the atomic trie height map repair in the background on mainnet/fuji
+	// TODO: remove after DUpgrade
+	if vm.chainID.Cmp(params.AvalancheMainnetChainID) == 0 ||
+		vm.chainID.Cmp(params.AvalancheFujiChainID) == 0 {
+		_, lastCommitted := vm.atomicTrie.LastCommitted()
+		go vm.atomicTrie.RepairHeightMap(lastCommitted, iterationDelay)
```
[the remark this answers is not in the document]
darioush: It's a parameter to pass 0 in tests

plugin/evm/atomic_trie_height_map_repair.go
```diff
@@ -0,0 +1,127 @@
...
+			err := a.trieDB.Update(root, types.EmptyRootHash, trienode.NewWithNodeSet(nodes))
+			if err != nil {
+				return err
+			}
+		}
+		err = a.trieDB.Commit(root, false)
+		if err != nil {
+			return err
+		}
+		err = a.metadataDB.Put(database.PackUInt64(commitHeight), root[:])
```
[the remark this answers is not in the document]
darioush: yes, this is separate to
- Avoid the logging
- Avoid updating lastCommitted in database and the in memory pointers.
- Keep the migration logic easy to follow / review

so the code duplication is not a concern for me here.

plugin/evm/atomic_trie_height_map_repair.go
```diff
@@ -0,0 +1,127 @@
...
+				return err
+			}
+			lastCommit = next
+		}
+
+		if err := hasher.Update(it.Key(), it.Value()); err != nil {
+			return fmt.Errorf("could not update atomic trie at root %s: %w", root, err)
+		}
+
+		time.Sleep(iterationDelay) // pause to avoid putting a spike of load on the disk
```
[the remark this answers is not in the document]
darioush: This takes around 10 mins on mainnet running in the background

plugin/evm/atomic_trie_iterator.go
```diff
@@ -112,3 +112,11 @@ func (a *atomicTrieIterator) AtomicOps() *atomic.Requests {
 func (a *atomicTrieIterator) Key() []byte {
 	return a.key
 }
+
+// Value returns the current database value that the iterator is iterating
+func (a *atomicTrieIterator) Value() []byte {
+	if a.trieIterator == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
darioush: removed, should not be nil ever

plugin/evm/atomic_trie_height_map_repair.go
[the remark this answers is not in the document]
darioush: Good idea to batch the sleep to avoid too much context switching

plugin/evm/atomic_trie_height_map_repair.go
```diff
@@ -0,0 +1,133 @@
...
+	)
+	commitRepairedHeight := func(commitHeight uint64) error {
+		root, nodes := hasher.Commit(false)
+		if nodes != nil {
+			err := a.trieDB.Update(root, types.EmptyRootHash, trienode.NewWithNodeSet(nodes))
+			if err != nil {
+				return err
+			}
+		}
+		err = a.trieDB.Commit(root, false)
```
ceyonur, line 66: do we need to commit here if nodes are nil?

darioush: no, updated

plugin/evm/vm.go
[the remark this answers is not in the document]
darioush: actually, I agree with you in this case, updating.

ceyonur (approved):

aaronbuchwald (approved):

plugin/evm/atomic_trie_height_map_repair.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return fmt.Errorf("could not open atomic trie at root %s: %w", fromRoot, err)
+	}
+
+	// hashes values inserted in [hasher], and stores the result in the height
+	// map at [commitHeight]. Additionally, it updates the resume marker and
+	// re-opens [hasher] to respect the trie's no use after commit invariant.
+	var (
+		lastLog = time.Now()
+		logEach = 5 * time.Minute
```
patrick-ogrady, line 56: I think we should log more frequently given what we saw in testing? maybe 90s?

plugin/evm/atomic_trie_height_map_repair.go
patrick-ogrady: Seems like vdb was removed here, so resolving convo

patrick-ogrady (approved):

PR #416, opened 2023-12-05
Atomic trie repair script

darioush:


plugin/evm/bonus_blocks.go
```diff
@@ -0,0 +1,70 @@
...
+
+package evm
+
+// mainnetBonusBlocksRlp is a map of bonus block numbers to their RLP-encoded
+// block data. These blocks are hardcoded so nodes that do not have these blocks
+// can add their atomic operations to the atomic trie so all nodes on have a
+// canonical atomic trie.
+// Initially, bonus blocks were not indexed into the atomic trie. However, a
+// regression caused some nodes to index these blocks.
+var mainnetBonusBlocksRlp = map[uint64]string{
```
patrick-ogrady, line 12: Does it make sense to import these via go:embed instead?

plugin/evm/atomic_backend.go
```diff
@@ -96,6 +108,14 @@ func NewAtomicBackend(
...
 		return nil, err
 	}
+	if len(bonusBlocksRlp) > 0 {
+		if err := atomicTrie.repairAtomicTrie(bonusBlocks, bonusBlocksRlp); err != nil {
+			return nil, err
+		}
+		if err := db.Commit(); err != nil {
+			return nil, err
+		}
+	}
```
patrick-ogrady, line 118: After repairing the atomic trie, should we free the memory associated with `bonusBlocks`? May be overkill for its size.

plugin/evm/bonus_blocks.go
patrick-ogrady: May be overkill for its size.

plugin/evm/atomic_trie_repair.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	puts, removes := 0, 0
+	for height, block := range bonusBlocks {
+		if height > lastCommitted {
+			// Avoid applying the repair to heights not yet committed
+			continue
+		}
+
+		blockID, ok := bonusBlockIDs[height]
```
patrick-ogrady, line 49: Should we check that this will never happen via some init method instead?

plugin/evm/atomic_trie_repair.go
```diff
@@ -0,0 +1,101 @@
...
+		if !ok {
+			return fmt.Errorf("missing block ID for height %d", height)
+		}
+		txs, err := extractAtomicTxsFromRlp(block, a.codec, blockID)
+		if err != nil {
+			return fmt.Errorf("failed to extract atomic txs from bonus block at height %d: %w", height, err)
+		}
+		log.Info("repairing atomic trie", "height", height, "block", blockID, "txs", len(txs))
+		if len(txs) == 0 {
```
patrick-ogrady, line 58: Should we comment that this should not happen?

plugin/evm/atomic_trie_repair.go
```diff
@@ -0,0 +1,101 @@
...
+		if err := a.UpdateTrie(tr, height, combinedOps); err != nil {
+			return err
+		}
+		for _, op := range combinedOps {
+			puts += len(op.PutRequests)
+			removes += len(op.RemoveRequests)
+		}
+		a.heightsRepaired++
+	}
+	newRoot, nodes := tr.Commit(false)
```
patrick-ogrady, line 74: Should we commit after each update?

patrick-ogrady: seems wasteful.

plugin/evm/atomic_trie_repair.go
```diff
@@ -0,0 +1,101 @@
...
+			return err
+		}
+		if err := a.UpdateTrie(tr, height, combinedOps); err != nil {
+			return err
+		}
+		for _, op := range combinedOps {
+			puts += len(op.PutRequests)
+			removes += len(op.RemoveRequests)
+		}
+		a.heightsRepaired++
```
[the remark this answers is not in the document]
darioush: Mainly to check whether the migration runs in testing. Could modify the function to return bool if you prefer.

plugin/evm/atomic_backend.go
darioush: added code to free it in vm.Initialize()

plugin/evm/block.go
```diff
@@ -109,6 +122,19 @@ func init() {
 		}
 		bonusBlockMainnetHeights[height] = blkID
 	}
+
+	err := json.Unmarshal(mainnetBonusBlocksJson, &mainnetBonusBlocksRlp)
+	if err != nil {
+		panic(err)
+	}
+	for height := range mainnetBonusBlocksRlp {
```
patrick-ogrady, line 130: 👍

plugin/evm/vm.go
```diff
@@ -606,12 +606,19 @@ func (vm *VM) Initialize(
...
+		bonusBlockRepair      map[uint64]string
 		canonicalBlockHeights []uint64
 	)
 	if vm.chainID.Cmp(params.AvalancheMainnetChainID) == 0 {
 		bonusBlockHeights = bonusBlockMainnetHeights
+		bonusBlockRepair = mainnetBonusBlocksRlp
 		canonicalBlockHeights = canonicalBlockMainnetHeights
 	}
+	defer func() {
+		// Free memory after VM is initialized
```
patrick-ogrady, line 618: If this runs async, we'll need to avoid doing this?

plugin/evm/atomic_trie_repair.go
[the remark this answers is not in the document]
patrick-ogrady: I think you could just return an `int`? Seems cleanest and best for testing.

plugin/evm/atomic_trie_repair.go
```diff
@@ -0,0 +1,103 @@
...
+	puts, removes := 0, 0
+	for height, block := range bonusBlocks {
+		if height > lastCommitted {
+			// Avoid applying the repair to heights not yet committed
+			continue
+		}
+
+		blockID, ok := bonusBlockIDs[height]
+		if !ok {
+			return 0, fmt.Errorf("missing block ID for height %d", height)
```
patrick-ogrady, line 53: Probably should also comment that this shouldn't/can't happen?

plugin/evm/atomic_trie_repair.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	newRoot, nodes := tr.Commit(false)
+	if nodes != nil {
+		if err := a.trieDB.Update(newRoot, types.EmptyRootHash, trienode.NewWithNodeSet(nodes)); err != nil {
+			return 0, err
+		}
+	}
+	if err := a.commit(lastCommitted, newRoot); err != nil {
+		return 0, err
+	}
```
ceyonur, line 86: do we still need to commit here if `nodes == nil`?

aaronbuchwald (approved):
Small suggestions / questions only

plugin/evm/atomic_trie_repair.go
darioush: no it is not needed

plugin/evm/atomic_trie_repair_test.go
```diff
@@ -0,0 +1,101 @@
...
+	db.Abort()
+	atomicBackend, err = NewAtomicBackendWithBonusBlockRepair(
+		db, testSharedMemory(), bonusBlockMainnetHeights, mainnetBonusBlocksRlp,
+		repo, commitHeight, common.Hash{}, commitInterval,
+	)
+	require.NoError(err)
+	a = atomicBackend.AtomicTrie().(*atomicTrie)
+	heightsRepaired, err := a.repairAtomicTrie(bonusBlockMainnetHeights, mainnetBonusBlocksRlp)
+	require.NoError(err)
+	require.Zero(heightsRepaired) // migration should not run a second time
```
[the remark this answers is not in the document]
darioush: clarified the test by restructuring it a bit

ceyonur (approved):

plugin/evm/atomic_trie_repair.go
```diff
@@ -0,0 +1,90 @@
...
+		txs, err := ExtractAtomicTxs(block.ExtData(), false, a.codec)
+		if err != nil {
+			return 0, fmt.Errorf("failed to extract atomic txs from bonus block at height %d: %w", height, err)
+		}
+		log.Info("repairing atomic trie", "height", height, "block", blockID, "txs", len(txs))
+		combinedOps, err := mergeAtomicOps(txs)
+		if err != nil {
+			return 0, err
+		}
+		if err := a.UpdateTrie(tr, height, combinedOps); err != nil {
```
patrick-ogrady, line 63: to ask the dumb question, this is a no-op if we already have the atomic ops indexed right?

plugin/evm/atomic_trie_repair.go
```diff
@@ -0,0 +1,90 @@
...
+	if nodes != nil {
+		if err := a.trieDB.Update(newRoot, types.EmptyRootHash, trienode.NewWithNodeSet(nodes)); err != nil {
+			return 0, err
+		}
+		if err := a.commit(lastCommitted, newRoot); err != nil {
+			return 0, err
+		}
+	}
+
+	if err := a.metadataDB.Put(repairedKey, []byte{1}); err != nil {
```
patrick-ogrady, line 82: IIRC there may be a sentinel helper for storing this in `avalanchego`?

plugin/evm/atomic_backend.go
```diff
@@ -88,13 +89,35 @@ func NewAtomicBackend(
...
 	if err != nil {
-		return nil, err
+		return nil, 0, err
+	}
+	var heightsRepaired int
+	if len(bonusBlocksParsed) > 0 {
+		if heightsRepaired, err = atomicTrie.repairAtomicTrie(bonusBlocks, bonusBlocksParsed); err != nil {
+			return nil, 0, err
+		}
+		if err := db.Commit(); err != nil {
```
patrick-ogrady, line 118: if heightsRepaired == 0, should we not commit?

plugin/evm/atomic_trie_repair.go
darioush: yes

plugin/evm/atomic_trie_repair.go
```diff
@@ -79,7 +80,7 @@ func (a *atomicTrie) repairAtomicTrie(bonusBlockIDs map[uint64]ids.ID, bonusBloc
 		}
 	}
 
-	if err := a.metadataDB.Put(repairedKey, []byte{1}); err != nil {
+	if database.PutBool(a.metadataDB, repairedKey, true); err != nil {
```
patrick-ogrady, line 83: Is there a helper that could be used for `done, err := a.metadataDB.Has(repairedKey)` too?

darioush: we could use GetBool but then special case database.ErrNotFound , so using Has seemed cleaner to me, but I can change if you prefer

patrick-ogrady: I think `Has` is fine, you could comment above this method that putting `true/false` here both are considered repaired.

patrick-ogrady (approved):

PR #419, opened 2023-12-06
Vm app error

joshua-kim:
## Why this should be merged

Supports the new `p2p.AppError` type.

## How this works

Passes the new app error through into the SDK's networking layer.

## How this was tested

CI

### DIFF
--- a/scripts/versions.sh
+++ b/scripts/versions.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
 # Don't export them as they're used in the context of other calls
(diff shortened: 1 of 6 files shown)


scripts/versions.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
 # Don't export them as they're used in the context of other calls
-avalanche_version=${AVALANCHE_VERSION:-'v1.10.18-rc.0'}
+avalanche_version=${AVALANCHE_VERSION:-'app-error-0.12.9-rc.1'}
```
joshua-kim, line 4: Need to update with a real RC before merge

PR #421, opened 2023-12-09
Update to new SDK API

joshua-kim:
## Why this should be merged

See https://github.com/ava-labs/avalanchego/pull/2452

## How this works

Updates to new SDK API

## How this was tested

UTs

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.20
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.10.0
-	github.com/ava-labs/avalanchego v1.10.18-rc.0
+	github.com/ava-labs/avalanchego v1.10.18-0.20231211033727-3c26d85f63c9
 	github.com/cespare/cp v0.1.0
 	github.com/cockroachdb/pebble v0.0.0-20230209160836-829675f94811
 	github.com/davecgh/go-spew v1.1.1
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.20
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.10.0
-	github.com/ava-labs/avalanchego v1.10.18-0.20231212202631-654fab923075
+	github.com/ava-labs/avalanchego v1.10.18-0.20231213001625-4a76d7f73fa6
 	github.com/cespare/cp v0.1.0
 	github.com/cockroachdb/pebble v0.0.0-20230209160836-829675f94811
 	github.com/davecgh/go-spew v1.1.1
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.20
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.10.0
-	github.com/ava-labs/avalanchego v1.10.18-0.20231212202014-f5d1250b5745
+	github.com/ava-labs/avalanchego v1.10.18-0.20231212202631-654fab923075
 	github.com/cespare/cp v0.1.0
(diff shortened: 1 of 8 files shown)


peer/network.go
```diff
@@ -78,9 +78,10 @@ type Network interface {
 	// (length of response divided by request time), and with 0 if the response is invalid.
 	TrackBandwidth(nodeID ids.NodeID, bandwidth float64)
 
-	// NewAppProtocol reserves a protocol identifier and returns a corresponding
-	// client to send messages with
-	NewAppProtocol(protocol uint64, handler p2p.Handler, options ...p2p.ClientOption) (*p2p.Client, error)
+	// NewClient returns a client to send messages with for the given protocol
+	NewClient(protocol uint64, options ...p2p.ClientOption) (*p2p.Client, error)
+	// NewAppProtocol registers a server handler for an application protocol
```
abi87, line 83: nit: seems outdate comment, drop it

peer/network.go
```diff
@@ -78,9 +78,10 @@ type Network interface {
...
+	AddHandler(protocol uint64, handler p2p.Handler) error
```
abi87, line 84: q: could you comment on the difference between `AddHandler` and the various `SetGossip/Request/CrossChainRequest/Handler`?

peer/network.go
joshua-kim: Whoops... removed

peer/network.go
joshua-kim: `AddHandler` is for the p2p sdk handlers. The existing signatures are for the legacy handlers. Eventually we will delete all the old stuff from this once migration is complete

peer/network.go
```diff
@@ -543,8 +544,12 @@ func (n *network) TrackBandwidth(nodeID ids.NodeID, bandwidth float64) {
...
 }
 
-func (n *network) NewAppProtocol(protocol uint64, handler p2p.Handler, options ...p2p.ClientOption) (*p2p.Client, error) {
-	return n.network.NewAppProtocol(protocol, handler, options...)
+func (n *network) NewClient(protocol uint64, options ...p2p.ClientOption) (*p2p.Client, error) {
+	return n.network.NewClient(protocol, options...)
+}
+
+func (n *network) AddHandler(protocol uint64, handler p2p.Handler) error {
+	return n.network.AddHandler(protocol, handler)
```
ceyonur, line 552: can we rename the inner `network `struct (p2p)?

ceyonur: Or alternatively maybe we can directly embed `*p2p.Network` in to the struct.

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.20
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.10.0
-	github.com/ava-labs/avalanchego v1.10.18-rc.0
+	github.com/ava-labs/avalanchego v1.10.18-0.20231211033727-3c26d85f63c9
```
ceyonur, line 7: can we use the same tag for `1.10.123-refactor-client-1`?

peer/network.go
joshua-kim: Embedding makes this look weird still sine the name is then `n.Network`... i'll just rename this to something else

abi87 (approved):
lgtm

darioush (approved):

go.mod
joshua-kim: I'm using a commit hash now since I think the tags don't work with e2e.

StephenButtolph (approved):

PR #422, opened 2023-12-11
Implement `gossip.Set#Get`

joshua-kim:
## Why this should be merged

Required as a pre-req to migrate our push-based gossip

## How this works

Implements the `Get` interface. This is used so that we can see if a tx is new or not, to decide on whether we want to forward it to peers.

## How this was tested

Added UTs

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -344,6 +344,7 @@ type VM struct {
 	// Used to serve BLS signatures of warp messages over RPC
 	warpBackend warp.Backend
 
+	// Can be overridden in unit tests
 	p2pSender             commonEng.AppSender
 	ethTxGossipHandler    p2p.Handler
 	atomicTxGossipHandler p2p.Handler
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -1065,6 +1065,9 @@ func (vm *VM) initBlockBuilding() error {
 	ctx, cancel := context.WithCancel(context.TODO())
 	vm.cancel = cancel
 
+	ethTxGossipClient := vm.Network.NewClient(ethTxGossipProtocol, p2p.WithValidatorSampling(vm.validators))
+	atomicTxGossipClient := vm.Network.NewClient(atomicTxGossipProtocol, p2p.WithValidatorSampling(vm.validators))
+
 	// NOTE: gossip network must be initialized first otherwise ETH tx gossip will not work.
 	gossipStats := NewGossipStats()
 	vm.gossiper = vm.createGossiper(gossipStats)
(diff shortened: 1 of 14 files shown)


plugin/evm/vm.go
```diff
@@ -1147,6 +1131,31 @@ func (vm *VM) initBlockBuilding() error {
...
+
+	atomicTxGossipHandler = &p2p.ValidatorHandler{
+		ValidatorSet: vm.validators,
+		Handler: &p2p.ThrottlerHandler{
+			Throttler: p2p.NewSlidingWindowThrottler(throttlingPeriod, throttlingLimit),
+			Handler:   atomicTxGossipHandler,
+			Log:       vm.ctx.Log,
+		},
+		Log: vm.ctx.Log,
+	}
```
ceyonur, line 1153: is there an equivalent `New` function for this? If so can we use it (or create if possible)?

plugin/evm/vm.go
```diff
@@ -1071,55 +1071,39 @@ func (vm *VM) initBlockBuilding() error {
...
+		ethTxPool,
+		ethTxGossipHandlerConfig,
+		vm.sdkMetrics,
+	)
 	if err != nil {
 		return err
 	}
-
-	atomicTxGossipHandler = &p2p.ValidatorHandler{
+	ethTxGossipHandler = &p2p.ValidatorHandler{
```
ceyonur, line 1093: these reassignments seems a bit complicated. can we use a different naming for these or try to create a separate `New...` function?

plugin/evm/vm.go
```diff
@@ -1071,55 +1071,39 @@ func (vm *VM) initBlockBuilding() error {
...
 		ValidatorSet: vm.validators,
 		Handler: &p2p.ThrottlerHandler{
+			Handler:   ethTxGossipHandler,
```
ceyonur, line 1096: (from comment above) this looks a bit weird and seems like there is a circular dependency.

plugin/evm/vm.go
```diff
@@ -1071,55 +1071,39 @@ func (vm *VM) initBlockBuilding() error {
...
-		Log: vm.ctx.Log,
-	}
-
-	if err := vm.Network.AddHandler(ethTxGossipProtocol, ethTxGossipHandler); err != nil {
-		return err
+		return fmt.Errorf("failed to create eth tx gossip client: %w", err)
 	}
 
-	atomicTxGossipHandler, err = gossip.NewHandler[*GossipAtomicTx](vm.mempool, atomicTxGossipHandlerConfig, vm.sdkMetrics)
+	ethTxGossipHandler, err = gossip.NewHandler[GossipEthTx, *GossipEthTx](
```
ceyonur, line 1083: we definitely need more comments here as there are currently 4 handler references for eth txs (`NewGossipHandler`, `gossip.NewHandler`, ` &p2p.ValidatorHandler`, `gossip.NewPullGossiper`)

ceyonur (commented):
Left few comments. I think it would be much cleaner if we merge this into the `refactor-client` PR as they both change same files. Or we can just keep mempool changes here and move other stuff to `refactor-client` if possible.

joshua-kim:
> Left few comments. I think it would be much cleaner if we merge this into the refactor-client PR as they both change same files. Or we can just keep mempool changes here and move other stuff to refactor-client if possible

I think i was low on sleep when I pushed this out... yeah we should definitely be putting the non-mempool stuff into the other PR.

plugin/evm/vm.go
joshua-kim: Argh I missed this one. I'll add this as a follow-up... https://github.com/ava-labs/avalanchego/issues/2482

plugin/evm/vm.go
joshua-kim: Stale diff, I think this is fixed now in the refactor PR.

plugin/evm/vm.go
joshua-kim: Stale diff, I think this is more verbose in the refactor PR

joshua-kim:
Closing because we are removing `Get` in favor of just using an `error` on `Add` to indicate the tx already exists.

PR #423, opened 2023-12-11
Fix codec size for large blocks

darioush:
## Why this should be merged
Initializing the codec as such:

https://github.com/ava-labs/coreth/blame/abde8d63c32d75b8e966e41eabea0051a2ffa50d/plugin/evm/message/codec.go#L24-L25

does not result in the codec's max slice marshal capacity increasing (see https://github.com/ava-labs/avalanchego/blob/master/codec/linearcodec/codec.go#L55-L65)

Fixes https://github.com/ava-labs/avalanchego/issues/2454

Also this reduces the log level for retrying failed messages from peers to Debug so invalid responses cannot cause syncing clients to log @ info.

## How this works
Calls `NewCustomMaxLength`.

## How this was tested
Fixes the UT to encode a large block

### DIFF
--- a/plugin/evm/message/codec.go
+++ b/plugin/evm/message/codec.go
@@ -22,7 +22,7 @@ var (
 
 func init() {
 	Codec = codec.NewManager(maxMessageSize)
-	c := linearcodec.NewDefault()
+	c := linearcodec.NewCustomMaxLength(maxMessageSize)
 
 	errs := wrappers.Errs{}
 	errs.Add(
@@ -54,7 +54,7 @@ func init() {
 	}
 
 	CrossChainCodec = codec.NewManager(maxMessageSize)
-	ccc := linearcodec.NewDefault()
+	ccc := linearcodec.NewCustomMaxLength(maxMessageSize)
 
 	errs = wrappers.Errs{}
 	errs.Add(
--- a/plugin/evm/message/codec.go
+++ b/plugin/evm/message/codec.go
@@ -21,7 +21,7 @@ var (
 )
 
 func init() {
-	Codec = codec.NewManager(maxMessageSize)
+	Codec = codec.NewManager(maxMessageSize + wrappers.ShortLen) // Add room for the codec version
 	c := linearcodec.NewCustomMaxLength(maxMessageSize)
 
 	errs := wrappers.Errs{}
--- a/plugin/evm/message/codec.go
+++ b/plugin/evm/message/codec.go
@@ -53,7 +53,7 @@ func init() {
 		panic(errs.Err)
 	}
 
-	CrossChainCodec = codec.NewManager(maxMessageSize)
+	CrossChainCodec = codec.NewManager(maxMessageSize + wrappers.ShortLen) // Add room for the codec version
 	ccc := linearcodec.NewCustomMaxLength(maxMessageSize)
 
 	errs = wrappers.Errs{}
--- a/plugin/evm/message/codec.go
+++ b/plugin/evm/message/codec.go
@@ -21,7 +21,7 @@ var (
 )
 
 func init() {
-	Codec = codec.NewManager(maxMessageSize + wrappers.ShortLen) // Add room for the codec version
+	Codec = codec.NewManager(maxMessageSize)
 	c := linearcodec.NewCustomMaxLength(maxMessageSize)
 
 	errs := wrappers.Errs{}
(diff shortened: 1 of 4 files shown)


sync/client/client.go
```diff
@@ -351,7 +351,7 @@ func (c *client) get(ctx context.Context, request message.Request, parseFn parse
 			responseIntf, numElements, err = parseFn(c.codec, request, response)
 			if err != nil {
 				lastErr = err
-				log.Info("could not validate response, retrying", "nodeID", nodeID, "attempt", attempt, "request", request, "err", err)
+				log.Debug("could not validate response, retrying", "nodeID", nodeID, "attempt", attempt, "request", request, "err", err)
```
patrick-ogrady, line 354: This would be a great metric to track if we can?

plugin/evm/message/codec.go
```diff
@@ -22,7 +22,7 @@ var (
 
 func init() {
 	Codec = codec.NewManager(maxMessageSize)
-	c := linearcodec.NewDefault()
```
patrick-ogrady, line 25: May be worth comparing to how we initialize this in AvalancheGo: https://github.com/ava-labs/avalanchego/blob/7623ffd4be915a5185c9ed5e11fa9be15a6e1f00/vms/proposervm/block/codec.go#L24-L25

darioush: I think it would be safe to use large maximums here, though I would prefer to make a "smaller" change

patrick-ogrady: I think `Codec.Manager` adds `ShortLen` for the codec version?

So, setting them equal may not work: https://github.com/ava-labs/avalanchego/blob/7623ffd4be915a5185c9ed5e11fa9be15a6e1f00/codec/manager.go#L126

sync/handlers/block_request_test.go
```diff
@@ -167,7 +167,7 @@ func TestBlockRequestHandlerLargeBlocks(t *testing.T) {
 	blocks, _, err := core.GenerateChain(gspec.Config, genesis, engine, memdb, 96, 0, func(i int, b *core.BlockGen) {
 		var data []byte
 		switch {
-		case i < 32:
+		case i <= 32:
```
StephenButtolph, line 170: What does this do?

plugin/evm/message/codec.go
StephenButtolph: Other than compatibility reasons, I think we're safe to pass `MaxInt` for all the maximum slice length logic... So I think this is fine. Previously this could have caused 64 MiBs to be allocated when parsing `CodeRequest` with just a few bytes (because the slice would be pre-allocated). However, the codec doesn't do that anymore.

plugin/evm/message/codec.go
```diff
@@ -21,8 +21,8 @@ var (
 )
 
 func init() {
-	Codec = codec.NewManager(maxMessageSize)
-	c := linearcodec.NewDefault()
+	Codec = codec.NewManager(maxMessageSize + wrappers.ShortLen) // Add room for the codec version
```
StephenButtolph, line 24: I'd prefer to revert this `+ wrappers.ShortLen`. It isn't needed and including it makes it seem like it is.

sync/handlers/block_request_test.go
darioush: Causes [this test](https://github.com/ava-labs/coreth/pull/423/files#diff-ff24a27eb19d31cab32d1ffea0260444286838e6a3b56c279594b72b60f2e9baL200-L204) to serialize a large block (1MB).
Without this change, the test would serialize a small block, then skip adding the large block to the output, which is why we didn't detect this issue before.

plugin/evm/message/codec.go
darioush: @StephenButtolph  This has been limiting the size of the message that gets sent through SendAppGossip as well, is that safe too?

sync/client/client.go
ceyonur: you mean as a separate metric just for unparsable responses?

plugin/evm/message/codec.go
ceyonur: I think it will still be limited by the `maxMessageSize`, right?

ceyonur (approved):
changes LGTM just left few questions

aaronbuchwald (approved):

plugin/evm/message/codec.go
StephenButtolph: Yes this is safe.

StephenButtolph (approved):

PR #425, opened 2023-12-12
Refactor p2p unit tests

joshua-kim:
## Why this should be merged

the P2P unit tests are confusing + spin up goroutines, this PR refactors all of them to be simpler

## How this works

See above

## How this was tested

UTs

### DIFF
--- a/peer/network_test.go
+++ b/peer/network_test.go
@@ -846,9 +846,17 @@ func TestNetworkCrossChainAppRequestAfterShutdown(t *testing.T) {
 
 func TestNetworkRouting(t *testing.T) {
 	require := require.New(t)
+	sender := &testAppSender{
+		sendAppRequestFn: func(_ context.Context, s set.Set[ids.NodeID], u uint32, bytes []byte) error {
+			return nil
+		},
+		sendAppResponseFn: func(id ids.NodeID, u uint32, bytes []byte) error {
+			return nil
+		},
+	}
 	protocol := 0
 	handler := &testSDKHandler{}
-  p2pNetwork, err := p2p.NewNetwork(logging.NoLog{}, &common.SenderTest{}, prometheus.NewRegistry(), "")
+	p2pNetwork, err := p2p.NewNetwork(logging.NoLog{}, sender, prometheus.NewRegistry(), "")
 	require.NoError(err)
 	require.NoError(p2pNetwork.AddHandler(uint64(protocol), handler))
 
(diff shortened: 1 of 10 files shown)


plugin/evm/vm.go
```diff
@@ -343,6 +343,12 @@ type VM struct {
 	// Avalanche Warp Messaging backend
 	// Used to serve BLS signatures of warp messages over RPC
 	warpBackend warp.Backend
+
+	p2pSender             p2p.AppSender
+	ethTxGossipHandler    p2p.Handler
+	atomicTxGossipHandler p2p.Handler
+	ethTxGossiper         gossip.Gossiper
+	atomicTxGossiper      gossip.Gossiper
```
joshua-kim, line 351: Keeping these interfaces here lets us override these with noop implementations during testing.

scripts/versions.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
 # Don't export them as they're used in the context of other calls
-avalanche_version=${AVALANCHE_VERSION:-'v1.10.18-rc.0'}
+avalanche_version=${AVALANCHE_VERSION:-'sender-refactor'}
```
darioush, line 4: is this necessary? (only impacts e2e tests)

plugin/evm/vm.go
darioush: could we add this as a comment?

darioush (dismissed):
I think this is a good improvement so I am approving this PR.

I think we should find a coherent strategy for initializing & wiring objects during testing. May be worthwhile to look into golang best practices or agree on some standards for further improvement.

scripts/versions.sh
joshua-kim: I'm not sure how to get e2e working with a custom tag that isn't an official version tag... haven't dove to see why it doesn't work

peer/network_test.go
```diff
@@ -425,6 +425,7 @@ func TestRequestMinVersion(t *testing.T) {
 	responseBytes, _, err := client.SendAppRequestAny(
 		context.Background(),
 		&version.Application{
+			Name:  "avalanche",
```
joshua-kim, line 428: Diff included because of the upgrade to latest avalanchego

plugin/evm/vm.go
```diff
@@ -344,7 +344,8 @@ type VM struct {
 	// Used to serve BLS signatures of warp messages over RPC
 	warpBackend warp.Backend
 
-	p2pSender             p2p.AppSender
+	// Can be overridden in unit tests
```
darioush, line 347:
```suggestion
	// Initialize only sets these if nil so they can be overridden in tests
```

darioush (approved):

StephenButtolph (approved):

PR #428, opened 2023-12-18
Off chain warp

aaronbuchwald:
This PR addresses https://github.com/ava-labs/subnet-evm/issues/729

This PR explicitly only supports the payload type `AddressedCall` and only serves messages from the message handler (not block signature handler) since the planned use case for offchain warp messages is to used the `AddressedCall` type.


darioush (approved):

ceyonur (approved):

plugin/evm/config.go
```diff
@@ -205,6 +206,10 @@ type Config struct {
 	// This is useful for validators that don't need to index transactions.
 	// TxLookupLimit can be still used to control unindexing old transactions.
 	SkipTxIndexing bool `json:"skip-tx-indexing"`
+
+	// WarpOffChainMessages encodes off-chain messages (unrelated to on-chain event ie. block or AddressedCall)
```
minghinmatthewlam, line 210: will off-chain block hashes be supported later on? Below it looks like only addressed calls are supported

minghinmatthewlam: nvm PR descrption says AddressedCall only, but I think worth to add a comment here pointing that out

warp/backend.go
```diff
@@ -157,6 +189,9 @@ func (b *backend) GetBlockSignature(blockID ids.ID) ([bls.SignatureLen]byte, err
 }
 
 func (b *backend) GetMessage(messageID ids.ID) (*avalancheWarp.UnsignedMessage, error) {
+	if message, ok := b.offchainAddressedCallMsgs[messageID]; ok {
+		return message, nil
+	}
 	if message, ok := b.messageCache.Get(messageID); ok {
 		return message, nil
 	}
```
martineckardt, line 197: I assume the vast majority of warp messages will not be offchain messages, but regular messages. Could we do a small performance improvement here to check the messageCache before the offchainAddressedCallMsgs, since there is a much higher chance for the message to be in there?

plugin/evm/config.go
aaronbuchwald: Good call

warp/backend.go
aaronbuchwald: Sure

PR #430, opened 2023-12-19
Add `TestSnowContext` util

dhrubabasu:
## Why this should be merged

We're removing `snow.DefaultContextTest` in AvalancheGo (https://github.com/ava-labs/avalanchego/pull/2518). In-lining this function in coreth to unblock that PR

Don't have any particular attachment to putting it in `utils`, feel free to move wherever you think makes sense!

## How this works

add func and use it

## How this was tested

CI


plugin/evm/tx_gossip_test.go
```diff
@@ -34,12 +34,13 @@ import (
 
 	"github.com/ava-labs/coreth/core/types"
 	"github.com/ava-labs/coreth/params"
+	corethUtils "github.com/ava-labs/coreth/utils"
```
ceyonur, line 37: IMO it's better to special name the avalanchego/utils rather than coreth/utils

precompile/contracts/warp/contract_test.go
```diff
@@ -19,6 +18,7 @@ import (
 	"github.com/ava-labs/coreth/precompile/contract"
 	"github.com/ava-labs/coreth/precompile/testutils"
 	"github.com/ava-labs/coreth/predicate"
+	corethUtils "github.com/ava-labs/coreth/utils"
```
ceyonur, line 21: same comment about coreth vs avago import naming

ceyonur (commented):
we probably need a similar PR to subnet-evm as well.

ceyonur (approved):

JacobEverly (approved):

PR #443, opened 2023-12-28
refactor: move AllowUnfinalizedQueries out of vm.Config

darioush:
## Why this should be merged
I don't think `AllowUnfinalizedQueries` belongs in `core/vm.Config`, since that is the interpreter configuration.

## How this works
Moves `AllowUnfinalizedQueries` to the backend

## How this was tested
CI


ceyonur (commented):
just a nit

ceyonur (approved):

PR #451, opened 2024-01-04
Update avalanchego to remove slice length parsing

StephenButtolph:
## Why this should be merged

As part of durango, we are removing enforcement of the maximum slice length during codec management.

## How this works

- Updates the avalanchego dependency.
- Enables specifying the durango time to the global codec.

## How this was tested

- [ ] CI


plugin/evm/message/codec.go
```diff
@@ -22,7 +24,7 @@ var (
 
 func init() {
 	Codec = codec.NewManager(maxMessageSize)
-	c := linearcodec.NewCustomMaxLength(maxMessageSize)
+	c := linearcodec.NewDefault(time.Time{})
```
ceyonur, line 27: I guess these should be created with actual durango times? (since they were using custom max lenght before?)

StephenButtolph: `maxMessageSize` is still enforced by the codec manager. So we would need to be serializing 0-length types for this to change the validity rules

darioush (approved):

patrick-ogrady (approved):

PR #452, opened 2024-01-05
[utils/bloom] Dynamic Bloom Filter

patrick-ogrady:
## Why this should be merged

Rationale: https://github.com/ava-labs/avalanchego/pull/2591

## How this works

Modifies reset logic to grow the bloom filter if the number of target items grows.

## How this was tested

CI + Load Test

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.20
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.10.0
-	github.com/ava-labs/avalanchego v1.10.18-rc.15
+	github.com/ava-labs/avalanchego v1.10.18-rc.5.0.20240105233604-41587de9c530
 	github.com/cespare/cp v0.1.0
 	github.com/cockroachdb/pebble v0.0.0-20230209160836-829675f94811
 	github.com/davecgh/go-spew v1.1.1
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.20
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.10.0
-	github.com/ava-labs/avalanchego v1.10.18-rc.5.0.20240110070430-a905aaea3b06
+	github.com/ava-labs/avalanchego v1.10.18-rc.17
 	github.com/cespare/cp v0.1.0
 	github.com/cockroachdb/pebble v0.0.0-20230209160836-829675f94811
 	github.com/davecgh/go-spew v1.1.1
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.20
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.10.0
-	github.com/ava-labs/avalanchego v1.10.18-rc.5.0.20240106000936-790b8a522a5c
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.20
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.10.0
-	github.com/ava-labs/avalanchego v1.10.18-rc.15
+	github.com/ava-labs/avalanchego v1.10.18-rc.5.0.20240105233604-41587de9c530
```
patrick-ogrady, line 7: will update before merging

plugin/evm/vm.go
```diff
@@ -141,15 +141,15 @@ const (
...
-	txGossipBloomMaxItems          = 8 * 1024
-	txGossipBloomFalsePositiveRate = 0.01
-	txGossipMaxFalsePositiveRate   = 0.05
-	txGossipTargetMessageSize      = 20 * units.KiB
-	maxValidatorSetStaleness       = time.Minute
-	txGossipThrottlingPeriod       = 10 * time.Second
-	txGossipThrottlingLimit        = 2
-	gossipFrequency                = 10 * time.Second
-	txGossipPollSize               = 10
+	txGossipBloomMinTargetItems          = 1024
```
patrick-ogrady, line 144: now that this scales up automatically, we don't need to set such a high minimum

plugin/evm/gossip.go
```diff
@@ -139,10 +139,11 @@ func (g *GossipEthTxPool) Subscribe(ctx context.Context) {
 			return
 		case pendingTxs := <-g.pendingTxs:
 			g.lock.Lock()
+			optimalSize := g.mempool.PendingSize() + len(pendingTxs.Txs) // should never come close to overflowing
```
patrick-ogrady, line 142: consider adding a churn multiplier to this to avoid frequent refresh

plugin/evm/vm.go
patrick-ogrady: We may want to change the name to `txGossipBloomMinTargetElements`

plugin/evm/vm.go
```diff
@@ -141,15 +141,16 @@ const (
 	atomicTxGossipProtocol = 0x1
 
 	// gossip constants
-	txGossipBloomMaxItems          = 8 * 1024
```
patrick-ogrady, line 144: I think this change is worth a discussion (ok to leave as is)

StephenButtolph: I'd prefer to leave it

patrick-ogrady: Updated

StephenButtolph (approved):

plugin/evm/vm.go
ceyonur: would keeping this value same for new `OptimalParameters` mean that we expect 8*1024 entries in the filter?

patrick-ogrady: This actually does the same thing but the previous variable name was confusing (as you pointed out 😉).

ceyonur (approved):

PR #453, opened 2024-01-08
Add check for compatible version of golangci-lint

marun:


scripts/lint.sh
```diff
@@ -4,4 +4,17 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+KNOWN_GOOD_VERSION="v1.54.2"
+# Upstream is compatible with v1.54.x at time of this writing, and
+# checking for this specific version is an attempt to avoid skew
+# between local and CI execution. The latest version (v1.55.1) seems to
+# cause spurious failures
+VERSION="$(golangci-lint --version | sed -e 's+golangci-lint has version \(v1.*\) built.*+\1+')"
+if [[ "${VERSION}" != "${KNOWN_GOOD_VERSION}" ]]; then
```
ceyonur, line 13: if this doesn't allow patch versions then CI might fail once there is a new version (i.e 1.54.3 etc). We should either specify full version in CI yml files or skip enforcing patch versions.

darioush (approved):

Issue #454, opened 2024-01-10
Add metrics for bloom filter reset


ceyonur:
We should add metric if bloom filter got reset (with targetElements)

PR #459, opened 2024-01-16
add bloom filter metrics

ceyonur:
## Why this should be merged

Bumps AvalancheGo and adds bloom filter metrics. 

Closes: #454 

## How this works

Passes registerer and namespace as input to new bloom filter functions.  

## How this was tested

Existing and AvalancheGo UTs should cover this.

### DIFF
--- a/scripts/tests.e2e.sh
+++ b/scripts/tests.e2e.sh
@@ -46,7 +46,6 @@ ref=""
 if git rev-parse --quiet --verify "origin/${avalanche_version}^{commit}" >/dev/null; then
   echo "checking out branch ${avalanche_version}"
   ref="origin/${avalanche_version}"
-  git checkout -B "test-${avalanche_version}" "origin/${avalanche_version}"
 else
   echo "checking out tag ${avalanche_version}"
   ref="${avalanche_version}"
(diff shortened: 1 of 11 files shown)


aaronbuchwald:
Looks like this needs to bump the AvalancheGo version in `scripts/versions.sh` and then this LGTM

darioush (approved):

scripts/tests.e2e.sh
```diff
@@ -30,15 +33,25 @@ trap cleanup EXIT
...
+# check if given reference is a tag or branch
+if git rev-parse --quiet --verify "origin/${avalanche_version}^{commit}" >/dev/null; then
+  echo "checking out branch ${avalanche_version}"
+  ref="origin/${avalanche_version}"
+else
+  echo "checking out tag ${avalanche_version}"
+  ref="${avalanche_version}"
+fi
+
+git checkout -B "test-${avalanche_version}" "${ref}"
```
StephenButtolph, line 54: Does it make sense for us to be targeting a branch for CI? It feels like we should be explicit and target either a tag or a commit imo.

StephenButtolph (approved):
code lgtm, not sure if we want to keep the script changes

scripts/tests.e2e.sh
ceyonur: reverted it

StephenButtolph (approved):

PR #461, opened 2024-01-21
Bloom filter reset

ceyonur:
## Why this should be merged

There is a bug that prevents pending txs to be added to bloom filter after reset.

## How this works

Change variables from pendingTxs to iterated txs 

## How this was tested

Added UT

## How is this documented

No need

### DIFF
--- a/plugin/evm/gossip_test.go
+++ b/plugin/evm/gossip_test.go
@@ -204,7 +204,9 @@ func setupPoolWithConfig(t *testing.T, config *params.ChainConfig, fundedAddress
 	}
 	chain, err := core.NewBlockChain(diskdb, core.DefaultCacheConfig, gspec, engine, vm.Config{}, common.Hash{}, false)
 	require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


darioush (approved):

StephenButtolph (approved):

plugin/evm/gossip_test.go
```diff
@@ -134,3 +146,59 @@ func TestGossipEthTxMarshaller(t *testing.T) {
...
+
+	// create eth txs
+	ethTxs := getValidEthTxs(key, 10, big.NewInt(226*params.GWei))
+
+	// Notify mempool about txs
+	errs := txPool.AddRemotesSync(ethTxs)
+	for _, err := range errs {
+		require.NoError(err, "failed adding subnet-evm tx to remote mempool")
+	}
+	time.Sleep(1 * time.Second)
```
StephenButtolph, line 181: `time.Sleep` makes me sad

PR #463, opened 2024-01-23
align types/block.go with subnet-evm

darioush:
## Why this should be merged
This PR moves extra methods defined on the block type (compared to subnet-evm) to its own file, and also separates out methods that create the block with ExtData so the no-ExtData methods are textually compatible with upstream & subnet-evm

## How this works
Refactoring

## How this was tested
CI


ethclient/ethclient.go
```diff
@@ -260,7 +260,7 @@ func (ec *client) getBlock(ctx context.Context, method string, args ...interface
 		}
 		txs[i] = tx.tx
 	}
-	return types.NewBlockWithHeader(head).WithBody(txs, uncles, body.Version, (*[]byte)(body.BlockExtraData)), nil
+	return types.NewBlockWithHeader(head).WithBody(txs, uncles).WithBodyExt(body.Version, (*[]byte)(body.BlockExtraData)), nil
```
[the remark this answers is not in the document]
darioush: I renamed it to WithExtData for consistency with other references to `ExtData`

ceyonur (approved):

core/genesis.go
```diff
@@ -307,7 +307,7 @@ func (g *Genesis) toBlock(db ethdb.Database, triedb *trie.Database) *types.Block
 			panic(fmt.Sprintf("unable to commit genesis block: %v", err))
 		}
 	}
-	return types.NewBlock(head, nil, nil, nil, trie.NewStackTrie(nil), nil, false)
```
StephenButtolph, line 310: I see places that were `nil, false` and places that were `nil, true` and neither place calls `WithExtData`. Is `recalc` unused when `nil` is provided? It seems like in places where `nil, true` is passed, we should be setting `header.ExtDataHash = EmptyExtDataHash` right?

StephenButtolph: I guess this only happens in tests

darioush: Yep, that's only in tests so I don't think it matters.

PR #471, opened 2024-01-25
Fix stat emitted for eth txs gossip received error

darioush:
## Why this should be merged
`gossip_atomic_received_error` has been emitted incorrectly for eth txs on error, so this PR adds a new stat for `gossip_eth_txs_received_error` and emits `gossip_atomic_received_error` when the atomic tx cannot be added to the mempool instead.

## How this works
Changes name of function and stat

## How this was tested
Inspection


plugin/evm/gossiper.go
```diff
@@ -554,7 +554,7 @@ func (h *GossipHandler) HandleEthTxs(nodeID ids.NodeID, msg message.EthTxsGossip
 			if err == txpool.ErrAlreadyKnown {
 				h.stats.IncEthTxsGossipReceivedKnown()
 			} else {
-				h.stats.IncAtomicGossipReceivedError()
```
ceyonur, line 557: We can still use and increase this after `h.vm.mempool.AddTx(&tx); err != nil ` if we'd like.

ceyonur (commented):
overall lgtm, just a nit.

plugin/evm/gossiper.go
darioush: Yeah, good point.

StephenButtolph (approved):

ceyonur (approved):

PR #476, opened 2024-01-26
consensus/dummy: reduce diff with subnet-evm

darioush:
## Why this should be merged
Adds mockable clock, adopts naming conventions, and adds ModeSkipCoinbase (as no-op) to dummy/consensus, to reduce number of files that coreth differs from subnet-evm.

## How this works
Refactoring

## How this was tested
CI


ceyonur (approved):
this is not needed by coreth but changes LGTM for alignment.

PR #477, opened 2024-01-27
Update TxPool base fee prior to filtering txs based on effective tip

darioush:
## Why this should be merged
This change allows transactions that have used a base fee that is less than the last fee update's estimate but greater than the actual built block's fee to be included in the block.
This is particularly useful when the base fee is decreasing (with time) and the client is using an estimate that is based on the current time, but the periodic update has not taken effect yet.

## How this works
Sets the value used to filter txs before filtering them

## How this was tested
CI

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -192,10 +192,7 @@ func (w *worker) commitNewWork(predicateContext *precompileconfig.PredicateConte
 	}
 
 	// Fill the block with all available pending transactions.
-	if header.BaseFee != nil {
-		w.eth.TxPool().SetBaseFee(header.BaseFee)
-	}
-	pending := w.eth.TxPool().Pending(true)
+	pending := w.eth.TxPool().PendingWithBaseFee(true, header.BaseFee)
 
 	// Split the pending transactions into locals and remotes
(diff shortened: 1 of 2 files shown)


miner/worker.go
```diff
@@ -192,6 +192,9 @@ func (w *worker) commitNewWork(predicateContext *precompileconfig.PredicateConte
 	}
 
 	// Fill the block with all available pending transactions.
+	if header.BaseFee != nil {
+		w.eth.TxPool().SetBaseFee(header.BaseFee)
```
ceyonur, line 196: Can this basefee be overriden by update basefee interval in txpool? I wonder if we should be delaying the interval after setting this here so that the basefee set here will be used.

ceyonur (approved):
LGTM. I think this solution avoids that race condition.

darioush:
> Was there another proposal that contained a race condition?

https://github.com/ava-labs/coreth/pull/477/commits/24a0c348da0fb67ac814c0042a90b123f0fcd95c was the original fix which has a race condition with the periodic updater changing the base fee. I think it's better that this is more accurate even if it messes up the code a little to thread through this parameter.

PR #497, opened 2024-02-26
[network] Remove Legacy Gossiper + Only Push Gossip from RPC

patrick-ogrady:
Related: https://github.com/ava-labs/avalanchego/pull/2772

This PR removes the legacy tx gossiper and switches to exclusively use the `AvalancheGo PushGossiper` (https://github.com/ava-labs/avalanchego/blob/master/network/p2p/gossip/gossip.go). Additionally, this PR updates `coreth` to only send push gossip for transactions submitted over RPC.

## TODO
- [x] restore commented out tests
- [x] all tests passing
- [x] ensure all artifacts from old gossip handling are cleaned up (ex: regossip stats)
- [x] ~Add submitted txs to bloom filter to avoid getting gossiped back what we gossiped~ Already handled: https://github.com/ava-labs/coreth/blob/9d9639b028914d39a5b62b836d0b487c745b8031/plugin/evm/gossip.go#L133-L165
- [ ] ~Handle the case where we restore local transactions from the journal (I don't think this is currently handled correctly for push gossip anyways, so we may want to punt to when we actually allow people to use the journal): https://github.com/ava-labs/coreth/blob/9d9639b028914d39a5b62b836d0b487c745b8031/core/txpool/txpool.go#L354-L364~ (#500)
- [ ] ~Consider adding an "executable" check to the mempool so we only gossip txs that can be used (i.e. don't gossip txs with nonces) -> may not work great under load because peers won't be given a "run" of transactions to include~ If we take a pessimistic view on our ability to determine the usefulness of any particular transaction, what we do now is the most robust approach.


eth/api_backend.go
```diff
@@ -324,7 +324,11 @@ func (b *EthAPIBackend) SendTx(ctx context.Context, signedTx *types.Transaction)
 	if err := ctx.Err(); err != nil {
 		return err
 	}
-	return b.eth.txPool.AddLocal(signedTx)
+	if err := b.eth.txPool.AddLocal(signedTx); err != nil {
```
patrick-ogrady, line 327: Was torn on whether this should be handled by the `txPool` or outside of it. I thought outside of it seemed cleaner/better separation of concerns.

plugin/evm/service.go
```diff
@@ -225,7 +225,11 @@ func (service *AvaxAPI) Import(_ *http.Request, args *ImportArgs, response *api.
 	}
 
 	response.TxID = tx.ID()
-	return service.vm.mempool.AddLocalTx(tx)
+	if err := service.vm.mempool.AddLocalTx(tx); err != nil {
```
patrick-ogrady, line 228: Add this as a method on `vm` (so can just register for PushGossip internally)

eth/api_backend.go
```diff
@@ -324,7 +324,11 @@ func (b *EthAPIBackend) SendTx(ctx context.Context, signedTx *types.Transaction)
...
+		return err
+	}
+	b.eth.gossiper.SendPushGossip(signedTx)
```
patrick-ogrady, line 330: May make sense to define a function on `eth` that performs both of these actions.

plugin/evm/service.go
patrick-ogrady: (likewise for the other instances in this file)

eth/backend.go
```diff
@@ -70,13 +70,18 @@ type Settings struct {
 	MaxBlocksPerRequest int64 // Maximum number of blocks to serve per getLogs request
 }
 
+type Gossiper interface {
+	SendPushGossip(*types.Transaction)
```
patrick-ogrady, line 74: If we change this to `Add`, we can just drop in the `ethTxPushGossiper`

plugin/evm/mempool.go
```diff
@@ -70,8 +67,6 @@ type Mempool struct {
 	// Pending is a channel of length one, which the mempool ensures has an item on
 	// it as long as there is an unissued transaction remaining in [txs]
 	Pending chan struct{}
-	// newTxs is an array of [Tx] that are ready to be gossiped.
-	newTxs []*Tx
```
patrick-ogrady, line 74: This would grow unbounded if not removed.

aaronbuchwald:
If we're going to make this change, we may want to enable LocalTxs by default as well. Otherwise, this will break anyone who is issuing transactions to their own node (using the defaults) and relying on remote gossip to gossip it to the network.

plugin/evm/gossiper_atomic_gossiping_test.go
patrick-ogrady, line 1: Removed for now because it was tightly coupled to the old way gossip worked...will find a way to add back in to ensure items added to the mempool are actually gossiped (which checks mempool)

patrick-ogrady: We may just want to ensure there is support for parsing txs (as that is all that is left).

plugin/evm/gossip.go
```diff
@@ -204,3 +212,13 @@ type GossipEthTx struct {
 func (tx *GossipEthTx) GossipID() ids.ID {
 	return ids.ID(tx.Tx.Hash())
 }
+
+// EthPushGossiper is used by the ETH backend to push transactions
+// issued over the RPC and added to the mempool to peers.
+type EthPushGossiper struct {
```
patrick-ogrady, line 218: This seemed more sane to me than adding an obscure `Add` function to `VM`

plugin/evm/tx_gossip_test.go
```diff
@@ -404,16 +403,6 @@ func TestEthTxPushGossipInbound(t *testing.T) {
 	inboundGossipMsg := append(binary.AppendUvarint(nil, ethTxGossipProtocol), inboundGossipBytes...)
 	require.NoError(vm.AppGossip(ctx, ids.EmptyNodeID, inboundGossipMsg))
 
-	forwardedMsg := &sdk.PushGossip{}
```
joshua-kim, line 407: We no longer forward txs

joshua-kim (approved):

plugin/evm/gossip.go
```diff
@@ -170,6 +172,12 @@ func (g *GossipEthTxPool) Add(tx *GossipEthTx) error {
 	return g.mempool.AddRemotes([]*types.Transaction{tx.Tx})[0]
 }
 
+// Has should just return whether or not the [txID] is still in the mempool,
+// not whether it is in the mempool AND pending.
+func (g *GossipEthTxPool) Has(txID ids.ID) bool {
+	return g.mempool.Has(common.Hash(txID[:]))
```
StephenButtolph, line 178:
```suggestion
	return g.mempool.Has(common.Hash(txID))
```

plugin/evm/gossip.go
```diff
@@ -204,3 +212,13 @@ type GossipEthTx struct {
 func (tx *GossipEthTx) GossipID() ids.ID {
 	return ids.ID(tx.Tx.Hash())
 }
+
+// EthPushGossiper is used by the ETH backend to push transactions
+// issued over the RPC and added to the mempool to peers.
+type EthPushGossiper struct {
+	vm *VM
```
StephenButtolph, line 219: We could probably put something more specific here than `*VM`.

plugin/evm/vm.go
```diff
@@ -706,6 +704,7 @@ func (vm *VM) initializeChain(lastAcceptedHash common.Hash) error {
 		node,
 		&vm.ethConfig,
 		vm.createConsensusCallbacks(),
+		&EthPushGossiper{vm: vm},
```
StephenButtolph, line 707: It's a bit dubious that `vm.ethTxPushGossiper` isn't initialized yet

patrick-ogrady: We now drop txs if not initialized.

StephenButtolph: I think that if check would be racy... but like... 🤷. I feel like we could just document the assumption that the API can't be hit before bootstrapping has finished.

StephenButtolph (approved):

darioush (approved):

plugin/evm/vm.go
```diff
@@ -144,6 +144,7 @@ const (
 	atomicTxGossipProtocol = 0x1
 
 	// gossip constants
+	pushGossipDiscardedElements          = 16_384
```
ceyonur, line 147:
```suggestion
	pushGossipDiscardedElements          = 16 * units.KiB
```

ceyonur (commented):
just few nits

plugin/evm/vm.go
patrick-ogrady: I already reverted this change. This is a count, not a size.

ceyonur: can we put a comment about why it is specifically this value

eth/api_backend.go
```diff
@@ -324,7 +324,14 @@ func (b *EthAPIBackend) SendTx(ctx context.Context, signedTx *types.Transaction)
...
 		return err
 	}
-	return b.eth.txPool.AddLocal(signedTx)
+	if err := b.eth.txPool.AddLocal(signedTx); err != nil {
+		return err
+	}
+
+	// We only enqueue transactions for push gossip if they were submitted over the RPC and
+	// added to the mempool.
+	b.eth.gossiper.Add(signedTx)
```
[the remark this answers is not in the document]
patrick-ogrady: I think this is a fair suggestion but I prefer how it is now.

patrick-ogrady:
Failing test:
```
panic: test timed out after 30m0s
running tests:
	TestEthTxPushGossipInbound (29m23s)
```

aaronbuchwald (approved):

PR #500, opened 2024-02-28
Remove unused journaling configs

StephenButtolph:
## Why this should be merged

Currently the txpool journal is disabled by explicitly setting `Journal = ""` in the `eth/backend`. This is fairly odd - as we explicitly support users to configure the `Journal`.

## How this works

- Removes the `Journal` config options and sets the default value to be disabled.
- Adds a comment that we should ensure that local transactions are gossiped if we re-enable journaling.

## How this was tested

- [X] CI

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -224,7 +224,6 @@ func New(
 
 	eth.bloomIndexer.Start(eth.blockchain)
 
-	config.TxPool.Journal = ""
 	eth.txPool = txpool.NewTxPool(config.TxPool, eth.blockchain.Config(), eth.blockchain)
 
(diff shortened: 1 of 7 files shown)


eth/backend.go
```diff
@@ -224,7 +224,6 @@ func New(
 
 	eth.bloomIndexer.Start(eth.blockchain)
 
-	config.TxPool.Journal = ""
```
StephenButtolph, line 227: This line effectively made all of the configuration options useless.

darioush (approved):

darioush:
docs: https://github.com/ava-labs/avalanche-docs/pull/1692

patrick-ogrady (approved):

PR #501, opened 2024-02-29
Update app gossip params

StephenButtolph:
Corresponding PR to: https://github.com/ava-labs/avalanchego/pull/2785

### DIFF
--- a/plugin/evm/config.go
+++ b/plugin/evm/config.go
@@ -35,6 +35,9 @@ const (
 	defaultMaxBlocksPerRequest                        = 0 // Default to no maximum on the number of blocks per getLogs request
 	defaultContinuousProfilerFrequency                = 15 * time.Minute
 	defaultContinuousProfilerMaxFiles                 = 5
+	defaultPushGossipNumValidators                    = 10
+	defaultPushGossipNumNonValidators                 = 0
+	defaultPushGossipNumPeers                         = 0
 	defaultPushGossipFrequency                        = 100 * time.Millisecond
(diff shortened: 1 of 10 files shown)


plugin/evm/config.go
```diff
@@ -150,10 +153,13 @@ type Config struct {
...
-	PushGossipFrequency Duration `json:"push-gossip-frequency"`
-	PullGossipFrequency Duration `json:"pull-gossip-frequency"`
-	RegossipFrequency   Duration `json:"regossip-frequency"`
-	TxRegossipFrequency Duration `json:"tx-regossip-frequency"` // Deprecated: use RegossipFrequency instead
+	PushGossipNumValidators    int      `json:"push-gossip-num-validators"`
+	PushGossipNumNonValidators int      `json:"push-gossip-num-non-validators"`
+	PushGossipNumPeers         int      `json:"push-gossip-num-peers"`
+	PushGossipFrequency        Duration `json:"push-gossip-frequency"`
+	PullGossipFrequency        Duration `json:"pull-gossip-frequency"`
+	RegossipFrequency          Duration `json:"regossip-frequency"`
```
patrick-ogrady, line 161: nit: would change the var name to `PushRegossipFreqency` and put above `PullGossipFrequency`

Might make sense to deprecate `regossip-frequency` too but we can probably leave all of this for future work.

StephenButtolph: Yeah I agree that would be more clean - but I think that would make sense to do in its own PR. Seems unrelated to this change

patrick-ogrady (approved):

PR #509, opened 2024-03-07
geth v1.12.2 update

darioush:


core/txpool/legacypool/legacypool.go
```diff
@@ -0,0 +1,2123 @@
...
+
+	// ErrTxPoolOverflow is returned if the transaction pool is full and can't accept
+	// another remote transaction.
+	ErrTxPoolOverflow = errors.New("txpool is full")
+)
+
+var (
+	evictionInterval      = time.Minute      // Time interval to check for evictable transactions
+	statsReportInterval   = 8 * time.Second  // Time interval to report transaction pool stats
+	baseFeeUpdateInterval = 10 * time.Second // Time interval at which to schedule a base fee update for the tx pool after SubnetEVM is enabled
```
ceyonur, line 82: I think there are few SubnetEVM,subnet-evm remainings in the code.

plugin/evm/gasprice_update.go
ceyonur, line 1: Do we need this? I think this is safe to remove and we can directly set them in init?

ceyonur (approved):

PR #511, opened 2024-03-08
Integrate stake weighted gossip selection

StephenButtolph:
See: https://github.com/ava-labs/avalanchego/pull/2835


darioush (approved):
should we validate this option so that it is not > 1 ?

StephenButtolph:
> should we validate this option so that it is not > 1 ?

It isn't required (the lib will handle it gracefully). But the value would be unexpected

dhrubabasu (approved):

PR #525, opened 2024-04-10
Wait for chain state updates to report atomic tx as Accepted

darioush:
## Why this should be merged
https://github.com/ava-labs/avalanchego/issues/2853

## How this works
Since chain state updates run asynchronously with VM block acceptance,
avoid returning [Accepted] until the chain state reaches the block
containing the atomic tx.

## How this was tested
CI


StephenButtolph:
I'm not sure this would resolve the avalanchego issue unless we update `GetAtomicTxStatus` as well.

ceyonur (approved):

PR #529, opened 2024-04-19
Beacon root check

ceyonur:
## Why this should be merged

Adds parent beacon root check for cancun + non-cancun cases

## How this works

In non-cancun parent beacon root must be nil
In cancun it should be an empty hash 

## How this was tested

Added UT


plugin/evm/vm_test.go
```diff
@@ -4094,3 +4096,121 @@ func TestSkipChainConfigCheckCompatible(t *testing.T) {
...
+			}()
+
+			importTx, err := vm.newImportTx(vm.ctx.XChainID, testEthAddrs[0], initialBaseFee, []*secp256k1.PrivateKey{testKeys[0]})
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			if err := vm.mempool.AddLocalTx(importTx); err != nil {
+				t.Fatal(err)
+			}
```
darioush, line 4170: should we use a different tx to build the block since we would need this in subnet-evm as well?

plugin/evm/vm_test.go
```diff
@@ -4094,3 +4096,121 @@ func TestSkipChainConfigCheckCompatible(t *testing.T) {
...
+			beaconRoot:    nil,
+			expectedError: true,
+			errString:     "header is missing parentBeaconRoot",
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			importAmount := uint64(1000000000)
+			// use Durango VM first
```
darioush, line 4152: why? (what does this comment mean?)

darioush (approved):

plugin/evm/vm_test.go
ceyonur: I used eth txs in subnet-evm

PR #530, opened 2024-04-20
add matrix and race to UTs

ceyonur:
Adds matrix to tests and use race for all UTs


plugin/evm/atomic_syncer_test.go
```diff
@@ -177,6 +177,7 @@ func TestAtomicSyncerResume(t *testing.T) {
 }
 
 func TestAtomicSyncerResumeNewRootCheckpoint(t *testing.T) {
+	t.Skip("FLAKY")
```
darioush, line 180: this is fixed in master

Issue #515, opened 2024-03-17
Pull gossip not working on coreth?


t-anyu:
Metrics API does not report any eth_tx_gossip ... type="pull" information. No received_bytes or received_count. Also no sent metrics being reported. Only avalanche_X_vm reports type="pull" gossip_received_bytes. And avalanche_C_vm reports type="push" metrics only.

Further digging it looks that vm.validators.Has(vm.ctx.NodeID) returns false. Which results ValidatorGossiper to return. Is this a bug? Or pull gossip is turned off or I have problem with my validator settings?

PR #543, opened 2024-05-02
Mark ourself as connected

StephenButtolph:
See: https://github.com/ava-labs/avalanchego/pull/2981

### DIFF
--- a/peer/network.go
+++ b/peer/network.go
@@ -523,7 +523,11 @@ func (n *network) Disconnected(ctx context.Context, nodeID ids.NodeID) error {
 		return nil
 	}
 
-	n.peers.Disconnected(nodeID)
+	if nodeID != n.self {
+		// The legacy peer tracker doesn't expect to be connected to itself.
+		n.peers.Disconnected(nodeID)
+	}
+
 	return n.p2pNetwork.Disconnected(ctx, nodeID)
 }
 
--- a/peer/network.go
+++ b/peer/network.go
@@ -506,7 +506,7 @@ func (n *network) Connected(ctx context.Context, nodeID ids.NodeID, nodeVersion 
 	}
 
 	if nodeID != n.self {
-		log.Debug("skipping registering self as peer")
+		// The legacy peer tracker doesn't expect to be connected to itself.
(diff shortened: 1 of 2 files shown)


JacobEverly (approved):

peer/network.go
```diff
@@ -505,12 +505,11 @@ func (n *network) Connected(ctx context.Context, nodeID ids.NodeID, nodeVersion
 		return nil
 	}
 
-	if nodeID == n.self {
+	if nodeID != n.self {
 		log.Debug("skipping registering self as peer")
-		return nil
+		n.peers.Connected(nodeID, nodeVersion)
 	}
```
lebdron, line 511: I had the same idea initially, but having `nodeID != n.self` and logging `skipping registering self as peer` might be confusing. I ended up with this.
```go
if nodeID != n.self {
    n.peers.Connected(nodeID, nodeVersion)
} else {
    log.Debug("skipping registering self as peer")
}
```

lebdron:
Should close https://github.com/ava-labs/coreth/issues/515

peer/network.go
StephenButtolph: I think there is some work to be done in the p2p sdk to standardize how it handles the registration of itself as a peer. So this PR is likely to change a bit.

StephenButtolph: But yeah as this PR is currently written that log is incorrect.

peer/network.go
```diff
@@ -505,12 +505,11 @@ func (n *network) Connected(ctx context.Context, nodeID ids.NodeID, nodeVersion
 		return nil
 	}
 
-	if nodeID == n.self {
-		log.Debug("skipping registering self as peer")
-		return nil
+	if nodeID != n.self {
+		// The legacy peer tracker doesn't expect to be connected to itself.
+		n.peers.Connected(nodeID, nodeVersion)
 	}
```
ceyonur, line 511: I guess it's safe but should we also skip disconnecting self in legacy peer tracker as well? (or at least we can add a comment)

ceyonur (approved):

PR #544, opened 2024-05-03
gossip/mempool: refactor Add in GossipEthTxPool and Mempool

lebdron:
## Why this should be merged

Refactor Add to implement the refactored interface https://github.com/ava-labs/avalanchego/pull/2986 , allowing not to require mutex acquire and release on every added transaction in gossip.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/ava-labs/coreth
 
-go 1.21.9
+go 1.21.10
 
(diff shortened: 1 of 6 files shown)


ceyonur:
I think the main avalanchego PR is closed https://github.com/ava-labs/avalanchego/pull/2986, shall we also close this one? @lebdron

PR #551, opened 2024-05-14
Generalize p2p sampling

joshua-kim:
## Why this should be merged
See https://github.com/ava-labs/avalanchego/pull/3014

## How this works
See https://github.com/ava-labs/avalanchego/pull/3014

## How this was tested
Unit tests


ceyonur:
is this still required?

joshua-kim:
Yeah this is still required... just blocked on avalanchego. This will be needed because we need to use the bandwidth-based sampling for state-sync (but not immediately). Marking as paused for now.

PR #557, opened 2024-05-20
state sync: initialize snapshots in bootstrap set mode

darioush:

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -1055,6 +1055,9 @@ func (vm *VM) SetState(_ context.Context, state snow.State) error {
 		if err := vm.StateSyncClient.Error(); err != nil {
 			return err
 		}
+		// Ensure snapshots are initialized before bootstrapping (i.e., if state sync is skipped).
+		// Note calling this function has no effect if snapshots are already initialized.
+		vm.blockChain.InitializeSnapshots()
 		return vm.fx.Bootstrapping()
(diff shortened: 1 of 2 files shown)


plugin/evm/vm.go
```diff
@@ -1055,6 +1055,9 @@ func (vm *VM) SetState(_ context.Context, state snow.State) error {
 		if err := vm.StateSyncClient.Error(); err != nil {
 			return err
 		}
+		// Ensure snapshots are initialized before bootstrapping (i.e., if state sync is skipped).
+		// Note calling this function has no effect if snapshots are already initialized.
+		vm.blockChain.InitializeSnapshots()
```
patrick-ogrady, line 1060: To ask a dumb question....we always enter bootstrapping at least once, right?

darioush: I believe so, but @StephenButtolph knows this kind of invariant better

plugin/evm/vm.go
```diff
@@ -1055,6 +1055,13 @@ func (vm *VM) SetState(_ context.Context, state snow.State) error {
 		if err := vm.StateSyncClient.Error(); err != nil {
 			return err
 		}
+		// After starting bootstrapping, do not attempt to resume a previous state sync.
+		if err := vm.StateSyncClient.ClearOngoingSummary(); err != nil {
```
patrick-ogrady, line 1059: 👍

PR #574, opened 2024-06-07
attach ref. counting to triedb instead of statedb

darioush:


core/blockchain.go
```diff
@@ -331,6 +331,7 @@ func NewBlockChain(
 	}
 	// Open trie database with provided config
 	triedb := trie.NewDatabase(db, cacheConfig.triedbConfig())
+	triedb.EnableRefCounting()
```
darioush, line 334: can move this to the config too

ceyonur: agreed

PR #580, opened 2024-06-17
[wip] blockchain interface exploration

darioush:
### Overview
Uses a slice of the existing blockchain implementation to implement the `plugin/vm` expected [interface](https://github.com/ava-labs/coreth/pull/578/files#diff-b6c8a21338dd428e899904f1cd63f3e77c6e5cf4d13d8fe379ed8a5154376e84R20).

### Storage
For Storage, I plugged in the pathdb implementation from *coreth* (which is totally untested other than UT) for now:
- It's already hash compatible with geth code
- Doesn't require any shim layers

However this is behind a [storage/state abstraction](https://github.com/ava-labs/coreth/pull/580/files#diff-1a8349b068c08360255a34cfc4b8a0bf356b81e909fddad361b6eed24df1c2e7R74) layer inspired by the one [here](https://github.com/ava-labs/hypersdk/discussions/815#discussion-6457170).

### Components
This gives a picture of the components that are needed:
- BlockChain 
- Mempool
- Miner
- Storage
- VM components (atomic txs, engine interface)

Using the set of interfaces between these components, we can either work on replacing each one, or moving the interfaces to have better component boundaries.
Note I re-used some config structs for convenience, those can be cleaned up separately and don't represent the ideal interface.

### builds on assumptions listed [here](https://github.com/ava-labs/coreth/pull/578#issue-2352266878)

_Side note: I have considered the path of refactoring coreth in a similar way, but that is not a primary goal and it will not work drop-in (API codepaths and some initialization will be broken)._

### Testing
✅  UTs
TODO: fuji / mainnet

### Definitely broken stuff
- Items listed [here](https://github.com/ava-labs/coreth/pull/572)
- State sync


v2/chain/blockchain_ext.go
```diff
@@ -0,0 +1,85 @@
+package chain
```
ARR4N, line 1: This is as far as I got. Marking it for when I start again.

PR #581, opened 2024-06-20
logging: truncate file dirs up to "coreth/" or ''coreth@version/"

darioush:
## Why this should be merged
Avoid printing logs like `/Users/.../go/pkg/mod/github.com/ava-labs/coreth@v0.13.5-rc.0/trie/triedb/hashdb/database.go:542`

## How this works
Looks for '/' after finding the prefix

## How this was tested
CI


plugin/evm/log_test.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestTrimPrefixes(t *testing.T) {
+	tests := []struct {
+		before string
+		after  string
+	}{
+		{"", ""},
+		{"/path/to/coreth/path/file.go", "path/file.go"},
```
ceyonur, line 18: I'm slightly in favor of keeping coreth in the path though, just to differentiate plugins from each other.

ceyonur (approved):

PR #586, opened 2024-06-25
remove pre-Durango repair scripts

darioush:
The atomic trie repair scripts were all released pre-Durango.
Nodes operating post-Durango should not require these repair scripts.


ARR4N (approved):
I don't have sufficient context to comment on overall effect of this, but the deletion looks (locally) ok to me.

ceyonur (approved):

PR #587, opened 2024-06-25
wip: separate atomic tx handling

darioush:


plugin/atx/vm.go
ceyonur, line 1: early comment: would it make sense to implement this as a FX?

darioush: where can I find the interface (or details) for FX?

ceyonur: https://github.com/ava-labs/avalanchego/blob/master/vms/secp256k1fx/fx.go
I think this is more or less the same thing though.

PR #593, opened 2024-07-02
Remove block.Status

StephenButtolph:
## Why this should be merged

This PR removes the usage of `block.Status()` from the `snowman.Block` interface.

## How this works

The only actual code changes are:
1. In atomic tx verification
2. In warp block hash signing

## How this was tested

- [X] CI

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -76,7 +76,6 @@ import (
 	"github.com/ava-labs/avalanchego/database/versiondb"
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/snow"
-	"github.com/ava-labs/avalanchego/snow/choices"
 	"github.com/ava-labs/avalanchego/snow/consensus/snowman"
 	"github.com/ava-labs/avalanchego/snow/engine/snowman/block"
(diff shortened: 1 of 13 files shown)


plugin/evm/vm.go
```diff
@@ -1378,6 +1383,33 @@ func (vm *VM) getBlock(_ context.Context, id ids.ID) (snowman.Block, error) {
...
+	}
+
+	lastAcceptedBlock := vm.LastAcceptedBlock()
+	height := blk.Height()
+	if lastAcceptedBlock.Height() < height {
+		// The provided block is not decided.
+		return nil, database.ErrNotFound
+	}
+
+	acceptedBlkID, err := vm.GetBlockIDAtHeight(ctx, height)
```
StephenButtolph, line 1401: Why does `GetBlockIDAtHeight` fetch the full block from disk rather than using `GetCanonicalHash`?

StephenButtolph, line 1401: Should we move the last accepted height check into `GetBlockIDAtHeight`? The implementation in coreth does not follow the avalanchego specification for this method.

darioush (approved):

ceyonur (approved):

PR #594, opened 2024-07-02
Update to latest p2p API

joshua-kim:
## Why this should be merged

Updates to https://github.com/ava-labs/avalanchego/pull/2753 version of the p2p API which adds a feature to respond with application-defined errors. This is needed as part of the interface update.

## How this works

Updates interface signatures

## How this was tested

CI

### DIFF
--- a/plugin/evm/gossip.go
+++ b/plugin/evm/gossip.go
@@ -10,14 +10,14 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/ethereum/go-ethereum/common"
+	ethcommon "github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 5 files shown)


plugin/evm/gossip.go
```diff
@@ -10,14 +10,15 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/ava-labs/avalanchego/ids"
-	"github.com/ava-labs/avalanchego/utils/logging"
```
joshua-kim, line 14: My IDE auto-fixed this diff... can undo if people want.

darioush (approved):

Issue #602, opened 2024-07-18
Adjust min fee for EUpgrade


darioush:
This item tracks modifying the min fee for the EUpgrade.

PR #604, opened 2024-07-18
eupgrade: lowering the base fee

darioush:
## Why this should be merged
The EUpgrade is considering lowering the base fee: https://github.com/ava-labs/coreth/issues/602

## How this works
- Switches fee calculation based on EUpgrade activated or not
- Modifies min fee on txpool at time of upgrade

## How this was tested
CI


plugin/evm/vm.go
```diff
@@ -703,12 +703,41 @@ func (vm *VM) initializeChain(lastAcceptedHash common.Hash) error {
...
+		return
+	} else if vm.chainConfig.IsEUpgrade(uint64(now.Unix())) {
+		// If EUpgrade is activated, set the min fee to the EUpgrade min fee
+		vm.txPool.SetMinFee(big.NewInt(params.EUpgradeMinBaseFee))
+		return
+	}
+
+	vm.txPool.SetMinFee(big.NewInt(params.ApricotPhase4MinBaseFee))
+	vm.shutdownWg.Add(1)
+	go func() {
```
darioush, line 727: I left it here since it is temporary anyway, can be factored out to its own file if others prefer.

params/avalanche_params.go
```diff
@@ -30,6 +30,7 @@ const (
 	ApricotPhase4BaseFeeChangeDenominator uint64 = 12
 	ApricotPhase5TargetGas                uint64 = 15_000_000
 	ApricotPhase5BaseFeeChangeDenominator uint64 = 36
+	EUpgradeMinBaseFee                    int64  = GWei
```
aaronbuchwald, line 33: nit: convert all min/max base fee values to use GWei?

params/config.go
```diff
@@ -540,7 +540,7 @@ type ChainConfig struct {
 	// Note: EIP-4895 is excluded since withdrawals are not relevant to the Avalanche C-Chain or Subnets running the EVM.
 	DurangoBlockTimestamp *uint64 `json:"durangoBlockTimestamp,omitempty"`
 	// EUpgrade on the Avalanche network. (nil = no fork, 0 = already activated)
-	// It activates Cancun (TODO) and modifies the min base fee (TODO).
+	// It activates Cancun and reduces the min base fee.
```
aaronbuchwald, line 543: nit: follow comment style from above ie. `EUpgrade activates...`

PR #606, opened 2024-07-23
Use `avax.GetAtomicUTXOs`

dhrubabasu:
## Why this should be merged

This code is duplicated from avalanchego. We can call the AvalancheGo helper directly.

## How this works

Calls `avax.GetAtomicUTXOs`

## How this was tested

CI

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -1722,40 +1722,15 @@ func (vm *VM) GetAtomicUTXOs(
 		limit = maxUTXOsToFetch
 	}
 
-	addrsList := make([][]byte, addrs.Len())
-	for i, addr := range addrs.List() {
-		addrsList[i] = addr.Bytes()
-	}
-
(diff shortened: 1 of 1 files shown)


plugin/evm/vm.go
```diff
@@ -1722,40 +1722,15 @@ func (vm *VM) GetAtomicUTXOs(
 		limit = maxUTXOsToFetch
 	}
 
-	addrsList := make([][]byte, addrs.Len())
-	for i, addr := range addrs.List() {
-		addrsList[i] = addr.Bytes()
-	}
-
-	allUTXOBytes, lastAddr, lastUTXO, err := vm.ctx.SharedMemory.Indexed(
+	return avax.GetAtomicUTXOs(
```
dhrubabasu, line 1725: https://github.com/ava-labs/avalanchego/blob/master/vms/components/avax/atomic_utxos.go#L25

darioush (approved):

PR #610, opened 2024-07-24
enables cancun at EUpgrade

darioush:
## Why this should be merged
https://github.com/avalanche-foundation/ACPs/pull/131

## How this works
Enables Cancun upgrade at the same time as the Avalanche "EUpgrade"

## How this was tested
CI

### DIFF
--- a/params/config_extra.go
+++ b/params/config_extra.go
@@ -129,3 +129,10 @@ func getUpgradeTime(networkID uint32, upgradeTimes map[uint32]time.Time) *uint64
 	// genesis.
 	return utils.NewUint64(0)
 }
+
+// SetEVMUpgrades sets the mapped upgrades (Avalanche > EVM upgrades) for the chain config.
+func (c *ChainConfig) SetEVMUpgrades() {
+	if c.EUpgradeTime != nil {
+		c.CancunTime = utils.NewUint64(*c.EUpgradeTime)
+	}
+}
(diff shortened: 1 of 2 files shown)


params/config_extra.go
```diff
@@ -129,3 +129,10 @@ func getUpgradeTime(networkID uint32, upgradeTimes map[uint32]time.Time) *uint64
 	// genesis.
 	return utils.NewUint64(0)
 }
+
+// SetEVMUpgrades sets the mapped upgrades (Avalanche > EVM upgrades) for the chain config.
+func (c *ChainConfig) SetEVMUpgrades() {
```
ARR4N, line 134: How have all the other upgrades been enabled historically? It immediately jumped out to me as unexpected that this function would only include one.

Maybe the function should be called `EnableCancunAtEUpgrade()`? Maybe it shouldn't exist at all (this feels like it wouldn't be the correct answer though).

ceyonur: They just haven't been defined. We used AvalancheGo upgrades directly instead of defining eth upgrades. This is the first time we do this.

ceyonur (approved):

params/config_extra.go
darioush: To clarify the above comment, historically we have replaced geth's upgrades (after Istanbul), with Avalanche upgrade (ApricotPhase... Banff... in coreth, and SubnetEVM in subnet-evm)

Starting with Cancun, we are keeping the behavior switch for eth forks as they are named in the original code, and only specifying the same timestamp.

In the future we should try to put back the original geth fork names to control eth fork behavior and retain the avalanche upgrades for avalanche behavior modifications (fee changes, enabling precompiles, etc)

PR #611, opened 2024-07-24
eupgrade/cancun: verify no blobs in header

darioush:
## Why this should be merged
It is safer to disallow blobs in block verification, since there is no builder/mempool.

## How this works
Enforces a 0 blob gas, which implies no blobs (and is easy to check from the header without a larger code change)

## How this was tested
CI

### DIFF
--- a/consensus/dummy/consensus.go
+++ b/consensus/dummy/consensus.go
@@ -264,7 +264,7 @@ func (self *DummyEngine) verifyHeader(chain consensus.ChainHeaderReader, header 
 			return err
 		}
 		if *header.BlobGasUsed > 0 { // VerifyEIP4844Header ensures BlobGasUsed is non-nil
-			return fmt.Errorf("blobs not enabled on avalanche networks: have %d, expected 0", *header.BlobGasUsed)
+			return fmt.Errorf("blobs not enabled on avalanche networks: used %d blob gas, expected 0", *header.BlobGasUsed)
 		}
 	}
 	return nil
--- a/consensus/dummy/consensus.go
+++ b/consensus/dummy/consensus.go
@@ -263,7 +263,7 @@ func (self *DummyEngine) verifyHeader(chain consensus.ChainHeaderReader, header 
 		if err := eip4844.VerifyEIP4844Header(parent, header); err != nil {
 			return err
 		}
-		if *header.BlobGasUsed > 0 { // VerifyEIP4844Header ensures excessBlobGas is non-nil
+		if *header.BlobGasUsed > 0 { // VerifyEIP4844Header ensures BlobGasUsed is non-nil
 			return fmt.Errorf("blobs not enabled on avalanche networks: have %d, expected 0", *header.BlobGasUsed)
 		}
 	}
--- a/consensus/dummy/consensus.go
+++ b/consensus/dummy/consensus.go
@@ -263,6 +263,9 @@ func (self *DummyEngine) verifyHeader(chain consensus.ChainHeaderReader, header 
 		if err := eip4844.VerifyEIP4844Header(parent, header); err != nil {
 			return err
 		}
+		if *header.BlobGasUsed > 0 { // VerifyEIP4844Header ensures excessBlobGas is non-nil
+			return fmt.Errorf("blobs not enabled on avalanche networks: have %d, expected 0", *header.BlobGasUsed)
+		}
 	}
 	return nil
 }
--- a/plugin/evm/vm_test.go
+++ b/plugin/evm/vm_test.go
@@ -4065,3 +4065,44 @@ func TestParentBeaconRootBlock(t *testing.T) {
 		})
 	}
 }
+
+func TestNoBlobsAllowed(t *testing.T) {
+	require := require.New(t)
+	importAmount := uint64(1000000000)
+	issuer, vm, _, _, _ := GenesisVMWithUTXOs(t, true, genesisJSONCancun, "", "", map[ids.ShortID]uint64{
+		testShortIDAddrs[0]: importAmount,
+	})
+	defer func() { require.NoError(vm.Shutdown(context.Background())) }()
+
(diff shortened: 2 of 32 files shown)


consensus/dummy/consensus.go
```diff
@@ -263,6 +263,9 @@ func (self *DummyEngine) verifyHeader(chain consensus.ChainHeaderReader, header
 		if err := eip4844.VerifyEIP4844Header(parent, header); err != nil {
 			return err
 		}
+		if *header.BlobGasUsed > 0 { // VerifyEIP4844Header ensures excessBlobGas is non-nil
```
darioush, line 266:
```suggestion
		if *header.BlobGasUsed > 0 { // VerifyEIP4844Header ensures BlobGasUsed is non-nil
```

consensus/dummy/consensus.go
```diff
@@ -263,6 +263,9 @@ func (self *DummyEngine) verifyHeader(chain consensus.ChainHeaderReader, header
 		if err := eip4844.VerifyEIP4844Header(parent, header); err != nil {
 			return err
 		}
+		if *header.BlobGasUsed > 0 { // VerifyEIP4844Header ensures BlobGasUsed is non-nil
```
ARR4N, line 266: I like the simplicity of this solution.

consensus/dummy/consensus.go
```diff
@@ -263,6 +263,9 @@ func (self *DummyEngine) verifyHeader(chain consensus.ChainHeaderReader, header
...
+			return fmt.Errorf("blobs not enabled on avalanche networks: have %d, expected 0", *header.BlobGasUsed)
```
ARR4N, line 267: This error is a bit misleading because it implies a number of blobs. `: used %d blob gas, expected 0"` is more precise.

consensus/dummy/consensus.go
ARR4N, line 1: (no action required) The package name `dummy` suggests a test double to me; am I incorrect? Trying to understand why this change is done in here.

ARR4N (changes requested):
Please add a test that results in an expected error because of blobs.

consensus/dummy/consensus.go
ceyonur: agreed, we should rename it

ceyonur, line 1: should we also check it in `SyntacticVerify`? Also I feel like we should probably handle this in mempool (it will be activated with cancun) and in worker (seems that currently they both will be enabled with cancun making it unable to build valid blocks)

darioush: Sure, only I suggest we do this in a separate PR after finishing the current geth update to avoid conflicts with wip.

darioush: Added to `SyntacticVerify`, if nothing else it is some defense in depth.

BlobTxs are handled by their own mempool, which we don't instantiate.
So the worker would not be getting any txs with blobs.

darioush: The fields will be populated:
https://github.com/ava-labs/coreth/blob/master/miner/worker.go#L183-L185

which sets the blob gas used to 0.

plugin/evm/vm_test.go
```diff
@@ -4090,6 +4090,39 @@ func TestParentBeaconRootBlock(t *testing.T) {
 	}
 }
 
+func TestNoBlobsAllowed(t *testing.T) {
+	require := require.New(t)
+	importAmount := uint64(1000000000)
+	issuer, vm, _, _, _ := GenesisVMWithUTXOs(t, true, genesisJSONCancun, "", "", map[ids.ShortID]uint64{
```
ARR4N, line 4096: Confirm that this doesn't return an error.

plugin/evm/vm_test.go
```diff
@@ -4090,6 +4090,39 @@ func TestParentBeaconRootBlock(t *testing.T) {
 	}
 }
 
+func TestNoBlobsAllowed(t *testing.T) {
```
ARR4N, line 4093: Is there a way to test this by actually sending a blob instead of by setting the blob gas used? I know it may seem pedantic, but I think it's good to be precise by testing the specific scenario of interest and not a proxy[^1] for it. Testing by mutating gas risks a future change breaking this if, for example, someone sets blob-gas price to zero because they assume it's not used.

[^1]: I'm completely ok with the _implementation_ checking for a proxy (gas used) of blobs being sent.

plugin/evm/vm_test.go
darioush: the ignored return value is not an error here.

plugin/evm/vm_test.go
darioush: Modified how the (invalid) block is created to include a blob tx

plugin/evm/vm_test.go
```diff
@@ -4091,35 +4093,46 @@ func TestParentBeaconRootBlock(t *testing.T) {
...
+		fee := big.NewInt(500)
+		fee.Add(fee, b.BaseFee())
+		tx, err := types.SignTx(types.NewTx(&types.BlobTx{
+			Nonce:      0,
+			GasTipCap:  uint256.NewInt(1),
+			GasFeeCap:  uint256.MustFromBig(fee),
+			Gas:        params.TxGas,
+			To:         testEthAddrs[0],
+			BlobFeeCap: uint256.NewInt(1),
+			BlobHashes: []common.Hash{{1}},
```
ARR4N, line 4116: (nit) I think it's worth putting a comment pointing out that this is what's important. Like:

```go
BlobHashes: []common.Hash{{1}}, // <----- The existence of this blob MUST cause an error
```

ARR4N (approved):

plugin/evm/vm_test.go
darioush: will apply this in a new PR to avoid voiding the approvals here.

PR #613, opened 2024-07-28
Refactor upgrade schedule

aaronbuchwald:
This PR corresponds to https://github.com/ava-labs/avalanchego/pull/3207


params/config_extra.go
```diff
@@ -120,12 +118,3 @@ func (c *ChainConfig) ToWithUpgradesJSON() *ChainConfigWithUpgradesJSON {
...
 	}
 }
-
-func getUpgradeTime(networkID uint32, upgradeTimes map[uint32]time.Time) *uint64 {
-	if upgradeTime, ok := upgradeTimes[networkID]; ok {
-		return utils.TimeToNewUint64(upgradeTime)
-	}
-	// If the upgrade time isn't specified, default being enabled in the
-	// genesis.
-	return utils.NewUint64(0)
```
aaronbuchwald, line 130: The previous behavior was to default to the zero timestamp, whereas the "default time" used for the local network in AvalancheGo uses a non-zero value in the past for forks that should be activated by default on the local network.

This is causing the upgrade test to fail due to a mismatched genesis from v1.11.0 to the new code here: https://github.com/ava-labs/avalanchego/actions/runs/10115885965/job/27977591879?pr=3207.

We can either update the last compatible version for the local network or keep the previous slightly divergent behavior between the activation time used by AvalancheGo and Coreth for local network forks that come pre-activated.

ceyonur (approved):

darioush (approved):

PR #617, opened 2024-08-02
nits, update comments

darioush:
Better comments and some minor refactoring


params/avalanche_params.go
ceyonur, line 1: shall we also change LaunchMinGasPrice&ApricotPhase1MinGasPrice  to gwei?

ceyonur (approved):

PR #620, opened 2024-08-06
fix: use `avalanchego` new `*test` packages

ARR4N:
## Why this should be merged

`avalanchego` was refactored and is already depending on one of the commits in this branch.

## How this works

Magic 🧙 Also _Find and Replace_, but mostly magic.

## How this was tested

Existing tests


ceyonur (approved):
should we also follow the same approach for testing packages in coreth?

changes LGTM

ARR4N:
> should we also follow the same approach for testing packages in coreth?
> 
> changes LGTM

Yup. IMO there should never be test-helper code in production.

The only reason I used tags at the beginning instead of doing this refactor was because they were a quick fix.

PR #630, opened 2024-08-09
get upgrade configs from context

ceyonur:
## Why this should be merged

AvalancheGo added upgrade config to snow context

Related AvalancheGo PR https://github.com/ava-labs/avalanchego/pull/3283

However we don't really need to change anything in Coreth since upgrades were written to chainConfig in genesis thus we have other source of information rather than upgrade configs in avalanchego/snow.Context. 

## How this works

* uses network ID rather than chain ID in config selection

## How this was tested

e2e should mostly cover this. I also did a local network testing with avalanchego to test out if network upgrades are passed correctly (but we don't actually use them)

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -464,8 +464,8 @@ func (vm *VM) Initialize(
 	default:
 		// TODO: This overrides the chain config in the given genesis (in genesisBytes)
 		// Do we want to do this?
-		config := params.GetChainConfig(chainCtx.NetworkUpgrades, new(big.Int).Set(g.Config.ChainID))
-		g.Config = config
+		// config := params.GetChainConfig(chainCtx.NetworkUpgrades, new(big.Int).Set(g.Config.ChainID))
+		// g.Config = config
 	}
 	// If the Durango is activated, activate the Warp Precompile at the same time
 	if g.Config.DurangoBlockTimestamp != nil {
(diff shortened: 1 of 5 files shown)


plugin/evm/vm.go
```diff
@@ -449,22 +449,31 @@ func (vm *VM) Initialize(
...
 		g.Config = &config
 		extDataHashes = fujiExtDataHashes
+	case avalanchegoConstants.LocalID:
+		config := *params.AvalancheLocalChainConfig
+		g.Config = &config
+	default:
+		// TODO: This overrides the chain config in the given genesis (in genesisBytes)
+		// Do we want to do this?
+		// config := params.GetChainConfig(chainCtx.NetworkUpgrades, new(big.Int).Set(g.Config.ChainID))
+		// g.Config = config
```
ceyonur, line 468: this would break many vm tests as vm tests pass their own upgrade config as part of the genesis and expect them to be activated. We need to either start using context.NetworkUpgrade in vm tests or let it be overridden by genesis config (at least for unit test network etc).

ceyonur: There is little to no gain respecting the context upgrades as we already have avalanchego upgrades in genesis chain config. The only possible gain is to stop unmarshaling Avalanche upgrades in Coreth ChainConfig, and let them only be passed from snow context (chain config in genesis json would not matter).

darioush (approved):

PR #631, opened 2024-08-09
Fix test tags

ceyonur:
## Why this should be merged

Test build tags was removed from AvalancheGo

## How this works

Removes build test tags from files and cmds

## How this was tested

e2e/local tests should cover this.


ARR4N (approved):
Btw, the test tags were replaced with specific `*test` packages as well as a linter to stop test-only code being imported in production: https://github.com/ava-labs/avalanchego/pull/3238

darioush (approved):

PR #634, opened 2024-08-13
Refactor params config

ceyonur:
## Why this should be merged

Refactors the param/config file to reduce the upstream diff and moves extra functions/variables to relative files.

## How this works

Creates a new struct `NetworkUpgrades` that wraps the Avalanche Network Upgrades which is then embedded into the `ChainConfig`. This is a refactor and nothing (at least in theory) should change. 

## How this was tested

Existing tests should cover this.


params/network_upgrades.go
```diff
@@ -1,25 +1,196 @@
-// (c) 2022, Ava Labs, Inc. All rights reserved.
+// (c) 2022, Ava Labs, Inn. All rights reserved.
```
darioush, line 1:
```suggestion
// (c) 2022, Ava Labs, Inc. All rights reserved.
```

darioush (approved):

PR #636, opened 2024-08-21
[wip] testing subnet-evm sync workflow

darioush:
don't merge it yet.


.github/workflows/sync-subnet-evm-branch.yml
```diff
@@ -17,4 +17,49 @@ jobs:
...
+          git checkout subnet-evm/coreth-diff-script -- ./scripts/apply_diff_and_rename.sh
+
+      - name: Apply diff
+        run: |
+          ./scripts/apply_diff_and_rename.sh subnet-evm/${{ github.event.inputs.remoteBranch }}
+
+      # TODO: Should this step be removed?
+      - name: Restore .github/workflows
+        run: |
+          git checkout -- .github/workflows
```
ceyonur, line 47: AFAIK it requires a special token if you want to override a workflow with a workflow

.github/workflows/sync-subnet-evm-branch.yml
ceyonur, line 1: This looks dope. Once upon a time I tried to come up with a similar workflow but ended up needing a special token. I think this script won't require token since it creates the PR in the same repo (coreth). What I had been previously trying to accomplish was:
* Once a PR is merged in Subnet-EVM master, create a subsequent PR in coreth by cherry-picking those commits. The workflow can poll from Coreth side or pushes from Subnet-EVM side (pushing requires a special token though).
* I preferred to open PR for single commits to make it easier to review since the context and diffs would be fewer than doing moving diffs in bulk. 
* This requires PRs to be merged in squashed commits, otherwise ends up having too much sync PRs.

I'm not saying we should have these, just sharing my previous ideas.

PR #649, opened 2024-09-16
fix modifying common.Big1

ceyonur:
Fixes modified common.Big1

Failing regression UT: https://github.com/ava-labs/coreth/actions/runs/10891616469/job/30222904746?pr=649

### DIFF
--- a/consensus/dummy/dynamic_fees_test.go
+++ b/consensus/dummy/dynamic_fees_test.go
@@ -580,5 +580,5 @@ func TestCalcBaseFeeRegression(t *testing.T) {
 
 	_, _, err := CalcBaseFee(params.TestChainConfig, parentHeader, timestamp)
 	require.NoError(t, err)
-	require.Equalf(t, common.Big1.Cmp(big.NewInt(1)), 0, "big1 should be 1, got %s", common.Big1)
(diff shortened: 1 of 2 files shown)


consensus/dummy/dynamic_fees_test.go
```diff
@@ -564,3 +565,21 @@ func TestDynamicFeesEtna(t *testing.T) {
...
+
+	parentHeader := &types.Header{
+		Time:    parentTimestamp,
+		GasUsed: 14_999_999,
+		Number:  big.NewInt(1),
+		BaseFee: big.NewInt(1),
+		Extra:   make([]byte, params.DynamicFeeExtraDataSize),
+	}
+
+	require.Equal(t, int64(1), common.Big1.Int64())
```
ARR4N, line 581: Both of the checks need to either assert `common.Big1.IsInt64()` or do a direct comparison with `common.Big1.Compare(big.NewInt(1)) == 0`.

If `IsInt64() == false` then the output of `Int64()` is undefined. It's improbable that it would be 1, but not impossible.

ARR4N (approved):
I don't need to re-review the requested change.

darioush (approved):

consensus/dummy/dynamic_fees_test.go
ceyonur: done, thanks

PR #651, opened 2024-09-18
remove CALLEX, BALANCEMC

darioush:
## Why this should be merged
These disabled opcodes were not used on fuji and mainnet

## How this works
Removes code

## How this was tested
CI, bootstrapping fuji & mainnet


ARR4N (approved):

ceyonur (approved):
🚀

PR #657, opened 2024-09-19
use upstream instruction sets

darioush:


core/contracts_stateful_test.go
```diff
@@ -0,0 +1,448 @@
...
+	genesisContractAddr         = common.HexToAddress("0x0100000000000000000000000000000000000000")
+	NativeAssetBalanceAddr      = vm.NativeAssetBalanceAddr
+	PackNativeAssetBalanceInput = vm.PackNativeAssetBalanceInput
+	NativeAssetCallAddr         = vm.NativeAssetCallAddr
+	PackNativeAssetCallInput    = vm.PackNativeAssetCallInput
+)
+
+type StateDB = vm.StateDB
+
+func TestStatefulPrecompile(t *testing.T) {
```
darioush, line 31: this test has to move out of core/vm, since we want to get the statedb with the hook configured.

core/contracts_stateful_test.go
```diff
@@ -0,0 +1,448 @@
...
+var (
+	big0                        = big.NewInt(0)
+	genesisContractAddr         = common.HexToAddress("0x0100000000000000000000000000000000000000")
+	NativeAssetBalanceAddr      = vm.NativeAssetBalanceAddr
+	PackNativeAssetBalanceInput = vm.PackNativeAssetBalanceInput
+	NativeAssetCallAddr         = vm.NativeAssetCallAddr
+	PackNativeAssetCallInput    = vm.PackNativeAssetCallInput
+)
+
+type StateDB = vm.StateDB
```
darioush, line 29: these are just temporary, we can do the replace in the text below before merging

PR #659, opened 2024-09-23
params: add BerlinBlock, LondonBlock, ShanghaiTime

darioush:
## Why this should be merged
Adds back BerlinBlock, LondonBlock, ShanghaiTime upgrades from upstream.
This is towards using more upstream packages and activating the desired behaviors via setting the Eth upgrade, instead of directly referencing the Avalanche upgrade names from predominantly upstream code.

This PR does not include modifying any of the Avalanche upgrade to the Eth upgrade, it only prepares the appropriate fields in the chain config.

## How this works
In Avalanche AP2 & AP3 upgrades were enabled by timestamp however the corresponding Eth upgrade (Berlin, London) enable via block number. Since these upgrades already occurred, we can use their block numbers for canonical chains and use 0 in tests.

## How this was tested
CI. (Note there is no behavioral changes for execution in this PR)

### DIFF
--- a/params/config.go
+++ b/params/config.go
@@ -65,6 +65,8 @@ var (
 		PetersburgBlock:     big.NewInt(0),
 		IstanbulBlock:       big.NewInt(0),
 		MuirGlacierBlock:    big.NewInt(0),
+		BerlinBlock:         big.NewInt(0),
+		LondonBlock:         big.NewInt(0),
 		NetworkUpgrades: NetworkUpgrades{
 			ApricotPhase1BlockTimestamp:     utils.NewUint64(0),
 			ApricotPhase2BlockTimestamp:     utils.NewUint64(0),
@@ -155,6 +157,7 @@ var (
 		PetersburgBlock:     big.NewInt(0),
 		IstanbulBlock:       big.NewInt(0),
 		MuirGlacierBlock:    big.NewInt(0),
+		BerlinBlock:         big.NewInt(0),
 		NetworkUpgrades: NetworkUpgrades{
 			ApricotPhase1BlockTimestamp:     utils.NewUint64(0),
 			ApricotPhase2BlockTimestamp:     utils.NewUint64(0),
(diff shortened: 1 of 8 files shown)


params/config.go
```diff
@@ -606,14 +632,29 @@ func (c *ChainConfig) IsIstanbul(num *big.Int) bool {
...
+}
+
+// IsCancun returns whether num is either equal to the Cancun fork time or greater.
 func (c *ChainConfig) IsCancun(num *big.Int, time uint64) bool {
-	return isTimestampForked(c.CancunTime, time)
+	return c.IsLondon(num) && isTimestampForked(c.CancunTime, time)
 }
 
-// IsVerkle returns whether time is either equal to the Verkle fork time or greater.
+// IsVerkle returns whether num is either equal to the Verkle fork time or greater.
```
darioush, line 655: it is like this upstream

params/config_extra.go
```diff
@@ -173,6 +176,25 @@ func GetChainConfig(agoUpgrade upgrade.Config, chainID *big.Int) *ChainConfig {
...
+		c.BerlinBlock = big.NewInt(184985) // https://testnet.snowtrace.io/block/184985?chainid=43113, AP2 activation block
+		c.LondonBlock = big.NewInt(805078) // https://testnet.snowtrace.io/block/805078?chainid=43113, AP3 activation block
+	} else if AvalancheMainnetChainID.Cmp(c.ChainID) == 0 {
+		c.BerlinBlock = big.NewInt(1640340) // https://snowtrace.io/block/1640340?chainid=43114, AP2 activation block
+		c.LondonBlock = big.NewInt(3308552) // https://snowtrace.io/block/3308552?chainid=43114, AP3 activation block
+	} else {
+		// In testing or local networks, we only support enabling Berlin and London prior
+		// to the initially active time. This is likely to correspond to an intended block
+		// number of 0 as well.
+		initiallyActive := uint64(upgrade.InitiallyActiveTime.Unix())
```
darioush, line 189: this is mostly because `params.GetChainConfig` is called from e2e tests (avalanchego) and specifies `initiallyActiveTime` for the genesis

params/config_extra.go
```diff
@@ -173,6 +176,25 @@ func GetChainConfig(agoUpgrade upgrade.Config, chainID *big.Int) *ChainConfig {
 		MuirGlacierBlock:    big.NewInt(0),
 		NetworkUpgrades:     getNetworkUpgrades(agoUpgrade),
 	}
+	if AvalancheFujiChainID.Cmp(c.ChainID) == 0 {
+		c.BerlinBlock = big.NewInt(184985) // https://testnet.snowtrace.io/block/184985?chainid=43113, AP2 activation block
+		c.LondonBlock = big.NewInt(805078) // https://testnet.snowtrace.io/block/805078?chainid=43113, AP3 activation block
```
ARR4N, line 181: Confirmed: https://github.com/ava-labs/avalanchego/blob/1b6288f5056fbf3682456e9e2ba89d3caf4d5cf7/upgrade/upgrade.go#L42

PR #663, opened 2024-09-26
Keep deprecated precompiles in cancun (as deprecated)

darioush:
## Why this should be merged
No one should really be calling these contracts, but if they are we can keep the current behavior to revert the call and avoid treating them as normal addresses.

## How this works
Adds the precompiles to the cancun set

## How this was tested
CI


core/state_transition_test.go
```diff
@@ -167,10 +167,25 @@ func TestNativeAssetContractCall(t *testing.T) {
...
+			want:    "",
+		},
 	}
 
-	for name, stTest := range phase6Tests {
+	for name, stTest := range tests {
 		t.Run(name, func(t *testing.T) {
+			cpy := *stTest.config
+			cpy.SetEthUpgrades()
+			stTest.config = &cpy
```
darioush, line 188: alternatively I can set the times directly in the chain configs.

ceyonur (approved):

PR #666, opened 2024-10-01
trie_prefetcher: alternate structure

darioush:
## Why this should be merged
The current trie_prefetcher is modified significantly compared to upstream, which makes it difficult to merge changes from upstream.
This PR aims to preserve the existing behavior but only with minor inline modifications to upstream code.

I would be open to other structures with the same idea.

## How this works
Uses a wrapper for DB to track the central worker pool.
Adds a couple hooks for stopping the background processes.

## How this was tested
CI

### DIFF
--- a/core/state/prefetcher_database.go
+++ b/core/state/prefetcher_database.go
@@ -1,0 +1,133 @@
+// (c) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package state
+
+import (
+	"sync"
+
+	"github.com/ava-labs/coreth/core/types"
+	"github.com/ava-labs/coreth/utils"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// ForPrefetchingOnly returns a new database that is only suitable for prefetching
+// operations. It will not be safe to use for any other operations.
+// Close must be called on the returned database when it is no longer needed
+// to wait on all spawned goroutines.
+func (*cachingDB) ForPrefetchingOnly(db Database, maxConcurrency int) Database {
+	return newPrefetcherDatabase(db, maxConcurrency)
+}
+
+type prefetcherDatabase struct {
+	Database
+
+	maxConcurrency int
+	workers        *utils.BoundedWorkers
+}
+
+func newPrefetcherDatabase(db Database, maxConcurrency int) *prefetcherDatabase {
+	return &prefetcherDatabase{
+		Database:       db,
+		maxConcurrency: maxConcurrency,
+		workers:        utils.NewBoundedWorkers(maxConcurrency),
+	}
+}
+
+func (p *prefetcherDatabase) OpenTrie(root common.Hash) (Trie, error) {
+	t, err := p.Database.OpenTrie(root)
+	return newPrefetcherTrie(p, t), err
+}
+
+func (p *prefetcherDatabase) OpenStorageTrie(stateRoot common.Hash, address common.Address, root common.Hash, trie Trie) (Trie, error) {
+	t, err := p.Database.OpenStorageTrie(stateRoot, address, root, trie)
+	return newPrefetcherTrie(p, t), err
+}
+
+func (p *prefetcherDatabase) CopyTrie(t Trie) Trie {
+	switch t := t.(type) {
+	case *prefetcherTrie:
+		return t.getCopy()
+	default:
+		return p.Database.CopyTrie(t)
+	}
+}
+
+func (p *prefetcherDatabase) Close() {
+	p.workers.Wait()
+}
+
+type prefetcherTrie struct {
+	p *prefetcherDatabase
+
+	Trie
+	copyLock sync.Mutex
+
+	copies chan Trie
+	wg     sync.WaitGroup
+}
+
+func newPrefetcherTrie(p *prefetcherDatabase, t Trie) *prefetcherTrie {
+	prefetcher := &prefetcherTrie{
+		p:      p,
+		Trie:   t,
+		copies: make(chan Trie, p.maxConcurrency),
+	}
+	prefetcher.copies <- prefetcher.getCopy()
+	return prefetcher
+}
+
+func (p *prefetcherTrie) Wait() {
+	p.wg.Wait()
+}
+
+func (p *prefetcherTrie) getCopy() Trie {
+	select {
+	case copy := <-p.copies:
+		return copy
+	default:
+		p.copyLock.Lock()
+		defer p.copyLock.Unlock()
+		return p.p.Database.CopyTrie(p.Trie)
+	}
+}
+
+func (p *prefetcherTrie) putCopy(copy Trie) {
+	select {
+	case p.copies <- copy:
+	default:
+	}
+}
+
+func (p *prefetcherTrie) GetAccount(address common.Address) (*types.StateAccount, error) {
+	p.wg.Add(1)
+	f := func() {
+		defer p.wg.Done()
+
+		tr := p.getCopy()
+		_, err := tr.GetAccount(address)
+		if err != nil {
+			log.Error("GetAccount failed in prefetcher", "err", err)
+		}
+		p.putCopy(tr)
+	}
+	p.p.workers.Execute(f)
+	return nil, nil // Note this result is never used by the prefetcher
+}
+
+func (p *prefetcherTrie) GetStorage(address common.Address, key []byte) ([]byte, error) {
+	p.wg.Add(1)
+	f := func() {
+		defer p.wg.Done()
+
+		tr := p.getCopy()
+		_, err := tr.GetStorage(address, key)
+		if err != nil {
+			log.Error("GetStorage failed in prefetcher", "err", err)
+		}
+		p.putCopy(tr)
+	}
+	p.p.workers.Execute(f)
+	return nil, nil // Note this result is never used by the prefetcher
+}
--- a/core/state/trie_prefetcher.go
+++ b/core/state/trie_prefetcher.go
@@ -28,16 +28,16 @@ package state
 
 import (
 	"sync"
-	"time"
 
 	"github.com/ava-labs/coreth/metrics"
-	"github.com/ava-labs/coreth/utils"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
 )
 
-// triePrefetchMetricsPrefix is the prefix under which to publish the metrics.
-const triePrefetchMetricsPrefix = "trie/prefetch/"
+var (
+	// triePrefetchMetricsPrefix is the prefix under which to publish the metrics.
+	triePrefetchMetricsPrefix = "trie/prefetch/"
+)
 
 // triePrefetcher is an active prefetcher, which receives accounts or storage
 // items and does trie-loading of them. The goal is to get as much useful content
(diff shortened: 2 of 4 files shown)


core/state/trie_prefetcher.go
```diff
@@ -50,91 +50,65 @@ type triePrefetcher struct {
...
-		largestLoad     int64
-	)
+	defer func() {
+		type closer interface {
+			Close()
+		}
+		if closer, ok := p.db.(closer); ok {
+			closer.Close()
+		}
+	}()
```
darioush, line 94: added compared to upstream

core/state/trie_prefetcher.go
```diff
@@ -293,348 +248,143 @@ type subfetcher struct {
...
 	}
-	return sf.to.skipCount()
-}
-
-func (sf *subfetcher) copies() int {
-	if sf.to == nil {
-		// Unable to open trie
-		return 0
+	if trie, ok := sf.trie.(waiter); ok {
+		trie.Wait()
```
darioush, line 314: added compared to upstream

core/state/trie_prefetcher_test.go
```diff
@@ -59,7 +59,8 @@ func filledStateDB() *StateDB {
 
 func TestCopyAndClose(t *testing.T) {
 	db := filledStateDB()
-	prefetcher := newTriePrefetcher(db.db, db.originalRoot, "", maxConcurrency)
+	prefetchDb := newPrefetcherDatabase(db.db, maxConcurrency)
```
darioush, line 62: to preserve existing UT behavior for now

core/state/trie_prefetcher.go
```diff
@@ -293,348 +248,143 @@ type subfetcher struct {
...
+					// Somebody wants a copy of the current trie, grant them
+					ch <- sf.db.CopyTrie(sf.trie)
+
+				default:
+					// No termination request yet, prefetch the next entry
+					if _, ok := sf.seen[string(task)]; ok {
+						sf.dups++
+					} else {
+						if len(task) == common.AddressLength {
+							sf.trie.GetAccount(common.BytesToAddress(task))
```
darioush, line 372: we could do something more explicit here instead of wrapping GetAccount, if desired.

core/state/prefetcher_database.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (p *prefetcherTrie) GetStorage(address common.Address, key []byte) ([]byte, error) {
+	p.wg.Add(1)
+	f := func() {
+		defer p.wg.Done()
+
+		tr := p.getCopy()
+		_, err := tr.GetStorage(address, key)
+		if err != nil {
+			log.Error("GetAccount failed in prefetcher", "err", err)
```
darioush, line 127:
```suggestion
			log.Error("GetStorage failed in prefetcher", "err", err)
```

core/state/trie_prefetcher.go
```diff
@@ -50,91 +50,68 @@ type triePrefetcher struct {
...
+	storageDupMeter   metrics.Meter
+	storageSkipMeter  metrics.Meter
+	storageWasteMeter metrics.Meter
 }
 
-func newTriePrefetcher(db Database, root common.Hash, namespace string, maxConcurrency int) *triePrefetcher {
+func newTriePrefetcher(db Database, root common.Hash, namespace string) *triePrefetcher {
+	if pdb, ok := db.(withPrefetcherDB); ok {
+		db = pdb.PrefetcherDB()
+	}
```
darioush, line 67: added compared to upstream

core/state/trie_prefetcher.go
```diff
@@ -293,348 +251,139 @@ type subfetcher struct {
...
+					// No termination request yet, prefetch the next entry
+					if _, ok := sf.seen[string(task)]; ok {
+						sf.dups++
+					} else {
+						tr := withPrefetcherDefaults{sf.trie}
+						if len(task) == common.AddressLength {
+							tr.PrefetchAccount(common.BytesToAddress(task))
+						} else {
+							tr.PrefetchStorage(sf.addr, task)
+						}
```
darioush, line 374: I made this explicit because I don't want to accidentally modify GetAccount / GetStorage to something that would return empty

core/state/trie_prefetcher.go
```diff
@@ -28,113 +28,85 @@ package state
...
+)
 
 // triePrefetcher is an active prefetcher, which receives accounts or storage
 // items and does trie-loading of them. The goal is to get as much useful content
 // into the caches as possible.
 //
 // Note, the prefetcher's API is not thread safe.
 type triePrefetcher struct {
-	db       Database               // Database to fetch trie nodes through
+	db       PrefetcherDB           // Database to fetch trie nodes through
```
darioush, line 48: changed compared to upstream

core/state/trie_prefetcher.go
```diff
@@ -28,113 +28,85 @@ package state
...
+	storageWasteMeter metrics.Meter
 }
 
-func newTriePrefetcher(db Database, root common.Hash, namespace string, maxConcurrency int) *triePrefetcher {
+func newTriePrefetcher(db Database, root common.Hash, namespace string) *triePrefetcher {
+	var pdb PrefetcherDB
+	pdb = withPrefetcherDefaults{db}
+	if db, ok := db.(withPrefetcherDB); ok {
+		pdb = db.PrefetcherDB()
+	}
```
darioush, line 69: changed compared to upstream

core/state/trie_prefetcher.go
```diff
@@ -28,113 +28,85 @@ package state
...
-	if p.fetches != nil {
-		return
-	}
-
-	// Collect stats from all fetchers
-	var (
-		storageFetchers int64
-		largestLoad     int64
-	)
+	defer p.db.Close()
```
darioush, line 92: added compared to upstream

core/state/trie_prefetcher.go
```diff
@@ -276,15 +224,20 @@ func (p *triePrefetcher) trieID(owner common.Hash, root common.Hash) string {
 // main prefetcher is paused and either all requested items are processed or if
 // the trie being worked on is retrieved from the prefetcher.
 type subfetcher struct {
-	p *triePrefetcher
-
-	db    Database       // Database to load trie nodes through
+	db    PrefetcherDB   // Database to load trie nodes through
```
darioush, line 227: modified compared to upstream

core/state/trie_prefetcher.go
```diff
@@ -293,348 +246,138 @@ type subfetcher struct {
...
-}
+	<-sf.term
 
-func (sf *subfetcher) copies() int {
-	if sf.to == nil {
-		// Unable to open trie
-		return 0
-	}
-	return sf.to.copies
+	sf.db.WaitTrie(sf.trie)
```
darioush, line 308: added compared to upstream

core/state/trie_prefetcher.go
```diff
@@ -293,348 +246,138 @@ type subfetcher struct {
...
+
+				default:
+					// No termination request yet, prefetch the next entry
+					if _, ok := sf.seen[string(task)]; ok {
+						sf.dups++
+					} else {
+						if len(task) == common.AddressLength {
+							sf.db.PrefetchAccount(sf.trie, common.BytesToAddress(task))
+						} else {
+							sf.db.PrefetchStorage(sf.trie, sf.addr, task)
```
darioush, line 367: modified. compared to upstream (made this more explicit to avoid overriding GetAccount and GetStorage)

darioush:
Adding do not merge label prior to verifying performance

core/state/prefetcher_database.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (*prefetcherDatabase) WaitTrie(t Trie) {
+	t.(*prefetcherTrie).Wait()
+}
+
+func (p *prefetcherDatabase) Close() {
+	p.workers.Wait()
+}
+
+type PrefetcherDB interface {
```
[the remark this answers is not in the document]
darioush: Moving it to the top.

The implementation will later stay in coreth but the interfaces (and trie_prefetcher modifications) will be part of libevm.

core/state/prefetcher_database.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func (p *prefetcherTrie) getCopy() Trie {
+	select {
+	case copy := <-p.copies:
+		return copy
+	default:
+		p.copyLock.Lock()
+		defer p.copyLock.Unlock()
+		return p.p.Database.CopyTrie(p.Trie)
```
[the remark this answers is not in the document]
darioush: putCopy is non-blocking and discards the copy if there is no room to put it back, because of the default clause, will add some comments.

core/state/prefetcher_database.go
```diff
@@ -0,0 +1,179 @@
...
+	switch t := t.(type) {
+	case *prefetcherTrie:
+		return t.getCopy()
+	default:
+		return p.Database.CopyTrie(t)
+	}
+}
+
+func (*prefetcherDatabase) PrefetchAccount(t Trie, address common.Address) {
+	t.(*prefetcherTrie).PrefetchAccount(address)
```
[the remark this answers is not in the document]
darioush: PrefetcherDB type is designed to minimize changes in trie_prefetcher so these methods are only called from the trie_prefetcher.
In this case, sf.trie is initialized with the return values of OpenTrie or OpenStorageTrie, which is the intended use of the PrefetcherDB type.

We could try to return another type from OpenTrie, however this lead to more changes in trie_prefetcher than I would consider "minimal", open to your suggestions.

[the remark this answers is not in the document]
darioush: Also existing code https://github.com/ava-labs/coreth/blob/master/core/state/database.go#L218-L225 couples the DB type to the trie type (CopyTrie). This code seems "gone" from the current versions of upstream which is very promising for the future

core/state/trie_prefetcher_extra_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+	root := types.EmptyRootHash
+	count := uint64(0)
+	block := uint64(0)
+
+	rootKey := []byte("root")
+	countKey := []byte("count")
+	blockKey := []byte("block")
+	got, err := levelDB.Get(rootKey)
+	if err == nil {
```
ARR4N, line 58: (nit, optional) Consider a comment to highlight the use of `err ==` here because it's so common to expect the error path to be indented.

core/state/trie_prefetcher_extra_test.go
```diff
@@ -0,0 +1,200 @@
...
+	statedb.SetNonce(address1, 1)
+	for i := 0; i < count/2; i++ {
+		key := make([]byte, 32)
+		value := make([]byte, 32)
+		rand.Read(key)
+		rand.Read(value)
+
+		statedb.SetState(address1, common.BytesToHash(key), common.BytesToHash(value))
+	}
+	statedb.SetNonce(address2, 1)
```
ARR4N, line 186: This is identical to the `address1` case so they can be collapsed into a `for _, addr := range []...{address1, address1} {` loop.

core/state/trie_prefetcher_extra_test.go
```diff
@@ -0,0 +1,200 @@
...
+	commit := func(levelDB ethdb.Database, snaps *snapshot.Tree, db Database) {
+		require.NoError(db.TrieDB().Commit(root, false))
+
+		for i := lastCommit + 1; i <= block; i++ {
+			require.NoError(snaps.Flatten(fakeHash(i)))
+		}
+		lastCommit = block
+
+		// update the tracking keys
+		err = levelDB.Put(rootKey, root.Bytes())
```
ARR4N, line 92: None of the `err` variables are necessary here and all function calls can just be wrapped in `require.NoError()`.

core/state/trie_prefetcher_extra_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+		// Commit every 10 blocks or on the last iteration
+		if block%10 == 0 || count >= uint64(wantKVs) {
+			commit(levelDB, snaps, db)
+			b.Logf("Root: %v, kvs: %d, block: %d (committed)", root, count, block)
+		}
+		if previous != root {
+			require.NoError(db.TrieDB().Dereference(previous))
+		} else {
+			panic("root did not change")
```
ARR4N, line 120: Why panic and not `b.Fatal()`? You don't really need the dump that comes with a panic.

core/state/trie_prefetcher_extra_test.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+	}
+	require.NoError(levelDB.Close())
+	b.Logf("Starting benchmarks")
```
ARR4N, line 124: Call `b.ResetTimer()` here as you've done a lot of setup work already.

core/state/trie_prefetcher_extra_test.go
```diff
@@ -0,0 +1,200 @@
...
+				if prefetchers > 0 {
+					db = WithPrefetcher(db, prefetchers)
+				}
+				getMetric := func(metric string) int64 {
+					meter := metrics.GetOrRegisterMeter(triePrefetchMetricsPrefix+namespace+"/storage/"+metric, nil)
+					return meter.Snapshot().Count()
+				}
+				startLoads := getMetric("load")
+				for i := 0; i < b.N; i++ {
+					_ = addBlock(db, snaps, updates, prefetchers)
```
ARR4N, line 145: You never use the returned value from `addBlock()` so it can be removed to clean up the `_ = ` in both places.

core/state/trie_prefetcher.go
```diff
@@ -332,6 +330,29 @@ func (sf *subfetcher) loop() {
 		}
 		sf.trie = trie
 	}
+	handleTask := func(task []byte) {
```
ARR4N, line 333: Is this just from upstream?

ARR4N (changes requested):

core/state/trie_prefetcher_extra_test.go
darioush: It's not needed as actual benchmarks are run in a sub-test via b.Run.

> A benchmark that calls Run at least once will not be measured itself and will be called once with N=1.

core/state/trie_prefetcher.go
darioush: I moved this from https://github.com/ava-labs/coreth/pull/666/files/7b9959050f89377e30b15957fd5081e012cd720c..96575b96ada7cc778f0a53c5045a9539f5ec530b#diff-58140c10803d5c33def41482e87d096548e88382aacc21aa7dbee1e918915870L361-L370 so I can call it in defer.

Using defer seems cleaner than to modify the channel logic in the loop.
As we would encounter a merge conflict for the code that handles the task since we modified GetAccount to PrefetchAccount  (to avoid wrapping GetAccount)

Here's another approach for example https://github.com/ava-labs/subnet-evm/commit/5a48a7d358e3cbe621f6fb7455826c635e193f40#diff-58140c10803d5c33def41482e87d096548e88382aacc21aa7dbee1e918915870R361 (but this does not support the original pattern).

core/state/prefetcher_database.go
```diff
@@ -0,0 +1,208 @@
...
+
+// withPrefetcherDefaults extends Database and implements PrefetcherDB by adding
+// default implementations for PrefetchAccount and PrefetchStorage that read the
+// account and storage slot from the trie.
+type withPrefetcherDefaults struct {
+	Database
+}
+
+func (withPrefetcherDefaults) PrefetchAccount(t Trie, address common.Address) {
+	_, _ = t.GetAccount(address)
```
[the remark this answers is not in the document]
darioush: I added this https://github.com/ava-labs/coreth/pull/666/commits/66e4674884c7b4b4e690bab2b583bc7ce752e073 however I don't really think it is helping with clarity. I'd slightly prefer to revert it and have the empty return signal that the callsite would be ignoring any errors so the callee should eg, log it if they care.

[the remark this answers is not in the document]
ARR4N: Yeah I'm happy if you revert and log instead.

core/state/trie_prefetcher_extra_test.go
ARR4N: Oh cool, TIL, thank you!

ARR4N (approved):

ceyonur (commented):
few nits, otherwise LGTM.

PR #667, opened 2024-10-02
fix: set eth upgrades in backwards compatible way

darioush:
## Why this should be merged
Current code does not function for mainnet/fuji.
When running a node with those chain IDs, the existing chain config will not have berlin/london & shanghai.
Therefore, we should apply the SetEthUpgrade anytime we unmarshal the config, including in the genesis,
prior to checking for compatibility.

## How this works
As explained.

## How this was tested
[x] Added UT
[x] Tested by using database from running node, then continuing with this version (fuji)

### DIFF
--- a/params/config_extra.go
+++ b/params/config_extra.go
@@ -40,6 +40,24 @@ type AvalancheContext struct {
 // code in place of their Ethereum counterparts. The original Ethereum names
 // should be restored for maintainability.
 func (c *ChainConfig) SetEthUpgrades() {
+	if c.ChainID != nil && AvalancheFujiChainID.Cmp(c.ChainID) == 0 {
+		c.BerlinBlock = big.NewInt(184985) // https://testnet.snowtrace.io/block/184985?chainid=43113, AP2 activation block
+		c.LondonBlock = big.NewInt(805078) // https://testnet.snowtrace.io/block/805078?chainid=43113, AP3 activation block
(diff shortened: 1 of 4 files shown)


params/config_extra.go
```diff
@@ -40,6 +40,24 @@ type AvalancheContext struct {
 // code in place of their Ethereum counterparts. The original Ethereum names
 // should be restored for maintainability.
 func (c *ChainConfig) SetEthUpgrades() {
+	if c.ChainID != nil && AvalancheFujiChainID.Cmp(c.ChainID) == 0 {
```
ceyonur, line 43: Can we rather check with network ID from chain context? This should be good but in practice there can be local genesis confs using prod chain IDs (even they should not be doing that).

darioush: I made this modification, however we should explicitly discourage use of coreth other than fuji/mainnet.
Everyone else will be better off using subnet-evm, as we have already deviated from preserving hypothetical historical processing (eg, if some local network used previous opcodes which we removed they will no longer function).

darioush: Also since we don't store the networkID, it creates the need to use the stored config with the new network ID, so I'm not very sure what we should do

ceyonur: hmm yea sorry I did not notice that. I think we can use the chainID only for block based eth upgrades if using network ID seems weird in genesis.

ceyonur (approved):

PR #670, opened 2024-10-08
fix chainID in TestWaitDeployedCornerCases

darioush:
## Why this should be merged
Prior to this change the tx is rejected because the chainID does not match the backend's expected number 1337.

## How this works
Uses the signer for the correct chain ID.

## How this was tested
CI


ceyonur (approved):

darioush:
it's kind of funny because this is also actually a flaky test b/c it doesn't wait for the tx pool properly, and I don't plan to fix this test just now in upstream, so I'll add it to the flake list

PR #673, opened 2024-10-14
Sync subnet evm

ceyonur:
Syncs Subnet-EVM up to e93eb1e7fd4948049282eab22137899fe9e6617b

### DIFF
--- a/accounts/abi/bind/bind_test.go
+++ b/accounts/abi/bind/bind_test.go
@@ -124,8 +124,7 @@ var bindTests = []struct {
 	},
 	// Test that named and anonymous inputs are handled correctly
 	{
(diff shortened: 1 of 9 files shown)


accounts/abi/bind/bind_test.go
```diff
@@ -124,7 +124,8 @@ var bindTests = []struct {
 	},
 	// Test that named and anonymous inputs are handled correctly
 	{
-		`InputChecker`, ``, []string{``},
+		`InputChecker`, ``,
```
darioush, line 127: could we revert the wspace changes in this file?

darioush (approved):

PR #675, opened 2024-10-21
pass extra bytes directly to precompiles

darioush:

### DIFF
--- a/params/hooks_libevm.go
+++ b/params/hooks_libevm.go
@@ -106,16 +106,12 @@ func makePrecompile(contract contract.StatefulPrecompiledContract) libevm.Precom
 		if err != nil {
 			panic(err) // Should never happen
 		}
-		var predicateResultsBytes []byte
-		if len(header.Extra) >= DynamicFeeExtraDataSize {
-			predicateResultsBytes = header.Extra[DynamicFeeExtraDataSize:]
-		}
 		accessableState := accessableState{
 			env: env,
 			blockContext: &BlockContext{
-				number:                env.BlockNumber(),
-				time:                  env.BlockTime(),
-				predicateResultsBytes: predicateResultsBytes,
+				number: env.BlockNumber(),
+				time:   env.BlockTime(),
+				extra:  header.Extra,
 			},
 		}
 		return contract.Run(accessableState, env.Addresses().Caller, env.Addresses().Self, input, suppliedGas, env.ReadOnly())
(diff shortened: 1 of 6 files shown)


precompile/contracts/warp/contract_warp_handler.go
```diff
@@ -61,7 +61,8 @@ func handleWarpMessage(accessibleState contract.AccessibleState, input []byte, s
 	warpIndex := int(warpIndexInput) // This conversion is safe even if int is 32 bits because we checked above.
 	state := accessibleState.GetStateDB()
 	predicateBytes, exists := state.GetPredicateStorageSlots(ContractAddress, warpIndex)
-	predicateResultsBytes := accessibleState.GetBlockContext().GetPredicateResultsBytes()
+	extra := accessibleState.GetBlockContext().Extra()
+	predicateResultsBytes := predicate.GetPredicateResultBytes(extra)
```
darioush, line 65: IMO, the logic in this file from line 63-72, should be moved somewhere else.

ie, this contract is called in the context of some block and some tx, so it should not have to specify the hash of the tx currently being executed to the statedb to query it for the predicates, then query the results from the block header.

precompile/contract/interfaces.go
```diff
@@ -71,9 +71,7 @@ type ConfigurationBlockContext interface {
 
 type BlockContext interface {
 	ConfigurationBlockContext
-	// GetPredicateResults returns an byte array result of verifying the predicates
-	// of the given block.
-	GetPredicateResultsBytes() []byte
+	Extra() []byte // Exta returns the Extra field of the block header
```
ARR4N, line 74: How about `HeaderExtra() []byte` and remove the comment? That way it's also implicitly documented at the call site.

precompile/contracts/warp/contract_warp_handler.go
ARR4N: Agreed. I'm happy to review that as part of this PR, or as a later one if you'd like. I'm approving in case you choose the latter so please ping me if you want a re-review.

ARR4N (approved):

params/hooks_libevm.go
```diff
@@ -241,6 +237,6 @@ func (b *BlockContext) Timestamp() uint64 {
 	return b.time
 }
 
-func (b *BlockContext) GetPredicateResultsBytes() []byte {
-	return b.predicateResultsBytes
+func (b *BlockContext) Extra() []byte {
```
ceyonur, line 240: I think this PR conflicts with my comment [here](https://github.com/ava-labs/coreth/pull/662#discussion_r1816857990) 

If we can extend the BlockContext here, why we avoid using full and parsed types here?

precompile/contracts/warp/contract_warp_handler.go
ceyonur, line 65: IMHO:
warp contract should not need to deal with parsing predicate results (we already parse them for verification). I'm looking at the upstream and seems they also don't include header in the block context. If we can be flexible in BlockContext (we should be), we don't actually need to abstract everything. I do think BlockContext should be the place where we can have custom stuff derived from block/header to present for the execution. 
warp contract should not know about extra at all
Predicaters does not need to be just for warp messaging, any other predicater should not re-implement all this parse-from-extra logic.
These are my observations after looking at just diffs, block context and warp handler. I haven't considered libevm or any other cyclic dependency. LMK if there is another concern that outweighs these considerations (especially parsing).

darioush: This approach https://github.com/ava-labs/coreth/pull/679 tries to follow the comment above, let me know your thoughts.

PR #678, opened 2024-10-24
Use upstream account packages

darioush:
## Why this should be merged
Less code to maintain.

## How this works
Since we took types.Transaction, some more interfaces can be unified, and we can drop this code from our codebase.

## How this was tested
CI


plugin/evm/syncervm_test.go
```diff
@@ -426,7 +426,7 @@ type syncVMSetup struct {
 	serverAppSender *enginetest.Sender
 
 	includedAtomicTxs []*Tx
-	fundedAccounts    map[*keystore.Key]*types.StateAccount
+	fundedAccounts    map[*utils.Key]*types.StateAccount
```
ARR4N, line 429: Since one of the fields in `utils.Key` is a pointer, this [doesn't guarantee uniqueness](https://go.dev/play/p/azN5EW1T39N). Is that ok?

sync/statesync/test_sync.go
```diff
@@ -116,7 +116,7 @@ func fillAccountsWithStorage(t *testing.T, serverDB ethdb.Database, serverTrieDB
 // returns the new trie root and a map of funded keys to StateAccount structs.
 func FillAccountsWithOverlappingStorage(
 	t *testing.T, trieDB *triedb.Database, root common.Hash, numAccounts int, numOverlappingStorageRoots int,
-) (common.Hash, map[*keystore.Key]*types.StateAccount) {
+) (common.Hash, map[*utils.Key]*types.StateAccount) {
```
ARR4N, line 119: See earlier re unique keys.

utils/key.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+type Key struct {
+	Address    common.Address
+	PrivateKey *ecdsa.PrivateKey
+}
+
+func NewKey(rand io.Reader) (*Key, error) {
```
ARR4N, line 19: Why accept the reader if it's always `crypto/rand.Reader` (at least in the code I can see)? I understand wanting stable keys in tests, but it doesn't look like that's being done and this risks someone may accidentally pass a non-secure source.

darioush: I think this is fine since it's the same in the existing code it was adapted from and also in ecdsa.GenerateKey
Additionally it's recommended to receive interfaces in golang as opposed to concrete types

plugin/evm/syncervm_test.go
darioush: It seems okay as randomly generated keys would likely not be duplicate and this change does not modify this part of the behavior

utils/key.go
ARR4N: > receive interfaces in golang as opposed to concrete types

I mean not receive anything at all and just use `crypto/rand.Reader` inside the function. I'm happy with whatever you decide here—consider my original question as just a prompt to spur thinking.

ARR4N (approved):

PR #679, opened 2024-10-29
predicate results: alternate approach (similar to master)

darioush:
## Why this should be merged
This PR is an alternative to https://github.com/ava-labs/coreth/pull/675
This PR makes the precompile interface as it is in master, i.e. with 
`GetPredicateResults(txHash common.Hash, precompileAddress common.Address)`

## How this works
This PR removes the dependency from predicators package to the params package by:
- Moving the logic related to DynamicFeeExtraDataSize out of the predicates package,
- Breaking the dependency on params.Rules by introducing an interface that can check for the existence of predicators. As an alternative to only this bullet point, we can move the function PreparePredicateStorageSlots 

After this dependency is removed, the parsing functions from predicates package can be used in params.

## How this was tested
CI


params/hooks_libevm.go
```diff
@@ -192,6 +196,6 @@ func (b *BlockContext) Timestamp() uint64 {
 	return b.time
 }
 
-func (b *BlockContext) GetPredicateResultsBytes() []byte {
-	return b.predicateResultsBytes
+func (b *BlockContext) GetPredicateResults(txHash common.Hash, precompileAddress common.Address) []byte {
```
ceyonur, line 199: should this return `predicate.Results`?

params/predicate_bytes.go
ceyonur, line 1: what is the cyclic import that prevents us to have these in predicate pkg?

precompile/contracts/warp/contract_warp_handler.go
```diff
@@ -61,13 +61,7 @@ func handleWarpMessage(accessibleState contract.AccessibleState, input []byte, s
...
 	state := accessibleState.GetStateDB()
 	predicateBytes, exists := state.GetPredicateStorageSlots(ContractAddress, warpIndex)
-	predicateResultsBytes := accessibleState.GetBlockContext().GetPredicateResultsBytes()
-	predicateResults, err := predicate.ParseResults(predicateResultsBytes)
-	if err != nil {
-		// Note this should never occur as the results are parsed in the block header.
-		return nil, remainingGas, fmt.Errorf("%w: error parsing predicate results", err)
-	}
-	result := predicateResults.GetPredicateResults(state.GetTxHash(), ContractAddress)
+	result := accessibleState.GetBlockContext().GetPredicateResults(state.GetTxHash(), ContractAddress)
```
ceyonur, line 64: looks way cleaner! 🧹

predicate/predicate_slots.go
```diff
@@ -5,19 +5,21 @@ package predicate
...
+type predicaters interface {
+	PredicaterExists(common.Address) bool
+}
+
 // PreparePredicateStorageSlots populates the the predicate storage slots of a transaction's access list
 // Note: if an address is specified multiple times in the access list, each storage slot for that address is
 // appended to a slice of byte slices. Each byte slice represents a predicate, making it a slice of predicates
 // for each access list address, and every predicate in the slice goes through verification.
-func PreparePredicateStorageSlots(rules params.Rules, list types.AccessList) map[common.Address][][]byte {
+func PreparePredicateStorageSlots(rules predicaters, list types.AccessList) map[common.Address][][]byte {
```
ceyonur, line 20: hmm could the same thing be used in CheckPredicates (/core/evm.go) if importing params also makes it complicated there?

params/hooks_libevm.go
ceyonur: or keep the byte in name?

params/hooks_libevm.go
```diff
@@ -171,16 +175,16 @@ func (a accessableState) Call(addr common.Address, input []byte, gas uint64, val
...
-	number                *big.Int
-	time                  uint64
-	predicateResultsBytes []byte
+	number           *big.Int
+	time             uint64
+	predicateResults *predicate.Results
 }
 
-func NewBlockContext(number *big.Int, time uint64, predicateResultsBytes []byte) *BlockContext {
+func NewBlockContext(number *big.Int, time uint64, predicateResults *predicate.Results) *BlockContext {
```
ceyonur, line 183: nit: we should probably rename this to PrecompileBlockContext

eth/tracers/api.go
```diff
@@ -967,8 +967,15 @@ func (api *API) TraceCall(ctx context.Context, args ethapi.TransactionArgs, bloc
...
-		blockContext := params.NewBlockContext(vmctx.BlockNumber, vmctx.Time, predicateBytes)
+		var predicateResults *predicate.Results
+		predicateBytes := params.GetPredicateResultBytes(vmctx.Header.Extra)
+		if len(predicateBytes) > 0 {
+			predicateResults, err = predicate.ParseResults(predicateBytes)
+			if err != nil {
+				return nil, fmt.Errorf("failed to parse predicate results: %w", err)
+			}
+		}
+		blockContext := params.NewBlockContext(vmctx.BlockNumber, vmctx.Time, predicateResults)
```
ceyonur, line 978: why do we need to create a new blockContext? `vmctx` should work here, no?

params/hooks_libevm.go
```diff
@@ -106,16 +107,19 @@ func makePrecompile(contract contract.StatefulPrecompiledContract) libevm.Precom
...
+		var predicateResults *predicate.Results
+		if predicateResultsBytes := GetPredicateResultBytes(header.Extra); len(predicateResultsBytes) > 0 {
+			predicateResults, err = predicate.ParseResults(predicateResultsBytes)
+			if err != nil {
+				panic(err) // Should never happen, as results are already validated in block validation
+			}
 		}
 		accessableState := accessableState{
 			env: env,
 			blockContext: &BlockContext{
```
ceyonur, line 119: why we cannot use vm.BlockContext directly from env? It should already have a parsed results (and other similar functions to implement required interface)

params/hooks_libevm.go
darioush: This is how it is in master so I left it like that for now.

params/predicate_bytes.go
darioush: predicates imports params which needs to import predicates again to parse the predicates to make the block context for the precompiles.

I'm somewhat hopeful that we can fix this by finding a better location for the libevm hooks other than "params", which we can address next.

predicate/predicate_slots.go
darioush: the core package already has to import params so it's not an issue there.

eth/tracers/api.go
darioush: ApplyUpgrades requires `contract.ConfigurationBlockContext`. However block itself implements that soooo we can just use that here which is much cleaner.

params/hooks_libevm.go
darioush: perhaps you are looking at the subnet-evm or coreth source code?
as the vm.BlockContext here would be from the upstream libevm code [here](https://github.com/ava-labs/libevm/blob/main/core/vm/evm.go#L64), which we modified to include the full header so we can access the extra bits (aka predicate results) here and pass it to the precompiles.

params/predicate_bytes.go
ceyonur: params in general is a really bad package for constants, and in our case it's even worse.

ceyonur (commented):
IMO this looks way good! left a comment other than LGTM.

ceyonur (approved):

PR #681, opened 2024-11-06
rename: ethereum/go-ethereum -> ava-labs/libevm

darioush:
## Why this should be merged
In golang, the proper way to maintain a forked library is to perform the textual rename of the module to its new (forked) location.

## How this works
Unfortunately, depending on both libevm and go-ethereum is not possible (even in transitive dependencies), as this will pull in duplicate symbols with CGO (similar to this issue https://github.com/cosmos/cosmos-sdk/issues/18232, not exactly)
This avalanchego [branch](https://github.com/ava-labs/avalanchego/pull/3526) performs the same rename and depends on c1ae67c to get rid of the transitive dependency to go-ethereum

## How this was tested
CI


go.mod
```diff
@@ -122,20 +122,16 @@ require (
...
+	google.golang.org/genproto/googleapis/api v0.0.0-20240604185151-ef581f913117 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20240827150818-7e3bb234dfed // indirect
+	google.golang.org/grpc v1.66.0 // indirect
 	gopkg.in/ini.v1 v1.67.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	rsc.io/tmplfunc v0.0.3 // indirect
 )
-
-replace github.com/ethereum/go-ethereum => github.com/ava-labs/go-ethereum v0.0.0-20241007222654-0752a11d4aee
```
michaelkaplan13, line 139: 🙌

michaelkaplan13 (approved):
LGTM, only question is if we should also update the  `//go:generate` directives that reference `go-ethereum`. I don't think it matters too much either way.

scripts/geth-allowed-packages.txt
ARR4N, line 1: Do we still need this to be enforced?

ARR4N (approved):
See my question about the geth allowlist.

scripts/geth-allowed-packages.txt
darioush: I think it is better to keep it since we still have some shared packages.

PR #687, opened 2024-11-15
pass block root via optional params

darioush:


core/state/snapshot/snapshot.go
```diff
@@ -117,6 +117,11 @@ var (
 	ErrNotConstructed = errors.New("snapshot is not constructed")
 )
 
+type Opt struct {
+	BlockHash  common.Hash
+	ParentHash common.Hash
+}
```
darioush, line 123: not sure where this should live or if it should have a different structure. would also be used for triedb update

PR #694, opened 2024-11-29
use libevm v1.13.14-0.1.0.rc-2

darioush:
### Dependency Updates:
* Updated `github.com/ava-labs/avalanchego` and `github.com/ava-labs/libevm` to newer versions in `go.mod`.

### Account Handling Improvements:
* Modified `IsMultiCoin` function to use `SlimAccount` instead of `StateAccount` in `core/types/state_account.go`.
* Simplified the `TestGenerateMultiCoinAccounts` function by removing the RLP decoding step and directly checking the `IsMultiCoin` status in `core/state/statedb_multicoin_test.go`.

### Import Cleanup:
* Removed the unused `rlp` import from `core/state/statedb_multicoin_test.go`.

### DIFF
--- a/core/state/statedb_multicoin_test.go
+++ b/core/state/statedb_multicoin_test.go
@@ -150,8 +150,8 @@ func TestGenerateMultiCoinAccounts(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if isMultiCoin := types.IsMultiCoin(snapAccount); !isMultiCoin {
-		t.Fatalf("Expected SnapAccount to return IsMultiCoin: true, found: %v", isMultiCoin)
+	if !types.IsMultiCoin(snapAccount) {
+		t.Fatalf("Expected SnapAccount to return IsMultiCoin: true, found: false")
 	}
 
 	NormalizeCoinID(&assetID)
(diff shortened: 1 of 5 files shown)


core/state/statedb_multicoin_test.go
```diff
@@ -147,16 +146,12 @@ func TestGenerateMultiCoinAccounts(t *testing.T) {
...
 		t.Fatal(err)
 	}
-	account := new(types.StateAccount)
-	if err := rlp.DecodeBytes(snapAccount, account); err != nil {
-		t.Fatal(err)
-	}
-	if !types.IsMultiCoin(account) {
-		t.Fatalf("Expected SnapAccount to return IsMultiCoin: true, found: %v", types.IsMultiCoin(account))
+	if isMultiCoin := types.IsMultiCoin(snapAccount); !isMultiCoin {
+		t.Fatalf("Expected SnapAccount to return IsMultiCoin: true, found: %v", isMultiCoin)
```
ARR4N, line 154: (nit) Including the value via %v is redundant as it's always false so it can be hard-coded, and the variable removed.

core/types/state_account.go
```diff
@@ -48,6 +48,6 @@ type isMultiCoin bool
 
 var IsMultiCoinPayloads = ethtypes.RegisterExtras[isMultiCoin]()
 
-func IsMultiCoin(a *StateAccount) bool {
-	return bool(IsMultiCoinPayloads.FromStateAccount(a))
+func IsMultiCoin(a *SlimAccount) bool {
```
ARR4N, line 51: (optional) You could use `types.ExtraPayloadCarrier` as the parameter type to accept either.

ARR4N (approved):

PR #700, opened 2024-12-11
 fix passing options to Call (for native asset historical correctness)

darioush:


qdm12 (approved):

accounts/abi/bind/bind_extra_test.go
```diff
@@ -0,0 +1,121 @@
...
+	name     string
+	contract string
+	bytecode []string
+	abi      []string
+	imports  string
+	tester   string
+	fsigs    []map[string]string
+	libs     map[string]string
+	aliases  map[string]string
+	types    []string
```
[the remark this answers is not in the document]
darioush: you got it right: it's geth-we-don't-want-to-touch-code

accounts/abi/bind/bind_extra_test.go
```diff
@@ -0,0 +1,121 @@
...
+}{
+	// This test checks that the NativeAssetCall proxies the caller address
+	// This behavior is disabled on the network and is only to test previous
+	// behavior. Note the test uses ApricotPhase2Config.
+	{
+		`GetSenderNativeAssetCall`,
+		`
```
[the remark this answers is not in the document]
darioush: I followed the style in bind_test.go, but if you feel strongly I can add the names

accounts/abi/bind/bind_extra_test.go
ARR4N, line 1: Why test it via the `bind` tests and not directly on an instance of `simulated.Backend`?

accounts/abi/bind/bind_extra_test.go
```diff
@@ -0,0 +1,121 @@
...
+			_, _, interactor, err := DeployGetSenderNativeAssetCall(auth, sim)
+			if err != nil {
+				t.Fatalf("Failed to deploy interactor contract: %v", err)
+			}
+			sim.Commit(false)
+			_, err = interactor.SetSender(
+				&bind.TransactOpts{
+					From: auth.From,
+					Signer: auth.Signer,
+					NativeAssetCall: &bind.NativeAssetCallOpts{
```
ARR4N, line 95: What effect does this field in the opts have? Does it make the first EVMInterpreter context of the call be to the NAC precompile instead of `interactor`? With my non-Avalanche Solidity hat on I thought that the test wasn't hitting the changed code because the standard behaviour of `interactor.SetSender` (i.e. with a vanilla `TransactOpts`) would be to just call the function on the contract.

darioush: Added some comment, also it is has a good explanation on the comment for `NativeAssetCall`'s definition.

I tried to call native asset directly but sim protects against calling methods with no bytecode and other tests are using `bind.NativeAssetCallOpts` as well.

accounts/abi/bind/bind_extra_test.go
darioush: updated

accounts/abi/bind/bind_extra_test.go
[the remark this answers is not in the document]
qdm12: Yap it would be nice, given how long those test cases are / a bit original with go code as a string etc.

accounts/abi/bind/bind_extra_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/ava-labs/coreth/params"
+	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/libevm/crypto"
+)
+
+// TestGetSenderNativeAssetCall checks that the NativeAssetCall proxies the
+// caller address This behavior is disabled on the network and is only to test
+// previous behavior. Note the test uses ApricotPhase2Config.
+func TestGetSenderNativeAssetCall(t *testing.T) {
+	// pragma solidity >=0.8.0 <0.9.0;
```
ARR4N, line 27: Remind me to show you my automated Solidity <-> Go bindings in my personal project! I'd be happy to open-source it if we feel like it's useful.

accounts/abi/bind/bind_extra_test.go
```diff
@@ -0,0 +1,111 @@
...
+	defer sim.Close()
+
+	// Deploy the get/setSender contract
+	_, _, interactor, err := bind.DeployContract(auth, parsedABI, bytecode, sim)
+	if err != nil {
+		t.Fatalf("Failed to deploy interactor contract: %v", err)
+	}
+	sim.Commit(false)
+
+	// Setting NativeAssetCall in the transact opts will proxy the call through
```
ARR4N, line 85: Now I understand why that old vulnerability arose

ARR4N (approved):

PR #702, opened 2024-12-11
refactor: params.ChainConfigExtra & RulesExtra -> params/extras

darioush:
Moves the `ChainConfigExtra` type to a new package which doesn't register libevm type configurations.


params/rules_extra.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 func GetRulesExtra(r Rules) *RulesExtra {
-	return extras.PointerFromRules(&r)
+	return payloads.PointerFromRules(&r)
 }
 
 type RulesExtra struct {
```
darioush, line 15: We could also move RulesExtra, but I don't see much advantage to doing it as it doesn't need to be externally serialized.

params/extras/config_extra.go
```diff
@@ -0,0 +1,261 @@
...
+
+type ChainConfigExtra struct {
+	NetworkUpgrades // Config for timestamps that enable network upgrades. Skip encoding/decoding directly into ChainConfig.
+
+	AvalancheContext `json:"-"` // Avalanche specific context set during VM initialization. Not serialized.
+
+	UpgradeConfig `json:"-"` // Config specified in upgradeBytes (avalanche network upgrades or enable/disabling precompiles). Skip encoding/decoding directly into ChainConfig.
+}
+
+func (c *ChainConfigExtra) CheckConfigCompatible(newcfg_ *ethparams.ChainConfig, headNumber *big.Int, headTimestamp uint64) *ConfigCompatError {
```
darioush, line 40: This fn should also pass newcfgExtra as the Extra type won't have access to the registered type getters.
Then the `// XXX` loophole can be removed.

params/extras/config_extra.go
```diff
@@ -0,0 +1,261 @@
...
+	// Config for enabling and disabling precompiles as network upgrades.
+	PrecompileUpgrades []PrecompileUpgrade `json:"precompileUpgrades,omitempty"`
+}
+
+// AvalancheContext provides Avalanche specific context directly into the EVM.
+type AvalancheContext struct {
+	SnowCtx *snow.Context
+}
+
+type ChainConfigExtra struct {
```
ARR4N, line 32: (nit) Since this is now in the `extras` package and will always be accessed like that, the `Extra` suffix is unnecessary.

`extras.ChainConfig` vs `extras.ChainConfigExtra`

params/extras/config_extra.go
ARR4N: The loophole being the inclusion of `GetExtra` above?

params/rules_extra.go
ARR4N: Hehe I had to go back and delete an earlier comment asking about this. The rationale I'd given for moving it is that there's then a logical grouping of related functionality within the `extras` package. I'll leave it up to you as it's subjective.

ARR4N (approved):

params/extras/config_extra.go
darioush: yep, if it's a separate type it can't call `GetExtra`

params/extras/config_extra.go
darioush: Sure, will rename it.

params/extras/config.go
```diff
@@ -0,0 +1,268 @@
...
+	// Alias ChainConfigExtra to avoid recursion
+	type _ChainConfigExtra ChainConfig
+	return json.Marshal(_ChainConfigExtra(*c))
+}
+
+type fork struct {
+	name      string
+	block     *big.Int // some go-ethereum forks use block numbers
+	timestamp *uint64  // Avalanche forks use timestamps
+	optional  bool     // if true, the fork may be nil and next fork is still allowed
```
qdm12, line 169: nit OOS/part-of-new-git-deltas: what does `the fork may be nil` mean?

params/extras/config.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+func (c *ChainConfig) CheckConfigForkOrder() error {
+	if c == nil {
+		return nil
+	}
```
qdm12, line 175: OOS/part-of-new-git-deltas: I see a few of these checks, and sometimes (i.e. MarshalJSON) it's not there. Does this ever happen that c is `nil`? If so, should we add missing checks?

params/extras/config.go
```diff
@@ -0,0 +1,268 @@
...
+	// the block number forks since it would not be a meaningful comparison.
+	// Instead, we check only that Phases are enabled in order.
+	// Note: we do not add the optional stateful precompile configs in here because they are optional
+	// and independent, such that the ordering they are enabled does not impact the correctness of the
+	// chain config.
+	if err := checkForks(c.forkOrder(), false); err != nil {
+		return err
+	}
+
+	return nil
```
qdm12, line 188: nit OOS/part-of-new-git-deltas: simplify to

```suggestion
	return checkForks(c.forkOrder(), false)
```

params/extras/config.go
```diff
@@ -0,0 +1,268 @@
...
+	optional  bool     // if true, the fork may be nil and next fork is still allowed
+}
+
+func (c *ChainConfig) CheckConfigForkOrder() error {
+	if c == nil {
+		return nil
+	}
+	// Note: In Avalanche, hard forks must take place via block timestamps instead
+	// of block numbers since blocks are produced asynchronously. Therefore, we do not
+	// check that the block timestamps in the same way as for
```
qdm12, line 178: OOS/part-of-new-git-deltas: that sentence has an issue, hoping this fixes it:

```suggestion
	// check the block timestamps forks in the same way as for
```

params/extras/config.go
```diff
@@ -0,0 +1,268 @@
...
+	// the block number forks since it would not be a meaningful comparison.
+	// Instead, we check only that Phases are enabled in order.
+	// Note: we do not add the optional stateful precompile configs in here because they are optional
+	// and independent, such that the ordering they are enabled does not impact the correctness of the
```
qdm12, line 182: OOS/part-of-new-git-deltas: hoping this fixes this line, but a better suggestion might be needed.

```suggestion
	// and independent, such that the ordering for which they are enabled does not impact the correctness of the
```

params/extras/config.go
```diff
@@ -0,0 +1,268 @@
...
+	// chain config.
+	if err := checkForks(c.forkOrder(), false); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// checkForks checks that forks are enabled in order and returns an error if not
+// [blockFork] is true if the fork is a block number fork, false if it is a timestamp fork
```
qdm12, line 192: nit OOS/part-of-new-git-deltas: a dot to stop the sentence does help reading

```suggestion
// checkForks checks that forks are enabled in order and returns an error if not.
// [blockFork] is true if the fork is a block number fork, false if it is a timestamp fork
```

params/extras/config.go
```diff
@@ -0,0 +1,268 @@
...
+
+				// Timestamp based forks can follow block based ones, but not the other way around
+				if lastFork.timestamp != nil && cur.block != nil {
+					return fmt.Errorf("unsupported fork ordering: %v used timestamp ordering, but %v reverted to block ordering",
+						lastFork.name, cur.name)
+				}
+			}
+		}
+		// If it was optional and not set, then ignore it
+		if !cur.optional || (cur.block != nil || cur.timestamp != nil) {
```
qdm12, line 226: nit OOS/part-of-new-git-deltas: no need for parentheses in this case

```suggestion
		if !cur.optional || cur.block != nil || cur.timestamp != nil {
```

params/extras/config.go
```diff
@@ -0,0 +1,268 @@
...
+			switch {
+			// Non-optional forks must all be present in the chain config up to the last defined fork
+			case lastFork.block == nil && lastFork.timestamp == nil && (cur.block != nil || cur.timestamp != nil):
+				if cur.block != nil {
+					return fmt.Errorf("unsupported fork ordering: %v not enabled, but %v enabled at block %v",
+						lastFork.name, cur.name, cur.block)
+				} else {
+					return fmt.Errorf("unsupported fork ordering: %v not enabled, but %v enabled at timestamp %v",
+						lastFork.name, cur.name, cur.timestamp)
+				}
```
qdm12, line 206: nit OOS/part-of-new-git-deltas: reduce indentation with

```suggestion
				}
				return fmt.Errorf("unsupported fork ordering: %v not enabled, but %v enabled at timestamp %v",
						lastFork.name, cur.name, cur.timestamp)
```

params/extras/config.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+		// If it was optional and not set, then ignore it
+		if !cur.optional || (cur.block != nil || cur.timestamp != nil) {
+			lastFork = cur
+		}
+	}
+	return nil
+}
+
+// Verify verifies chain config and returns error
```
qdm12, line 233: nit OOS/part-of-new-git-deltas: no need for the `and returns error` it's kind of obvious

```suggestion
// Verify verifies the chain config.
```

params/extras/config.go
```diff
@@ -0,0 +1,268 @@
...
+func (c *ChainConfig) IsPrecompileEnabled(address common.Address, timestamp uint64) bool {
+	config := c.GetActivePrecompileConfig(address, timestamp)
+	return config != nil && !config.IsDisabled()
+}
+
+// IsForkTransition returns true if [fork] activates during the transition from
+// [parent] to [current].
+// Taking [parent] as a pointer allows for us to pass nil when checking forks
+// that activate during genesis.
+// Note: this works for both block number and timestamp activated forks.
```
qdm12, line 253: nit OOS/part-of-new-git-deltas: that confused me for a minute, suggesting a more explicit phrasing

```suggestion
// Note: [parent] and [current] can be either both timestamp values, or both block number values, 
// since this function  works for both block number and timestamp activated forks.
```

qdm12 (approved):
Good job 👍 

All my comments are on a small things to change in existing code that got simply moved to params/extras, but which shows as "new code" by git (git doesn't detect the file movement); I  prefixed these comments with `OOS/part-of-new-git-deltas:` to highlight I'm aware this is not new new code.

How is our team feeling about this: are we ok to improve/fix small things when we have code showing as new git additions, despite it being out of scope with the PR (as long as it's not geth specific code)? I personally find this as a good way to improve code/comments occasionally, especially since we will likely not focus on reviewing all the code otherwise, but happy to not do this in the future if we feel it hurts us more than it helps us.

params/hooks_libevm.go
```diff
@@ -19,6 +20,13 @@ import (
 	"golang.org/x/exp/maps"
 )
 
+type RulesExtra extras.Rules
```
ARR4N, line 23: Why can't you register `extras.Rules` as the type?

darioush: We can do this: https://github.com/ava-labs/coreth/pull/708/files however I prefer as is since:
- The call to GetExtra to construct the `accessibleState` for the precompiles cannot be type safe (also doesn't seem there's a path to return an error easily, namely `PrecompileOverride` doesn't have an error return).
- The behaviors remain in the package that registers the type, which makes more sense to me. This wasn't feasible with ChainConfig since there was already significant uses of the methods outside of libevm

ARR4N (approved):

params/extras/config.go
```diff
@@ -0,0 +1,268 @@
...
+	ethparams "github.com/ava-labs/libevm/params"
+)
+
+// UpgradeConfig includes the following configs that may be specified in upgradeBytes:
+// - Timestamps that enable avalanche network upgrades,
+// - Enabling or disabling precompiles as network upgrades.
+type UpgradeConfig struct {
+	// Config for enabling and disabling precompiles as network upgrades.
+	PrecompileUpgrades []PrecompileUpgrade `json:"precompileUpgrades,omitempty"`
+}
```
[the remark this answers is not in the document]
darioush: `upgradeBytes` are the contents of `upgrades.json` in the chain configuration directory, passed to the VM during initialization and allows the network operators to coordinate a network upgrade without modifying code.
The first bullet point applies to subnet-evm but not coreth

params/extras/config.go
```diff
@@ -0,0 +1,268 @@
...
+		return newTimestampCompatError(
+			fmt.Sprintf("ChainConfig is not of the expected type *extras.ChainConfig, got %T", newcfg_.Hooks()),
+			utils.NewUint64(0),
+			nil,
+		)
+	}
+
+	// Check avalanche network upgrades
+	if err := c.checkNetworkUpgradesCompatible(&newcfg.NetworkUpgrades, headTimestamp); err != nil {
+		return err
```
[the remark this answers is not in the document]
darioush: This error should not be wrapped because it should be `ConfigCompatError` not `error`

params/extras/config.go
darioush: nil means the fork (aka upgrade) is not scheduled for activation.

params/extras/config.go
darioush: trying to improve this comment

params/extras/config.go
darioush: added comment about code from upstream.

darioush: added comment about code from upstream.

params/extras/config.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	return *x == *y
+}
+
+// ConfigCompatError is raised if the locally-stored blockchain is initialised with a
+// ChainConfig that would alter the past.
+type ConfigCompatError = ethparams.ConfigCompatError
+
+func newTimestampCompatError(what string, storedtime, newtime *uint64) *ConfigCompatError {
+	var rew *uint64
```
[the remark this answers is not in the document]
darioush: added comment about code taken from upstream.

params/extras/config.go
darioush: Let's tackle this one separately.

params/extras/config.go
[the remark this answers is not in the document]
qdm12: Indeed, my apologies for not seeing this!!

PR #704, opened 2024-12-13
alternative refactor/wrapper type for ChainConfigCompat

darioush:


ARR4N (commented):
Is this all so you can have access to the `GetExtra`?

For `var cc params.ChainConfig` you can also just use `ex, ok := cc.Hooks().(ChainConfigExtra)`, which is effectively what `GetExtra` is doing under the hood (with some invariant guarantees).

PR #719, opened 2024-12-26
feat(internal/ethapi): `historical-proof-query-window` option for archive nodes `eth_getProof`

### LINKED ISSUE #714: Deprecate historical proofs in `eth_getProof` API.
- Add a boolean VM configuration option like `historical-proofs` which defaults to `false`.
- In the API layer, reject calls to `eth_getProof` for blocks before `latest - tipBufferSize`, this is because with a path based trie, there is no good way to recall historical trie nodes and storing these nodes is a burden for validators. For most use cases (eg, message passing) it should suffice to verify state near the head of the chain, as often these use cases are time-sensitive.

Additional considerations / discussion topics:
- Entities with an interest in maintaining historical proofs can maintain their own indexes or use the hash db and modify the config option from its default.
- We can make the config option an integer which also controls `tipBufferSize`, which will also need to match the in-memory layers we keep at tip for the pathdb.
- Should we consider an upstream PR?

qdm12:
## Why this should be merged

Resolves #714 

## How this works

- [x] New option `historical-proof-query-window` for archive nodes only which:
  - can be set to a strictly positive number to accept get proof queries for a block number above the last accepted block number - `historical-proof-query-window`
  - defaults to `43200` to accept get proof queries for blocks dating from the last 24 hours approximately
  - can be set to `0` to accept any block number get proof query
- [x] For non-archive nodes, only the last 32 blocks before the accepted block are accepted for get proof queries (as before pretty much)

## How this was tested

- [x] `TestBlockChainAPI_stateQueryBlockNumberAllowed`

## Need to be documented?

- [x] https://github.com/ava-labs/avalanche-docs/pull/1996 - waiting on next avalanchego next release using coreth v0.14.1 (also still unreleased)

- [x] Done in RELEASES.md for the v0.14.1 release


internal/ethapi/api_test.go
```diff
@@ -2062,3 +2067,102 @@ func testRPCResponseWithFile(t *testing.T, testid int, result interface{}, rpc s
 	}
 	require.JSONEqf(t, string(want), string(data), "test %d: json not match, want: %s, have: %s", testid, string(want), string(data))
 }
+
+func TestBlockChainAPI_isAllowedToBeQueried(t *testing.T) {
+	t.Parallel()
```
qdm12, line 2072: nit I'm a `t.Parallel()` maniac. This with the test race detector did save me from time to time.

eth/ethconfig/config.go
```diff
@@ -70,6 +70,7 @@ func NewDefaultConfig() Config {
 		RPCEVMTimeout:             5 * time.Second,
 		GPO:                       DefaultFullGPOConfig,
 		RPCTxFeeCap:               1, // 1 AVAX
+		RecentBlocksWindow:        1024,
```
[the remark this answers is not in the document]
darioush: Can we plumb this configuration to control `tipBufferSize` here: https://github.com/ava-labs/coreth/blob/7a90507f3cbebfe85a65b4b26cf5383a87e951f3/core/state_manager.go#L49 as well?

Instead of being a const it can be taken in NewTrieWriter as a parameter or part of CacheConfig.

internal/ethapi/api.go
```diff
@@ -764,6 +770,43 @@ func (s *BlockChainAPI) GetProof(ctx context.Context, address common.Address, st
 	}, statedb.Error()
 }
 
+// isAllowedToBeQueried returns a nil error only if either:
+//   - the node is configured to accept historical proofs queries; or
+//   - the block number given is within the window of recent accepted
+//     blocks and not yet part of the historical blocks.
+//
+// Otherwise, a non-nil error is returned.
+func (s *BlockChainAPI) isAllowedToBeQueried(blockNumOrHash rpc.BlockNumberOrHash) (err error) {
```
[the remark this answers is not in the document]
darioush: let's make all additions. in files that are not shared with upstream code such as `internal/ethapi/api_extra.go`. This will help us handle upstream merges easier.
The same idea applies to test code as well.

internal/ethapi/api_test.go
```diff
@@ -2062,3 +2067,102 @@ func testRPCResponseWithFile(t *testing.T, testid int, result interface{}, rpc s
...
+	testCases := map[string]struct {
+		blockNumOrHash rpc.BlockNumberOrHash
+		makeBackend    func(ctrl *gomock.Controller) *MockBackend
+		wantErrMessage string
+	}{
+		"historical_proofs_accepted": {
+			blockNumOrHash: rpc.BlockNumberOrHashWithNumber(rpc.BlockNumber(0)),
+			makeBackend: func(ctrl *gomock.Controller) *MockBackend {
+				backend := NewMockBackend(ctrl)
+				backend.EXPECT().HistoricalConfig().Return(true, recentBlocksWindow)
```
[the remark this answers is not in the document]
darioush: I'm not too opinionated on this and defer to others.
Subjectively I would slightly prefer to use `testBackend` since it's already there.
In any case the new mocks need to be listed in `scripts/mocks.mockgen.txt`

internal/ethapi/api.go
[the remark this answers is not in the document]
qdm12: My apologies I did not check, I will in the future 👍 
Moved to api_extra.go and api_extra_test.go.

eth/ethconfig/config.go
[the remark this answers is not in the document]
qdm12: Done, however, shouldn't this be its own option? 🤔 Doesn't it feel a bit decoupled from the recent blocks window allowed for GetProofs? Plus, what should it be if the node has `historical-blocks` set to true??

license_header
```diff
@@ -0,0 +1,2 @@
+(c) 2019-2024, Ava Labs, Inc. All rights reserved.
```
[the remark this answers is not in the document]
qdm12: As per the mocks generation PR review, I removed the header;
however, whilst we're at it, this actually raises a good question: should we automate licensing of new files/new modified files? Perhaps something to track/do after merging coreth+subnet-evm

internal/ethapi/mocks_generate_test.go
```diff
@@ -1,3 +1,3 @@
 package ethapi
 
-//go:generate mockgen -package=$GOPACKAGE -destination=mocks_test.go . Backend
+//go:generate mockgen -package=$GOPACKAGE -destination=mocks_test.go -copyright_file=../../license_header . Backend
```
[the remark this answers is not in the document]
qdm12: Legal team will probably say yes 😆 But in reality I seriously doubt it as well.

plugin/evm/vm.go
```diff
@@ -545,6 +545,8 @@ func (vm *VM) Initialize(
 	vm.ethConfig.SnapshotDelayInit = vm.stateSyncEnabled(lastAcceptedHeight)
 	vm.ethConfig.SnapshotWait = vm.config.SnapshotWait
 	vm.ethConfig.SnapshotVerify = vm.config.SnapshotVerify
+	vm.ethConfig.HistoricalProofs = vm.config.HistoricalProofs
```
darioush, line 548: We should highlight this change in release notes as API providers for getProof may want to update their settings here.

eth/ethconfig/config.go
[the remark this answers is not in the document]
darioush: Let's make it a separate option as you suggest for maximum flexibility.
In the future these may be tangled as with the pathDB the # of states we keep in memory will determine what amount of proofs can be served.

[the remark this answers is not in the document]
qdm12: I would suggest either:

1. Have the tip buffer internally being set to
  - a fixed number (`1024`?) if `historical-proofs` is `true`
  - the value of `recent-blocks-window` if `historical-proofs` is `false`
2. Have a new option `trie-tip-buffer` defaulting to
  - a fixed number (`1024`?) if `historical-proofs` is `true`
  - the value of `recent-blocks-window` if `historical-proofs` is `false`

I would use `1.` for now, since we can do `2.` later with retro compatibility if a user needs such flexibility. My take is to give the least options possible to the user (less cognitive load, less prone to errors). If a user requests it, we can provide the option.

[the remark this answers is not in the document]
darioush: Since the tip buffer impacts memory usage, I would be cautious about a change from its current default of 32 --> 1024. This is why I prefer to either leave the tip buffer alone for now or to make it configurable separately without modifying its default value. In any case, if the API cannot access the state it will result in an error and the new setting of `historical-proofs` doesn't change that.

I think nodes that would serve historical proofs would also be archive nodes so the tip buffer is not applicable for them.

Let's also get thoughts from @ceyonur here.

internal/ethapi/api_extra.go
```diff
@@ -92,3 +92,40 @@ func (s *BlockChainAPI) GetBadBlocks(ctx context.Context) ([]*BadBlockArgs, erro
...
+	}
+
+	var oldestRecentNumber uint64
+	if lastNumber > recentBlocksWindow {
+		oldestRecentNumber = lastNumber - recentBlocksWindow
+	}
+	if number >= oldestRecentNumber {
+		return nil
+	}
+	return fmt.Errorf("block number %d is before the oldest non-historical block number %d (window of %d blocks)",
```
ceyonur, line 129: can we return an `bool, error`? this does look like function was successfull but actually returning the result rather than an error.

internal/ethapi/api_extra.go
```diff
@@ -92,3 +92,40 @@ func (s *BlockChainAPI) GetBadBlocks(ctx context.Context) ([]*BadBlockArgs, erro
...
+	lastBlock := s.b.LastAcceptedBlock()
+	lastNumber := lastBlock.NumberU64()
+
+	var number uint64
+	if blockNumOrHash.BlockNumber != nil {
+		number = uint64(blockNumOrHash.BlockNumber.Int64())
+	} else {
+		block, err := s.b.BlockByNumberOrHash(context.Background(), blockNumOrHash)
+		if err != nil {
+			return fmt.Errorf("getting block from hash: %s", err)
```
ceyonur, line 117: again this does not seem like an error but rather a process. can we just return the err or wrap it like "failed to get block: %w"?

also "%w" verb wraps the error [see](https://pkg.go.dev/fmt#Errorf)

internal/ethapi/api_extra.go
```diff
@@ -92,3 +92,40 @@ func (s *BlockChainAPI) GetBadBlocks(ctx context.Context) ([]*BadBlockArgs, erro
...
+
+// isHistoricalStateQueryAllowed returns a nil error only if:
+//   - the node is configured to accept historical proofs queries; or
+//   - the block number given is within the window of recent accepted
+//     blocks and not yet part of the historical blocks.
+//
+// Otherwise, a non-nil error is returned.
+func (s *BlockChainAPI) isHistoricalStateQueryAllowed(blockNumOrHash rpc.BlockNumberOrHash) (err error) {
+	historicalProofs, recentBlocksWindow := s.b.HistoricalConfig()
+	if historicalProofs {
```
ceyonur, line 104: hmm I was actually under impression that if this returns false we would not even look at the recentBlocksWindows config and just return the error/false so that we limit the API.

internal/ethapi/api_extra.go
qdm12: I explained why there: https://github.com/ava-labs/coreth/pull/719#discussion_r1897926695

(sorry conversation got resolved)

Otherwise, we would need to access a bunch of things at the caller level to have a useful enough error.

internal/ethapi/api_extra.go
qdm12: See our little discussion at https://github.com/ava-labs/libevm/pull/92#discussion_r1889940968 😉

internal/ethapi/api.go
```diff
@@ -688,7 +688,14 @@ func (n *proofList) Delete(key []byte) error {
 }
 
 // GetProof returns the Merkle-proof for a given account and optionally some storage keys.
+// If the requested block is part of historical blocks and the node does not accept
+// getting proofs for historical blocks, an error is returned.
 func (s *BlockChainAPI) GetProof(ctx context.Context, address common.Address, storageKeys []string, blockNrOrHash rpc.BlockNumberOrHash) (*AccountResult, error) {
+	err := s.isHistoricalStateQueryAllowed(blockNrOrHash)
+	if err != nil {
+		return nil, fmt.Errorf("checking if block is allowed to be queried: %w", err)
```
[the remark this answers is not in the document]
qdm12: See our little discussion at https://github.com/ava-labs/libevm/pull/92#discussion_r1889940968 😉 

It's to avoid cluttering the error logs. Otherwise you end up with N times "failed to" prefixes for an error returned from a call N times deep. And the error happened whilst we were processing ("checking ..." in this case), so it's not really misleading I think.

However I changed the error wrapping a bit to `fmt.Errorf("checking if possibly historical state query is allowed: %s", err)` to:

- match the renamed function name `isHistoricalStateQueryAllowed`
- Not wrap the error with `%w` and use `%s` instead - anyway %w is only useful for sentinel (or implementing wrap) errors and it's not used within `isHistoricalStateQueryAllowed` - if there is a need to programmatically react on this error, this should be changed however!

internal/ethapi/api_extra.go
qdm12: If `historicalProofs` is false, it means we don't support queries for historical blocks so we need to check the block is within the recent blocks window. Should we then rename `historical-proofs` to for example `historical-state-query-allowed` defaulting to false, to make things clearer?

eth/ethconfig/config.go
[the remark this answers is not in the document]
qdm12: - Previous options removed
- core.tipBufferSize kept constant at 32
- New option `historical-state-query-window` for archive nodes only which:
  - can be set to a strictly positive number to accept state queries for a block number above the last accepted block number - `historical-state-query-window`
  - defaults to `43200` to accept state queries for blocks dating from the last 24 hours approximately
  - can be set to `0` to accept any block number state query
- For non-archive nodes, `historical-state-query-window` is forced to the tip buffer size (now exported 🤷) of 32

plugin/evm/config.go
```diff
@@ -48,6 +48,7 @@ const (
 	defaultLogJSONFormat                          = false
 	defaultMaxOutboundActiveRequests              = 16
 	defaultPopulateMissingTriesParallelism        = 1024
+	defaultRecentBlocksWindow                     = 1024
```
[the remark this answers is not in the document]
qdm12: Set to 43200 (24 hours of blocks) for archive nodes, and to 32 (tip buffer size) for non-archive nodes.

plugin/evm/vm.go
qdm12: Done in RELEASES.md 😉

internal/ethapi/api_extra.go
qdm12: This is replaced now by:

https://github.com/ava-labs/coreth/pull/719/files#diff-7e84219e8a88f0a87eefdad20f859d8f4d6b142c5cba935a193a2904a658c5ecR103-R106

And the query window can only be `0` if the node is running in archive mode and the user specified it to be zero (accept all state queries whatever the block is)

eth/api_backend.go
```diff
@@ -58,15 +58,25 @@ type EthAPIBackend struct {
 	allowUnprotectedTxs      bool
 	allowUnprotectedTxHashes map[common.Hash]struct{} // Invariant: read-only after creation.
 	allowUnfinalizedQueries  bool
-	eth                      *Ethereum
-	gpo                      *gasprice.Oracle
+	// stateQueryWindow is the number of blocks before the last accepted block to be accepted for state queries.
+	// For archive nodes, it can be set to any strictly positive value OR 0 to indicate to accept any block query.
+	// For non-archive nodes, it MUST be set to the tip buffer size core.TipBufferSize.
+	stateQueryWindow uint64
```
qdm12, line 64: note to self: move below other fields to reduce deltas.

eth/ethconfig/config.go
```diff
@@ -55,21 +55,22 @@ var DefaultConfig = NewDefaultConfig()
...
+		SnapshotCache:              256,
+		AcceptedCacheSize:          32,
+		Miner:                      miner.Config{},
+		TxPool:                     legacypool.DefaultConfig,
+		BlobPool:                   blobpool.DefaultConfig,
+		RPCGasCap:                  25000000,
+		RPCEVMTimeout:              5 * time.Second,
+		GPO:                        DefaultFullGPOConfig,
+		RPCTxFeeCap:                1, // 1 AVAX
+		HistoricalStateQueryWindow: 43200,
```
qdm12, line 73: Since pruning is left to false, this should be for an archive node so I left the default to 43200... Should we not set this? 🤔 It seems like it's later overridden by the evm plugin config defaults anyway.

plugin/evm/config/config.go
```diff
@@ -283,6 +289,32 @@ func (c *Config) SetDefaults() {
 	c.AcceptedCacheSize = defaultAcceptedCacheSize
 }
 
+// UpdateWithDefaults updates the configuration with defaults depending
+// on fields already set.
+func (c *Config) UpdateWithDefaults() {
```
qdm12, line 294: a bit out of the topic, but I think the best way to handle config defaults is to have a config with each field zero value being **not valid** (nil or `0` or `""` etc. depending on the use). Then it should go as:

1. read from config sources and load fields that are set into the config
2. set the defaults on the config for any field that is still the zero value of that field. That can include a bit of logic for fields depending on other fields (such as in this case below).

So for example you would never have a field with a `bool` type, it would always be a `*bool` and you would default it if it's still nil after reading it from a config source(s).

plugin/evm/vm_test.go
```diff
@@ -405,6 +405,7 @@ func TestVMConfigDefaults(t *testing.T) {
 	vmConfig.SetDefaults()
 	vmConfig.RPCTxFeeCap = txFeeCap
 	vmConfig.EnabledEthAPIs = enabledEthAPIs
+	vmConfig.UpdateWithDefaults()
```
qdm12, line 408: I can include a call to `UpdateWithDefaults()` at the bottom of the body of `SetDefaults()` if you prefer.

eth/ethconfig/config.go
```diff
@@ -95,6 +96,10 @@ type Config struct {
 	SnapshotWait                    bool    // Whether to wait for the initial snapshot generation
 	SnapshotVerify                  bool    // Whether to verify generated snapshots
 	SkipSnapshotRebuild             bool    // Whether to skip rebuilding the snapshot in favor of returning an error (only set to true for tests)
+	// HistoricalStateQueryWindow is the number of blocks before the last accepted block to be accepted for state queries.
+	// For archive nodes, it defaults to 43200 and can be set to 0 to indicate to accept any block query.
+	// For non-archive nodes, it is forcibly set to the value of [core.TipBufferSize].
+	HistoricalStateQueryWindow uint64
```
darioush, line 102: Let's define this near AllowUnfinalizedQueries to keep avalanche modifications to the same area.

internal/ethapi/api.go
```diff
@@ -676,7 +676,14 @@ func (n *proofList) Delete(key []byte) error {
 }
 
 // GetProof returns the Merkle-proof for a given account and optionally some storage keys.
+// If the requested block is part of historical blocks and the node does not accept
+// getting proofs for historical blocks, an error is returned.
 func (s *BlockChainAPI) GetProof(ctx context.Context, address common.Address, storageKeys []string, blockNrOrHash rpc.BlockNumberOrHash) (*AccountResult, error) {
+	err := s.isHistoricalStateQueryAllowed(blockNrOrHash)
+	if err != nil {
+		return nil, fmt.Errorf("checking if possibly historical state query is allowed: %s", err)
```
darioush, line 684:
```suggestion
		return nil, fmt.Errorf("historical proof query not allowed: %s", err)
```

internal/ethapi/api_extra.go
darioush: seems more consistent with the code base to keep it "failed to: ..." or "could not ..." even if it's technically redundant language since it's an error.

eth/backend.go
```diff
@@ -267,6 +267,7 @@ func New(
 		allowUnprotectedTxs:      config.AllowUnprotectedTxs,
 		allowUnprotectedTxHashes: allowUnprotectedTxHashes,
 		allowUnfinalizedQueries:  config.AllowUnfinalizedQueries,
+		stateQueryWindow:         config.HistoricalStateQueryWindow,
```
darioush, line 270: nit: let's be consistent with names when we propagate something with a related name.

plugin/evm/vm.go
```diff
@@ -343,6 +343,8 @@ func (vm *VM) Initialize(
 			return fmt.Errorf("failed to unmarshal config %s: %w", string(configBytes), err)
 		}
 	}
+	vm.config.UpdateWithDefaults()
```
darioush, line 346: I see why the call to UpdateWithDefaults is added. Fixing the config scheme for the VM seems a bit out of scope here. Can we set it to `defaultBlocksWindow` for all nodes instead?
This change doesn't impact nodes with pruning as they will already fail the API requests with "state not found" or similar.

RELEASES.md
```diff
@@ -3,6 +3,10 @@
 ## [v0.14.1](https://github.com/ava-labs/coreth/releases/tag/v0.14.1)
 - Remove API eth_getAssetBalance that was used to query ANT balances (deprecated since v0.10.0)
 - Remove legacy gossip handler and metrics (deprecated since v0.10.0)
+- New option `historical-state-query-window` for archive nodes only which:
+  - can be set to a strictly positive number to accept state queries for a block number above the last accepted block number - `historical-state-query-window`
+  - defaults to `43200` to accept state queries for blocks dating from the last 24 hours approximately
```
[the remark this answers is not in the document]
qdm12: I changed it to:

> ⚠️ Archive node behavior change: by default, only the 24 hours preceeding the last accepted block are accepted for state proof queries. This can be adjusted with the new option `historical-proof-query-window` which defines the number of blocks before the last accepted block which should be accepted for state proof queries. This option can also be set to `0` to accept any block number state proof query.

eth/ethconfig/config.go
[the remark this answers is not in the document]
qdm12: Latest change is that the option is only used for archive nodes (pruning off), otherwise it's using the core.TipBufferSize internally. Resolving this, but feel free to unresolve/re-comment.

internal/ethapi/api.go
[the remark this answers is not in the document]
qdm12: This is changed to `fmt.Errorf("historical proof query not allowed: %s", err)` (suggested) - still not wrapping the error, but the context makes more sense I believe, resolving this.

internal/ethapi/api_extra.go
qdm12: This is further changed to:

```go
	queryWindow := uint64(core.TipBufferSize)
	if s.b.IsArchive() {
		queryWindow = s.b.HistoricalStateQueryWindow()
		if queryWindow == 0 {
			return nil
		}
	}
```

I hope this is clearer (hiding previous outdated comment)

RELEASES.md
```diff
@@ -1,6 +1,8 @@
 # Release Notes
 
 ## [v0.14.1](https://github.com/ava-labs/coreth/releases/tag/v0.14.1)
+
+- ⚠️ Archive node behavior change: by default, only the 24 hours preceeding the last accepted block are accepted for state proof queries. This can be adjusted with the new option `historical-proof-query-window` which defines the number of blocks before the last accepted block which should be accepted for state proof queries. This option can also be set to `0` to accept any block number state query.
```
darioush, line 5:
```suggestion
- IMPORTANT: `eth_getProof` calls for historical state will be rejected by default.
-- On archive nodes (`"pruning-enabled": false`): queries for historical proofs for state older than approximately 24 hours preceding the last accepted block will be rejected by default. This can be adjusted with the new option `historical-proof-query-window` which defines the number of blocks before the last accepted block which should be accepted for state proof queries, or set to `0` to accept any block number state query (previous behavior).
-- On `pruning` nodes: queries for proofs past the tip buffer (32 blocks) will be rejected. This is in support of moving to a path based storage scheme, which does not support historical state proofs.
```

plugin/evm/config/config.go
```diff
@@ -124,6 +127,10 @@ type Config struct {
 	PopulateMissingTriesParallelism int     `json:"populate-missing-tries-parallelism"` // Number of concurrent readers to use when re-populating missing tries on startup.
 	PruneWarpDB                     bool    `json:"prune-warp-db-enabled"`              // Determines if the warpDB should be cleared on startup
 
+	// HistoricalStateQueryWindow is, when running in archive mode only, the number of blocks before the
+	// last accepted block to be accepted for state queries.
+	HistoricalStateQueryWindow uint64 `json:"historical-proof-query-window,omitempty"`
```
darioush, line 132: nit: Should we update the name to match HistoricalProofQueryWindow ?

plugin/evm/vm.go
```diff
@@ -345,6 +345,7 @@ func (vm *VM) Initialize(
 			return fmt.Errorf("failed to unmarshal config %s: %w", string(configBytes), err)
 		}
 	}
+
```
darioush, line 348: nit: please undo unrelated wspace change

RELEASES.md
```diff
@@ -3,6 +3,10 @@
 ## [v0.14.1](https://github.com/ava-labs/coreth/releases/tag/v0.14.1)
 - Remove API eth_getAssetBalance that was used to query ANT balances (deprecated since v0.10.0)
 - Remove legacy gossip handler and metrics (deprecated since v0.10.0)
+- New option `historical-state-query-window` for archive nodes only which:
+  - can be set to a strictly positive number to accept state queries for a block number above the last accepted block number - `historical-state-query-window`
+  - defaults to `43200` to accept state queries for blocks dating from the last 24 hours approximately
+  - can be set to `0` to accept any block number state query
```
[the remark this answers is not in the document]
darioush: yes, once this PR is stable, please document it in this repo https://github.com/ava-labs/avalanche-docs/pull/1979/files (linked to a PR that deleted some API, but you would modify the relevant section instead)

darioush (approved):

RELEASES.md
darioush: was this change undone on a force push?

darioush (approved):

internal/ethapi/api_extra.go
```diff
@@ -92,3 +92,43 @@ func (s *BlockChainAPI) GetBadBlocks(ctx context.Context) ([]*BadBlockArgs, erro
...
+	queryWindow := uint64(core.TipBufferSize)
+	if s.b.IsArchive() {
+		queryWindow = s.b.HistoricalProofQueryWindow()
+		if queryWindow == 0 {
+			return nil
+		}
+	}
+
+	lastBlock := s.b.LastAcceptedBlock()
+	lastNumber := lastBlock.NumberU64()
```
ceyonur, line 112:
```suggestion
	lastAcceptedNumber := s.b.LastAcceptedBlock().NumberU64()
```

plugin/evm/config/config.go
```diff
@@ -59,6 +59,9 @@ const (
 	// - state sync time: ~6 hrs.
 	defaultStateSyncMinBlocks   = 300_000
 	DefaultStateSyncRequestSize = 1024 // the number of key/values to ask peers for per request
+
+	estimatedBlockAcceptPeriod = 2 * time.Second
+	defaultBlocksWindow        = uint64(24 * time.Hour / estimatedBlockAcceptPeriod)
```
ceyonur, line 64: nit: can we rename this to something like `defaultProofQueryWindow`

ceyonur (commented):
mostly nits

RELEASES.md
ceyonur: did we overwrite this suggestion with the old diff?

qdm12: Done in a65b60a97af5861fce1c81c24c35272aa2e6562d !
Just fixed the formatting to use ` -` instead of `--` which doesn't render fine (at least locally for me)

ceyonur (approved):

darioush (approved):

PR #720, opened 2024-12-26
add mockgen check for CI

darioush:
Copying this check from subnet-evm to ensure mocks are up to date.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,3 +123,16 @@ jobs:
         with:
           name: tmpnet-data
           path: ~/.tmpnet/networks/1000
+  mock_gen:
+    name: MockGen Check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.min_go_version }}
+          check-latest: true
+      - shell: bash
+        run: scripts/mock.gen.sh
+      - shell: bash
+        run: .github/workflows/check-clean-branch.sh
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -123,3 +123,16 @@ jobs:
         with:
           name: tmpnet-data
           path: ~/.tmpnet/networks/1000
+  mock_gen:
+    name: MockGen Check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
```
qdm12, line 131: We can use v5 😉 

```suggestion
      - uses: actions/setup-go@v5
```

.github/workflows/ci.yml
```diff
@@ -123,3 +123,16 @@ jobs:
...
+        with:
+          go-version: ${{ env.min_go_version }}
+          check-latest: true
```
qdm12, line 134: So we use the Go version specified and not the latest

```suggestion
```

.github/workflows/ci.yml
```diff
@@ -123,3 +123,16 @@ jobs:
...
+      - shell: bash
+        run: scripts/mock.gen.sh
+      - shell: bash
+        run: .github/workflows/check-clean-branch.sh
```
qdm12, line 138: Why not directly inline the script content in there + remove .github/workflows/check-clean-branch.sh (one less file is always a win 😄)

```suggestion
        run: |
            git update-index --really-refresh >> /dev/null
            git diff --exit-code
```

Also do we really need `git update-index --really-refresh >> /dev/null`? I never had the use case, I think `git diff --exit-code` should suffice to find diffs and exit with 1 if any diff is found.

qdm12 (commented):
Great, I was about to do it after our little mock conversation 😄 Thanks!

qdm12:
Bundled within #721 so closing this 😉

PR #723, opened 2024-12-28
Move atomic sync

ceyonur:
## Why this should be merged

Decouples atomic logic from syncervm server/client to atomic pkg.

## How this works

This pull request introduces a new atomic sync mechanism for the EVM plugin. The changes include the creation of new types and interfaces for atomic sync summaries, along with the implementation of the sync process. The most important changes are summarized below:

### Introduction of Atomic Sync Mechanism:

* **`plugin/evm/atomic/atomic_sync_extender.go`**: Added the `atomicSyncExtender` type and its methods to handle the synchronization process, including `Sync`, `OnFinishBeforeCommit`, and `OnFinishAfterCommit`.

* **`plugin/evm/atomic/atomic_sync_provider.go`**: Introduced the `atomicSyncProvider` type to provide state summaries at specific heights.

* **`plugin/evm/atomic/syncable.go`**: Defined the `AtomicBlockSyncSummary` type and related functions to parse and create sync summaries.

### Refactoring and Code Organization:

* **`plugin/evm/message/codec.go`**: Updated the codec manager to register the new sync summary type and introduced a variable to hold the sync summary type. [[1]](diffhunk://#diff-295ec2bc5b901db96bc911c27f3ef2df3eaf51e592081cab92d1dd51cbe372bcL18-R22) [[2]](diffhunk://#diff-295ec2bc5b901db96bc911c27f3ef2df3eaf51e592081cab92d1dd51cbe372bcL29-R33)

* **`plugin/evm/message/syncable.go`**: Refactored the `BlockSyncSummary` to implement the `Syncable` interface and added a parser for sync summaries.

### File and Type Renaming:

* **`plugin/evm/sync/syncervm_client.go`**: Renamed from `plugin/evm/syncervm_client.go` and updated to use the new `Syncable` interface and `atomicSyncExtender` for the sync process. [[1]](diffhunk://#diff-29367ce442e3b788023db853bd65e33d0b58fbb327dc8bbe9233b108499425fbL3-R3) [[2]](diffhunk://#diff-29367ce442e3b788023db853bd65e33d0b58fbb327dc8bbe9233b108499425fbL25) [[3]](diffhunk://#diff-29367ce442e3b788023db853bd65e33d0b58fbb327dc8bbe9233b108499425fbR48-R53) [[4]](diffhunk://#diff-29367ce442e3b788023db853bd65e33d0b58fbb327dc8bbe9233b108499425fbL67-R72) [[5]](diffhunk://#diff-29367ce442e3b788023db853bd65e33d0b58fbb327dc8bbe9233b108499425fbL77-R88) [[6]](diffhunk://#diff-29367ce442e3b788023db853bd65e33d0b58fbb327dc8bbe9233b108499425fbL134-R139) [[7]](diffhunk://#diff-29367ce442e3b788023db853bd65e33d0b58fbb327dc8bbe9233b108499425fbL156-R167) [[8]](diffhunk://#diff-29367ce442e3b788023db853bd65e33d0b58fbb327dc8bbe9233b108499425fbL172-R184) [[9]](diffhunk://#diff-29367ce442e3b788023db853bd65e33d0b58fbb327dc8bbe9233b108499425fbL291-R301) [[10]](diffhunk://#diff-29367ce442e3b788023db853bd65e33d0b58fbb327dc8bbe9233b108499425fbL323-R315) [[11]](diffhunk://#diff-29367ce442e3b788023db853bd65e33d0b58fbb327dc8bbe9233b108499425fbL338-R332) [[12]](diffhunk://#diff-29367ce442e3b788023db853bd65e33d0b58fbb327dc8bbe9233b108499425fbL356-R352)

## How this was tested

Existing UTs should cover

## Need to be documented?

No

No


plugin/evm/message/codec.go
```diff
@@ -15,7 +15,11 @@ const (
 	maxMessageSize = 2*units.MiB - 64*units.KiB // Subtract 64 KiB from p2p network cap to leave room for encoding overhead from AvalancheGo
 )
 
-var Codec codec.Manager
+var (
+	Codec codec.Manager
+	// TODO: Remove this once we have a better way to register types (i.e use a different codec version or use build flags)
```
ceyonur, line 20: we can move init to a function like (Initialize) and call in vm.Initialize like `Codec.Initialize(SyncSummaryType)`, but then tests should be doing the same thing everytime. But it might be the only way once we merge subnet-evm since two VMs should be using two different codecs since subnet-evm do not have atomic types (and thus codec has different order of type IDs).

plugin/evm/message/syncable.go
```diff
@@ -14,78 +14,107 @@ import (
...
 	BlockRoot   common.Hash `serialize:"true"`
-	AtomicRoot  common.Hash `serialize:"true"`
 
 	summaryID  ids.ID
 	bytes      []byte
-	acceptImpl func(SyncSummary) (block.StateSyncMode, error)
+	acceptImpl AcceptImplFn
+}
+
+type BlockSyncSummaryParser struct{}
```
ceyonur, line 44: these are not in use currently (but will be with subnet-evm merge).

plugin/evm/vm.go
```diff
@@ -617,11 +618,8 @@ func (vm *VM) Initialize(
 
 	vm.setAppRequestHandlers()
 
-	vm.StateSyncServer = NewStateSyncServer(&StateSyncServerConfig{
-		Chain:            vm.blockChain,
-		AtomicTrie:       vm.atomicTrie,
-		SyncableInterval: vm.config.StateSyncCommitInterval,
-	})
+	// TODO: register a different provider for the atomic txs
```
ceyonur, line 621:
```suggestion
```

plugin/evm/vm.go
```diff
@@ -726,9 +724,11 @@ func (vm *VM) initializeStateSyncClient(lastAcceptedHeight uint64) error {
 		}
 	}
 
-	vm.StateSyncClient = NewStateSyncClient(&StateSyncClientConfig{
-		Chain: vm.eth,
-		State: vm.State,
+	// register different extrasyncer and parser for atomic txs
```
ceyonur, line 727:
```suggestion
```

plugin/evm/message/codec.go
```diff
@@ -26,7 +30,7 @@ func init() {
 	c.SkipRegistrations(2)
 	errs.Add(
 		// Types for state sync frontier consensus
-		c.RegisterType(SyncSummary{}),
+		c.RegisterType(SyncSummaryType),
```
qdm12, line 33: Why not directly BlockSyncSummary{} though?

plugin/evm/message/syncable.go
```diff
@@ -14,78 +14,107 @@ import (
 	"github.com/ava-labs/avalanchego/snow/engine/snowman/block"
 )
 
-var _ block.StateSummary = &SyncSummary{}
+var _ Syncable = &BlockSyncSummary{}
```
qdm12, line 17:
```suggestion
var _ Syncable = (*BlockSyncSummary)(nil)
```

plugin/evm/message/syncable.go
```diff
@@ -14,78 +14,107 @@ import (
...
 
 	summaryID  ids.ID
 	bytes      []byte
-	acceptImpl func(SyncSummary) (block.StateSyncMode, error)
+	acceptImpl AcceptImplFn
+}
+
+type BlockSyncSummaryParser struct{}
+
+func NewBlockSyncSummaryParser() SyncableParser {
```
qdm12, line 46:
```suggestion
func NewBlockSyncSummaryParser() *BlockSyncSummaryParser {
```

plugin/evm/message/syncable.go
```diff
@@ -14,78 +14,107 @@ import (
...
+	return &BlockSyncSummaryParser{}
 }
 
-func NewSyncSummaryFromBytes(summaryBytes []byte, acceptImpl func(SyncSummary) (block.StateSyncMode, error)) (SyncSummary, error) {
-	summary := SyncSummary{}
+func (b *BlockSyncSummaryParser) ParseFromBytes(summaryBytes []byte, acceptImpl AcceptImplFn) (Syncable, error) {
+	summary := BlockSyncSummary{}
 	if codecVersion, err := Codec.Unmarshal(summaryBytes, &summary); err != nil {
-		return SyncSummary{}, err
+		return nil, err
```
qdm12, line 53: nit a bit out of scope, but whilst we're at changing this line:

```suggestion
		return nil, fmt.Errorf("decoding to block sync summary: %s", err)
```

plugin/evm/message/syncable.go
```diff
@@ -14,78 +14,107 @@ import (
...
 	} else if codecVersion != Version {
-		return SyncSummary{}, fmt.Errorf("failed to parse syncable summary due to unexpected codec version (%d != %d)", codecVersion, Version)
+		return nil, fmt.Errorf("failed to parse syncable summary due to unexpected codec version (%d != %d)", codecVersion, Version)
 	}
 
 	summary.bytes = summaryBytes
 	summaryID, err := ids.ToID(crypto.Keccak256(summaryBytes))
 	if err != nil {
-		return SyncSummary{}, err
+		return nil, err
```
qdm12, line 61: nit a bit out of scope, but whilst we're at changing this line:

```suggestion
		return nil, fmt.Errorf("converting summary bytes to id: %s", err)
```

plugin/evm/message/syncable.go
```diff
@@ -14,78 +14,107 @@ import (
...
 	}
 	summary.summaryID = summaryID
 	summary.acceptImpl = acceptImpl
-	return summary, nil
+	return &summary, nil
 }
 
-func NewSyncSummary(blockHash common.Hash, blockNumber uint64, blockRoot common.Hash, atomicRoot common.Hash) (SyncSummary, error) {
-	summary := SyncSummary{
+func NewBlockSyncSummary(blockHash common.Hash, blockNumber uint64, blockRoot common.Hash) (Syncable, error) {
```
qdm12, line 68:
```suggestion
func NewBlockSyncSummary(blockHash common.Hash, blockNumber uint64, blockRoot common.Hash) (*BlockSyncSummary, error) {
```

plugin/evm/message/syncable.go
```diff
@@ -14,78 +14,107 @@ import (
...
+	summary := BlockSyncSummary{
 		BlockNumber: blockNumber,
 		BlockHash:   blockHash,
 		BlockRoot:   blockRoot,
-		AtomicRoot:  atomicRoot,
 	}
 	bytes, err := Codec.Marshal(Version, &summary)
 	if err != nil {
-		return SyncSummary{}, err
+		return nil, err
```
qdm12, line 76: nit a bit out of scope, but whilst we're at changing this line:

```suggestion
		return nil, fmt.Errorf("encoding block sync summary: %s", err)
```

plugin/evm/message/syncable.go
```diff
@@ -14,78 +14,107 @@ import (
...
 	}
 
 	summary.bytes = bytes
 	summaryID, err := ids.ToID(crypto.Keccak256(bytes))
 	if err != nil {
-		return SyncSummary{}, err
+		return nil, err
```
qdm12, line 82: nit a bit out of scope, but whilst we're at changing this line:

```suggestion
		return nil, fmt.Errorf("converting summary bytes to id: %s", err)
```

plugin/evm/message/syncable.go
```diff
@@ -14,78 +14,107 @@ import (
...
 
-func (s SyncSummary) ID() ids.ID {
+func (s *BlockSyncSummary) ID() ids.ID {
 	return s.summaryID
 }
 
-func (s SyncSummary) String() string {
-	return fmt.Sprintf("SyncSummary(BlockHash=%s, BlockNumber=%d, BlockRoot=%s, AtomicRoot=%s)", s.BlockHash, s.BlockNumber, s.BlockRoot, s.AtomicRoot)
+func (s *BlockSyncSummary) String() string {
+	return fmt.Sprintf("SyncSummary(BlockHash=%s, BlockNumber=%d, BlockRoot=%s)", s.BlockHash, s.BlockNumber, s.BlockRoot)
```
qdm12, line 114:
```suggestion
	return fmt.Sprintf("BlockSyncSummary(BlockHash=%s, BlockNumber=%d, BlockRoot=%s)", s.BlockHash, s.BlockNumber, s.BlockRoot)
```

plugin/evm/vm.go
```diff
@@ -617,11 +618,7 @@ func (vm *VM) Initialize(
 
 	vm.setAppRequestHandlers()
 
-	vm.StateSyncServer = NewStateSyncServer(&StateSyncServerConfig{
-		Chain:            vm.blockChain,
-		AtomicTrie:       vm.atomicTrie,
-		SyncableInterval: vm.config.StateSyncCommitInterval,
-	})
+	vm.StateSyncServer = vmsync.NewStateSyncServer(vm.blockChain, atomic.NewAtomicProvider(vm.blockChain, vm.atomicTrie), vm.config.StateSyncCommitInterval)
```
qdm12, line 621: nit for tiny better readability:

```suggestion
	atomicProvider := atomic.NewAtomicProvider(vm.blockChain, vm.atomicTrie)
	vm.StateSyncServer = vmsync.NewStateSyncServer(vm.blockChain, atomicProvider, vm.config.StateSyncCommitInterval)
```

qdm12 (commented):
Nice tidying up 😉  And thanks for that thick detailed PR body 👍

I just left a few comments, mostly non-important things to be fair. Most of them you can apply as suggestions directly or discard them in case we should keep similar code with subnet-evm at this time.

plugin/evm/atomic/atomic_sync_extender.go
```diff
@@ -0,0 +1,66 @@
+// (c) 2021-2022, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+// TODO: move to separate package
+package atomic
+
+import (
+	"context"
+	"fmt"
+
+	syncclient "github.com/ava-labs/coreth/sync/client"
```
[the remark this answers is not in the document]
ceyonur: that pkg is named statesyncclient, and for some reason I cannot import it without aliasing.

plugin/evm/atomic/atomic_sync_extender.go
```diff
@@ -0,0 +1,66 @@
...
+		stateSyncRequestSize: stateSyncRequestSize,
+	}
+}
+
+func (a *atomicSyncExtender) Sync(ctx context.Context, client syncclient.Client, syncSummary message.Syncable) error {
+	atomicSyncSummary, ok := syncSummary.(*AtomicBlockSyncSummary)
+	if !ok {
+		return fmt.Errorf("expected AtomicBlockSyncSummary, got %T", syncSummary)
+	}
+	log.Info("atomic tx: sync starting", "root", atomicSyncSummary)
```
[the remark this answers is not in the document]
ceyonur: I think this is useful to track the sync progress for node operators (so should be info)

plugin/evm/atomic/atomic_sync_extender.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/ethereum/go-ethereum/log"
+)
+
+type atomicSyncExtender struct {
+	backend              AtomicBackend
+	stateSyncRequestSize uint16
+}
+
+func NewAtomicSyncExtender(backend AtomicBackend, stateSyncRequestSize uint16) sync.Extender {
+	return &atomicSyncExtender{
```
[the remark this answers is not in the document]
ceyonur: I'm not sure how much we follow this in other places, but makes sense.

plugin/evm/atomic/atomic_sync_extender.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (a *atomicSyncExtender) Sync(ctx context.Context, client syncclient.Client, syncSummary message.Syncable) error {
+	atomicSyncSummary, ok := syncSummary.(*AtomicBlockSyncSummary)
+	if !ok {
+		return fmt.Errorf("expected AtomicBlockSyncSummary, got %T", syncSummary)
+	}
+	log.Info("atomic tx: sync starting", "root", atomicSyncSummary)
+	atomicSyncer, err := a.backend.Syncer(client, atomicSyncSummary.AtomicRoot, atomicSyncSummary.BlockNumber, a.stateSyncRequestSize)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
ceyonur: but isn't that the case if we use `%s` in any of chained (wrapped) errs, then we would lose all the ability to check with `errors.Is`?

plugin/evm/atomic/atomic_sync_provider.go
```diff
@@ -0,0 +1,49 @@
...
+
+func NewAtomicProvider(chain *core.BlockChain, atomicTrie AtomicTrie) sync.SummaryProvider {
+	return &atomicSyncProvider{chain: chain, atomicTrie: atomicTrie}
+}
+
+// StateSummaryAtHeight returns the SyncSummary at [height] if valid and available.
+func (a *atomicSyncProvider) StateSummaryAtHeight(height uint64) (block.StateSummary, error) {
+	atomicRoot, err := a.atomicTrie.Root(height)
+	if err != nil {
+		return nil, fmt.Errorf("error getting atomic trie root for height (%d): %w", height, err)
```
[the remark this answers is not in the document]
ceyonur: I don't think there is any particular reason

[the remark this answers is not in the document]
ceyonur: I feel like "ing" sounds a bit towards system is progressing/working as expected. I agree that "error" can clutter but IMHO starting with "ing" looks a bit weird for an error at first sight.

plugin/evm/atomic/atomic_sync_provider.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	if (atomicRoot == common.Hash{}) {
+		return nil, fmt.Errorf("atomic trie root not found for height (%d)", height)
+	}
+
+	blk := a.chain.GetBlockByNumber(height)
```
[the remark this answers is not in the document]
ceyonur: block is an imported package name

plugin/evm/atomic/syncable.go
```diff
@@ -0,0 +1,116 @@
...
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/coreth/plugin/evm/message"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+
+	"github.com/ava-labs/avalanchego/snow/engine/snowman/block"
+)
+
+var _ message.Syncable = &AtomicBlockSyncSummary{}
```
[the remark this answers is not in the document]
ceyonur: we should probably do this for all other places (AFAIK we even had a linter in avalanchego to check this)

plugin/evm/atomic/syncable.go
```diff
@@ -0,0 +1,116 @@
...
+	message.SyncSummaryType = &AtomicBlockSyncSummary{}
+}
+
+type atomicSyncSummaryParser struct{}
+
+func NewAtomicSyncSummaryParser() message.SyncableParser {
+	return &atomicSyncSummaryParser{}
+}
+
+func (b *atomicSyncSummaryParser) ParseFromBytes(summaryBytes []byte, acceptImpl message.AcceptImplFn) (message.Syncable, error) {
```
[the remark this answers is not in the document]
ceyonur: This needs to be message.Syncable so that we can implement `message.SyncableParser`

plugin/evm/message/codec.go
ceyonur: BlockSyncSummary has changed (removed the AtomicRoot). This is a (probably janky) workaround in order to be compatible but let another pkg register it's own type.

darioush (changes requested):

plugin/evm/atomic/atomic_sync_provider.go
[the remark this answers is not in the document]
qdm12: Sure, whatever works then 😉 
Although, just to be clear, `[ERROR] ...: error getting ...: final error` versus `[ERROR] ...: getting ...: final error` looks the same to me, except error being repeated.

plugin/evm/atomic/sync/atomic_sync_extender.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+var _ sync.Extender = (*AtomicSyncExtender)(nil)
+
+type AtomicSyncExtender struct {
+	backend              state.AtomicBackend
+	stateSyncRequestSize uint16
+}
+
+func NewAtomicSyncExtender(backend state.AtomicBackend, stateSyncRequestSize uint16) *AtomicSyncExtender {
```
qdm12, line 25: nit you could define an interface `AtomicBackend` in this package instead of using `state.AtomicBackend` so that:

- reduces dependency on `state` package
- narrower interface: you only need the Root method here

plugin/evm/atomic/sync/atomic_sync_provider.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/ava-labs/coreth/core"
+	"github.com/ava-labs/coreth/plugin/evm/atomic/state"
+	"github.com/ava-labs/coreth/plugin/evm/sync"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+var _ sync.SummaryProvider = &AtomicSyncProvider{}
+
+type AtomicSyncProvider struct {
+	chain      *core.BlockChain
```
qdm12, line 19: you could define an interface `Chain` for this `chain` field with only `GetBlockByNumber` and `HasState`, so that it removes the dependency on the core package.

plugin/evm/atomic/sync/atomic_sync_extender.go
```diff
@@ -0,0 +1,79 @@
...
+	return &AtomicSyncExtender{
+		backend:              backend,
+		stateSyncRequestSize: stateSyncRequestSize,
+	}
+}
+
+func (a *AtomicSyncExtender) Sync(ctx context.Context, client syncclient.Client, verDB *versiondb.Database, syncSummary message.Syncable) error {
```
qdm12, line 32: `client syncclient.Client` can be the narrower interface `client syncclient.LeafClient` (`Extender` interface needs an update).

plugin/evm/atomic/sync/atomic_sync_extender.go
```diff
@@ -0,0 +1,79 @@
...
+	atomicSyncer, err := NewAtomicSyncer(
+		client,
+		verDB,
+		a.backend.AtomicTrie(),
+		atomicSyncSummary.AtomicRoot,
+		atomicSyncSummary.BlockNumber,
+		a.stateSyncRequestSize,
+	)
+	if err != nil {
+		return err
```
qdm12, line 47: nit 

```suggestion
		return fmt.Errorf("failed to create atomic syncer: %w", err)
```

plugin/evm/atomic/sync/atomic_sync_provider.go
```diff
@@ -0,0 +1,52 @@
...
+	atomicRoot, err := a.atomicTrie.Root(height)
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve atomic trie root for height (%d): %w", height, err)
+	}
+
+	if atomicRoot == (common.Hash{}) {
+		return nil, fmt.Errorf("atomic trie root not found for height (%d)", height)
+	}
+
+	blk := a.chain.GetBlockByNumber(height)
```
qdm12, line 38: nit perhaps (makes more sense to me I'd say)

1. alias `block` to `snowblock`
2. then rename `blk` to `block`

plugin/evm/message/syncable.go
```diff
@@ -14,78 +14,107 @@ import (
...
+type SyncableParser interface {
+	ParseFromBytes(summaryBytes []byte, acceptImpl AcceptImplFn) (Syncable, error)
+}
+
+type AcceptImplFn func(Syncable) (block.StateSyncMode, error)
+
+// BlockSyncSummary provides the information necessary to sync a node starting
 // at the given block.
-type SyncSummary struct {
+type BlockSyncSummary struct {
```
qdm12, line 34: nit did we need to rename `SyncSummary` to `BlockSyncSummary`? 🤔 
It seems like unneeded changes, unless there is a reason?

plugin/evm/atomic/atomic_sync_provider.go
```diff
@@ -0,0 +1,51 @@
+// (c) 2021-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+package atomic
+
+import (
+	"fmt"
+
+	"github.com/ava-labs/avalanchego/snow/engine/snowman/block"
+
+	"github.com/ava-labs/coreth/core"
```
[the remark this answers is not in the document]
qdm12: See https://github.com/ava-labs/coreth/pull/723#discussion_r1914909188

plugin/evm/atomic/sync/atomic_sync_provider.go
ceyonur: unfortunately it goes deeper than that as `GetBlockByNumber` returns `types.Block` so we end-up importing the core package anyway. We can create an adapter and wrap the passed blockchain here, but I think having a separate pkg for sync should be good enough.

plugin/evm/atomic/sync/atomic_sync_provider.go
ceyonur: let's keep this as it is.

plugin/evm/message/syncable.go
ceyonur: This is to distinguish between AtomicBlockSyncSummary

plugin/evm/atomic/sync/atomic_sync_extender.go
ceyonur: I moved them to interfaces pkg.

plugin/evm/atomic/sync/atomic_sync_extender.go
ceyonur: done

plugin/evm/atomic/atomic_sync_extender.go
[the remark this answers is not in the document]
darioush: We can agree to disagree on this one, unless you know the behavior of all callers + callees, wrapping an error could cause "spooky action at distance" between unrelated code that does check errors with `==` vs. `errors.Is` as part of an unrelated change. Life would be better if errors were properly handled as part of APIs, but we are dealing with go import packages, not modules that have defined their interfaces up to which errors are returned. Since we cannot control how other people code or re-write all code at once, "sprinkling in" error wraps where there was none seems to violate "if it is not broken don't change it", making refactors more risky than need be.

The reason some unwrapped errors are difficult to debug is because they are programming errors (and should panic imo), which would generate a stack trace and is easy to debug. Errors returned to users should always have the related information, unfortunately golang culture does not differentiate between programming errors and errors that are meaningful to report to users.

plugin/evm/atomic/sync/atomic_sync_provider.go
darioush: The separate package `sync` should be good enough, but just to clarify `types` doesn't import `core` as far as I know.
In libevm, we will use the upstream `type` and this will be less of a concern.

plugin/evm/atomic/sync/syncable.go
```diff
@@ -0,0 +1,120 @@
...
+	BlockRoot   common.Hash `serialize:"true"`
+	AtomicRoot  common.Hash `serialize:"true"`
+
+	summaryID  ids.ID
+	bytes      []byte
+	acceptImpl message.AcceptImplFn
+}
+
+func init() {
+	message.SyncSummaryType = &AtomicBlockSyncSummary{}
```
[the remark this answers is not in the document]
ceyonur: yea I removed that assumption and unexported codecs. VM should now create the codec itself with the proper type.

plugin/evm/atomic/state/interfaces/interfaces.go
```diff
@@ -0,0 +1,155 @@
+package interfaces
```
qdm12, line 1: Ideally have these in the package where they are needed. Creating an interfaces package is really strange imo.
Also, having them in a shared interfaces package means it will be harder to have narrow interfaces per package (i.e. package A might only need method X, package B needs methods X and Y, but if they both import interfaces.Interface they would both require X and Y - not great)
I usually have an `interfaces.go` file in the package where the interfaces are needed, since sometimes it's hard to know where to define an interface if it's used in different files in that package.

plugin/evm/atomic/atomic_sync_extender.go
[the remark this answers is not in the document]
qdm12: > "don't make unrelated changes while refactoring"

I **mostly** agree 😉 

> some caller may be using == or errors.Is (even if they arguably shouldn't)

Just to be clear:

- `return fmt.Errorf("wrapping: %w", err)` and `return err` both work if the caller checks with `errors.Is`
- only `return err` works if the caller checks with `err == someErr` (which is often the case in geth code)

> If anything this type of subtle behavior change is exactly the type of change to avoid in large refactors

It is time consuming to check what the caller(s) do with the error (aka does it check with `==`), and risky not to check it. This might change too in the future with a geth sync, so really this is very prone to error. I would suggest we fix this slowly but safely by:

1. Not wrapping errors at all from now on
1. (praying hard this doesn't get insta-closed) PR to geth to use `errors.Is`  and `errors.As` and turn on `errlint`, excluding the wrap-your-errors-with-fmt.Errorf requirement. So no more `==` error checks ever in geth.
1. Wait for sync with the *next future* geth release (I'll be patient!!)
1. Review our own code and add error wrappings where necessary

plugin/evm/atomic/state/interfaces/interfaces.go
darioush: I agree with this and would like to see less "shared interfaces" between packages, even though the existing code doesn't adhere to this.
This goes hand in hand with returning concrete types instead of interfaces, and paying more attention to which methods/fields are exported (as they are "the interface".

darioush (approved):

PR #731, opened 2025-01-03
chore(ci): define github labels as code

qdm12:
## Why this should be merged

Allow to track in source control the Github labels, and to easily migrate Github labels (i.e. renames).

## How this works

New Github workflow `.github/workflows/labels.yml` using `github/labels.yml` as a source for the labels. All the label data was extracted from [this API call](https://api.github.com/repos/ava-labs/coreth/labels?per_page=100) to keep it the exact same.

Removed useless labels according to Darioush 😉 

## How this was tested

Dry run of labeler job on this PR

## Need to be documented?

No

No


.github/labels.yml
```diff
@@ -0,0 +1,64 @@
...
+- name: "e upgrade"
+  color: "F5B85A"
+- name: "enhancement"
+  color: "a2eeef"
+  description: "New feature or request"
+- name: "good first issue"
+  color: "7057ff"
+  description: "Good for newcomers"
+- name: "gossiping upgrade"
+  color: "d93f0b"
```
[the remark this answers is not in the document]
qdm12: I have no idea 😆

.github/labels.yml
```diff
@@ -0,0 +1,64 @@
...
+- name: "Do not merge"
+  color: "b60205"
+- name: "documentation"
+  color: "0075ca"
+  description: "Improvements or additions to documentation"
+- name: "duplicate"
+  color: "cfd3d7"
+  description: "This issue or pull request already exists"
+- name: "e upgrade"
+  color: "F5B85A"
```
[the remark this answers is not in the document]
qdm12: I thought it was weird as well!

.github/labels.yml
```diff
@@ -0,0 +1,64 @@
...
+- name: "bug"
+  color: "d73a4a"
+  description: "Something isn't working"
+- name: "ci"
+  color: "15E277"
+- name: "cleanup"
+  color: "49C111"
+- name: "database upgrade"
+  color: "a4a3ed"
+  description: "Requires database upgrade"
```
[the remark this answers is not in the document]
qdm12: I have no idea 😆

qdm12, line 16:
```suggestion
```

.github/labels.yml
qdm12, line 29:
```suggestion
```

.github/labels.yml
qdm12, line 37:
```suggestion
```

.github/labels.yml
```diff
@@ -0,0 +1,64 @@
...
+- name: "gossiping upgrade"
+  color: "d93f0b"
+- name: "help wanted"
+  color: "008672"
+  description: "Extra attention is needed"
+- name: "incident response"
+  color: "DAD20E"
+- name: "integration-testing"
+  color: "0A0235"
+  description: "Refactoring of unit tests to fixture-based integration"
```
qdm12, line 45:
```suggestion
```

darioush (approved):

.github/labels.yml
```diff
@@ -0,0 +1,54 @@
+- name: "after coreth x subnet-evm"
+  color: "98E19D"
+  description: "To be done after merging the coreth and subnet-evm codebases"
+- name: "after libevm"
+  color: "98E19D"
+  description: "To be done only after using libevm for good"
```
meaghanfitzgerald, line 6: [no action required] these labels share the same color

meaghanfitzgerald (approved):

.github/labels.yml
qdm12: Indeed but:

1. already the case, we may change in a subsequent PR modifying this file
2. grouping similar labels (after xyz) with the same color is useful in my opinion 😉 Especially finding a distinguishable color for each label is rather difficult as soon as you have around 10 labels 😓

darioush (approved):

.github/workflows/labels.yml
```diff
@@ -0,0 +1,24 @@
+name: labels
+on:
+  push:
+    branches:
+      - main
```
qdm12, line 5:
```suggestion
      - master
```

darioush (approved):

.github/workflows/labels.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  labeler:
+    permissions:
+      contents: read
+      issues: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: crazy-max/ghaction-github-labeler@v5
```
ceyonur, line 22: I will play the devils advocate here. I'm not sure if using an external action justifies the advantage (at least in daily usage). Maybe I'm not seeing the exact advantage here, but it seems to me that this can just be a 5 min operation if we ever wanted to fully rename all labels (through https://github.com/ava-labs/coreth/labels). Not sure if it's worth to run a 3rd party action in every push/pull. even though the action can be something secure it still adds to the CI overhead.

qdm12: The aim is to have source-controlled, reviewable (through PRs) Github labels. The rename part is only useful once, for example to rename default label names.

The action also runs in its own workflow only when labels yml files are changed, so it won't pollute on every commit.

qdm12:
Closing this as not important.

PR #734, opened 2025-01-06
refactor: Split config package so it doesn't import core, core/vm

darioush:
## Why this should be merged
One main goal of the atomic package refactor is to have the client config imported / serialized without importing most of the VM functionality like core or core/vm.
As is on master, plugin/evm/config ends up importing stuff through legacypool & a direct import of "eth" (to return eth.Settings)

## How this works
The simplest approach seems to me as redeclaring the config type with its methods in plugin/evm.
I left "deprecate" to avoid moving the test but I can move the test if desired.

## How this was tested
CI. I also added a UT to enforce forbidden imports. I think this UT can live in plugin/evm since that's the "main" package.

## Need to be documented?
No

No


plugin/evm/config/config.go
```diff
@@ -224,65 +158,6 @@ type Config struct {
...
-	c.RPCTxFeeCap = defaultRpcTxFeeCap
-	c.MetricsExpensiveEnabled = defaultMetricsExpensiveEnabled
-
-	c.TxPoolPriceLimit = legacypool.DefaultConfig.PriceLimit
-	c.TxPoolPriceBump = legacypool.DefaultConfig.PriceBump
-	c.TxPoolAccountSlots = legacypool.DefaultConfig.AccountSlots
-	c.TxPoolGlobalSlots = legacypool.DefaultConfig.GlobalSlots
-	c.TxPoolAccountQueue = legacypool.DefaultConfig.AccountQueue
-	c.TxPoolGlobalQueue = legacypool.DefaultConfig.GlobalQueue
-	c.TxPoolLifetime.Duration = legacypool.DefaultConfig.Lifetime
```
qdm12, line 248: A perhaps cleaner (or not?) way to do this would be to pass in these fields in a struct you define in the plugin/evm/config package i.e. `config.LegacyPoolDefaults` which the caller could set with fields from the legacypool package.

plugin/evm/config/config.go
```diff
@@ -224,65 +158,6 @@ type Config struct {
...
 }
 
-// EthAPIs returns an array of strings representing the Eth APIs that should be enabled
-func (c Config) EthAPIs() []string {
-	return c.EnabledEthAPIs
-}
-
-func (c Config) EthBackendSettings() eth.Settings {
-	return eth.Settings{MaxBlocksPerRequest: c.MaxBlocksPerRequest}
-}
```
qdm12, line 234: For this, since it's used in only one place (unless it's meant to be exported as Go API), you could do the operation in plugin/evm by changing

```go
	vm.eth, err = eth.New(
		node,
		&vm.ethConfig,
		&EthPushGossiper{vm: vm},
		vm.chaindb,
		vm.config.EthBackendSettings(),
		lastAcceptedHash,
		dummy.NewFakerWithClock(callbacks, &vm.clock),
		&vm.clock,
	)
```

to

```go
	vm.eth, err = eth.New(
		node,
		&vm.ethConfig,
		&EthPushGossiper{vm: vm},
		vm.chaindb,
		eth.Settings{MaxBlocksPerRequest: vm.config.MaxBlocksPerRequest},
		lastAcceptedHash,
		dummy.NewFakerWithClock(callbacks, &vm.clock),
		&vm.clock,
	)
```

I think 🤔

qdm12 (commented):
Maybe there is an alternative approach to this (suggested in comments)?
Having a type alias with type assertion does work, but it's a tiny bit convoluted I think.

plugin/evm/imports_test.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+// getDependencies takes a fully qualified package name and returns a map of all
+// its recursive package imports (including itself) in the same format.
+func getDependencies(packageName string) (map[string]struct{}, error) {
+	// Configure the load mode to include dependencies
+	cfg := &packages.Config{Mode: packages.NeedDeps | packages.NeedImports | packages.NeedName | packages.NeedModule}
+	pkgs, err := packages.Load(cfg, packageName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load package %s: %v", packageName, err)
```
qdm12, line 21: nit doesn't `package.Load` include the package name in its error already 🤔  (in the best world, it should, maybe not the case haha)

```suggestion
		return nil, fmt.Errorf("failed to load package: %v", err)
```

plugin/evm/imports_test.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	// Start collecting dependencies
+	collectDeps(pkgs[0])
+	return deps, nil
+}
+
+func TestMustNotImport(t *testing.T) {
+	repo := "github.com/ava-labs/coreth"
+	withRepo := func(pkg string) string {
```
qdm12, line 47: nit use const whenever possible and inline variable where it's needed

```suggestion
	withRepo := func(pkg string) string {
		const repo = "github.com/ava-labs/coreth"
```

qdm12 (approved):

plugin/evm/imports_test.go
```diff
@@ -0,0 +1,66 @@
+// (c) 2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
```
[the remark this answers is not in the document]
darioush: Adding a comment. It seems the lack of such tests has caused inadvertent dependency cycles in the past, but if it seems extreme we can omit it.

qdm12 (approved):

plugin/evm/vm.go
```diff
@@ -660,7 +660,7 @@ func (vm *VM) initializeChain(lastAcceptedHash common.Hash) error {
 		&vm.ethConfig,
 		&EthPushGossiper{vm: vm},
 		vm.chaindb,
-		vm.config.EthBackendSettings(),
+		eth.Settings{MaxBlocksPerRequest: vm.config.MaxBlocksPerRequest},
```
ceyonur, line 663: why did we need to change this?

darioush: config package shouldn't import "eth" package (for eth.Settings)
overall config packages for use by clients should import as little as possible.

ceyonur: I wasn't even aware that we had such function in there (I thought this was eth.Settings already lol)

plugin/evm/config.go
```diff
@@ -0,0 +1,21 @@
...
+package evm
+
+import (
+	"github.com/ava-labs/coreth/core/txpool/legacypool"
+	"github.com/ava-labs/coreth/plugin/evm/config"
+)
+
+// defaultTxPoolConfig uses [legacypool.DefaultConfig] to make a [config.TxPoolConfig]
+// that can be passed to [config.Config.SetDefaults].
+var defaultTxPoolConfig = config.TxPoolConfig{
```
ceyonur, line 13: again I'm completely fine with directly copying default values from pool settings one by one if we don't want to define this and pass to config pkg.

ceyonur (approved):

plugin/evm/config.go
darioush: seems like a bit less duplication this way.

PR #737, opened 2025-01-07
Sync subnet-evm changes up to 981830ed

darioush:
## Why this should be merged
Takes recent development from subnet-evm

## How this works
Copies code

## How this was tested
CI

## Need to be documented?
N/A

N/A


scripts/build_docker_image.sh
```diff
@@ -2,7 +2,21 @@
 
 set -euo pipefail
 
-# Avalanche root directory
+# If set to non-empty, prompts the building of a multi-arch image when the image
```
darioush, line 5: not sure if this file is used at all?

ceyonur: No, I don't think it's used in Coreth as Coreth packed with AvalancheGo's docker.

ceyonur (approved):
LGTM

scripts/build_docker_image.sh
darioush: Let's remove it.

ceyonur (approved):
LEBTM (looking even better to me)

plugin/evm/vm_database.go
```diff
@@ -0,0 +1,82 @@
+// (c) 2019-2021, Ava Labs, Inc. All rights reserved.
```
qdm12, line 1: I suppose we want to keep it as a full copy-paste right?
Otherwise we might want to bump the end year:

```suggestion
// (c) 2019-2025, Ava Labs, Inc. All rights reserved.
```

qdm12 (approved):

plugin/evm/vm_database.go
darioush: Yes let's leave them the same in both codebases

PR #738, opened 2025-01-07
remove old rules

ceyonur:

### DIFF
--- a/consensus/dummy/consensus.go
+++ b/consensus/dummy/consensus.go
@@ -450,42 +450,39 @@ func (eng *DummyEngine) FinalizeAndAssemble(chain consensus.ChainHeaderReader, h
 			return nil, err
 		}
 	}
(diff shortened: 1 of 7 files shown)


consensus/dummy/consensus.go
```diff
@@ -450,42 +450,39 @@ func (eng *DummyEngine) FinalizeAndAssemble(chain consensus.ChainHeaderReader, h
...
-			header.BaseFee,
-			header.BlockGasCost,
-			txs,
-			receipts,
-			contribution,
-		); err != nil {
-			return nil, err
-		}
+
+	header.ExtDataGasUsed = extDataGasUsed
```
ceyonur, line 454: These can be remove because we don't expect block builders to assemble a block under old rules. However there are UTs that relies on older rules, so we cannot enforce new rules here without breaking UTs. Unfortunately most of these UTs are in blockchain_test (upstream)

PR #741, opened 2025-01-09
attempt to remove currentStateLock

darioush:


qdm12 (approved):
Maybe that's totally paranoid, but given the Init method on the LegacyPool does assign currentState, shall we add pool.mu.Lock() and unlock in that function, just in case ™️ Or not, if it's a copy pasta from geth 🤷

ceyonur (approved):

PR #743, opened 2025-01-09
 avoid import of params from plugin/evm/atomic 

darioush:
The params package will trigger one-time libevm type registrations so we should not import them in the client / shared types packages with avalanchego/wallets.

### DIFF
--- a/plugin/evm/atomic/params.go
+++ b/plugin/evm/atomic/params.go
@@ -1,0 +1,13 @@
+// (c) 2025 Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package atomic
+
+import "github.com/ava-labs/avalanchego/utils/units"
+
+const (
+	AvalancheAtomicTxFee = units.MilliAvax
+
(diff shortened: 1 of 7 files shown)


ceyonur (approved):

plugin/evm/atomic/params.go
```diff
@@ -0,0 +1,13 @@
...
+
+package atomic
+
+import "github.com/ava-labs/avalanchego/utils/units"
+
+const (
+	AvalancheAtomicTxFee = units.MilliAvax
+
+	// The base cost to charge per atomic transaction. Added in Apricot Phase 5.
+	AtomicTxBaseCost uint64 = 10_000
```
qdm12, line 12: nit perhaps a tiny tiny bit out of scope, but it would be nice to rename those whilst we're at it (same number of lines modified) to not have the atomic word in them given they're in the atomic package already. Something like

- `AvalancheAtomicTxFee` -> `AvalancheTxFee`
- `AtomicTxBaseCost` -> `TxBaseCost`

qdm12 (approved):
Nice 💯  Just a nit and feel free to discard it to address later in another PR another day 😄

PR #744, opened 2025-01-10
merge master to atomic base branch

ceyonur:
Merges master (at [b6b4dfb](https://github.com/ava-labs/coreth/commit/b6b4dfbc4bfc7f17e4cfe5f3c0fb44944176c884)) to atomic base branch


plugin/evm/atomic/state/atomic_backend.go
```diff
@@ -68,7 +75,7 @@ type AtomicBackend interface {
 }
 
 // atomicBackend implements the AtomicBackend interface using
-// the AtomicTrie, AtomicTxRepository, and the VM's shared memory.
+// the interfaces.AtomicTrie, AtomicTxRepository, and the VM's shared memory.
```
darioush, line 78: is this comment still correct?

plugin/evm/atomic/state/atomic_trie.go
```diff
@@ -140,7 +144,7 @@ func newAtomicTrie(
 	}
 	// initialize to EmptyRootHash if there is no committed root.
 	if root == (common.Hash{}) {
-		root = types.EmptyRootHash
+		root = EmptyAtomicRootHash
```
darioush: optional: we can import this from go-ethereum's upstream type if we don't want to duplicate this

plugin/evm/atomic/state/atomic_trie.go
```diff
@@ -16,30 +15,20 @@ import (
...
-
-	atomicTrieTipBufferSize = 1 // No need to support a buffer of previously accepted tries for the atomic trie
-	atomicTrieMemoryCap     = 64 * units.MiB
-)
-
-var (
-	_                            AtomicTrie = &atomicTrie{}
-	lastCommittedKey                        = []byte("atomicTrieLastCommittedBlock")
-	appliedSharedMemoryCursorKey            = []byte("atomicTrieLastAppliedToSharedMemory")
+	AtomicKeyLength = wrappers.LongLen + common.HashLength
```
darioush, line 34: nit: should we rename this to AtomicTrieKeyLength to be more descriptive?

darioush (approved):
we can also move the sync, state, txpool packages to "internal" if desired

plugin/evm/atomic/state/atomic_trie.go
ceyonur: I reverted it since it should be good after libevm transition

ceyonur:
> optional: we can also move the sync, state, txpool packages to "internal" if desired

We cannot move it in this PR as we use them in plugin/evm. We might be able to move them once we isolate the atomic pkg. (golang only allows internal pkgs to be used from their root)

PR #746, opened 2025-01-13
[wip] use header type from libevm

darioush:


core/types/header_ext.go
```diff
@@ -0,0 +1,197 @@
...
+
+	return nil
+}
+
+//go:generate go run github.com/fjl/gencodec -type HeaderSerializable -field-override headerMarshaling -out gen_header_json.go
+//go:generate go run github.com/ava-labs/libevm/rlp/rlpgen -type HeaderSerializable -out gen_header_rlp.go
+
+// HeaderSerializable defines the header of a block in the Ethereum blockchain,
+// as it is to be serialized into RLP and JSON.
+type HeaderSerializable struct {
```
qdm12, line 75: is it possible to unexport this type? 🤔

consensus/dummy/consensus.go
```diff
@@ -168,12 +168,13 @@ func (eng *DummyEngine) verifyHeaderGasFields(config *params.ChainConfig, header
 	}
 
 	// Verify BlockGasCost, ExtDataGasUsed not present before AP4
+	headerExtra := types.HeaderExtras(header)
```
qdm12, line 171: Can we rename `types.HeaderExtras` to `types.GetHeaderExtra`? I don't like Get prefixes, but I prefer it compared to having the plural of extras despite accessing a single extra (and can't name it `HeaderExtra` due to type name conflict)

core/types/header_ext.go
```diff
@@ -0,0 +1,197 @@
...
+		return err
+	}
+
+	updateToEth(in, eth)
+	updateToExtras(in, h)
+
+	return nil
+}
+
+//nolint:stdmethods
```
qdm12, line 47: I think this is ineffective, the warning comes from gopls AfAIK and no way to disable it unfortunately (again, AFAIK)

core/types/header_ext.go
```diff
@@ -0,0 +1,197 @@
...
+func (h *HeaderExtra) MarshalJSON(eth *ethtypes.Header) ([]byte, error) {
+	out := new(HeaderSerializable)
+
+	updateFromEth(out, eth)
+	updateFromExtras(out, h)
+
+	return out.MarshalJSON()
+}
+
+//nolint:stdmethods
```
qdm12, line 57: I think this is ineffective, the warning comes from gopls AfAIK and no way to disable it unfortunately (again, AFAIK)

core/types/header_ext.go
```diff
@@ -0,0 +1,197 @@
...
+	BlobGasUsed *uint64 `json:"blobGasUsed" rlp:"optional"`
+
+	// ExcessBlobGas was added by EIP-4844 and is ignored in legacy headers.
+	ExcessBlobGas *uint64 `json:"excessBlobGas" rlp:"optional"`
+
+	// ParentBeaconRoot was added by EIP-4788 and is ignored in legacy headers.
+	ParentBeaconRoot *common.Hash `json:"parentBeaconBlockRoot" rlp:"optional"`
+}
+
+// field type overrides for gencodec
```
qdm12, line 117: nit it would be great to expand on the comment since this has a sort of special role and may be thought of as unneeded by a developer

core/types/header_ext.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// Hash returns the block hash of the header, which is simply the keccak256 hash of its
+// RLP encoding.
+func (h *HeaderSerializable) Hash() common.Hash {
+	return rlpHash(h)
+}
+
+// updateFromEth updates the HeaderSerializable from the ethtypes.Header
+func updateFromEth(h *HeaderSerializable, eth *ethtypes.Header) {
```
qdm12, line 140: this and the next few methods should be methods on the headerSerializable struct, so they don't name-conflict with future other types such as blockSerializable

core/types/imports.go
```diff
@@ -23,11 +23,14 @@ type (
...
-	TxByNonce    = ethtypes.TxByNonce
-	TxData       = ethtypes.TxData
+	StateAccount      = ethtypes.StateAccount
+	SlimAccount       = ethtypes.SlimAccount
+	Header            = ethtypes.Header
+	BlockNonce        = ethtypes.BlockNonce
+	Transaction       = ethtypes.Transaction
+	Transactions      = ethtypes.Transactions
+	TxByNonce         = ethtypes.TxByNonce
+	TxData            = ethtypes.TxData
```
qdm12, line 33: We might want to sort these (and existing ones too) alphabetically to have a consistent order with subnet-evm as well.

core/types/imports.go
```diff
@@ -46,12 +49,17 @@ const (
...
+	BytesToBloom         = ethtypes.BytesToBloom
+	CreateBloom          = ethtypes.CreateBloom
+	NewReceipt           = ethtypes.NewReceipt
+	NewContractCreation  = ethtypes.NewContractCreation
+	NewTransaction       = ethtypes.NewTransaction
+	EncodeNonce          = ethtypes.EncodeNonce
+	NewEmptyStateAccount = ethtypes.NewEmptyStateAccount
+	SlimAccountRLP       = ethtypes.SlimAccountRLP
+	FullAccount          = ethtypes.FullAccount
+	FullAccountRLP       = ethtypes.FullAccountRLP
```
qdm12, line 62: We might want to sort these (and existing ones too) alphabetically to have a consistent order with subnet-evm as well.

core/types/libevm.go
```diff
@@ -0,0 +1,28 @@
...
+
+func IsMultiCoin(s ethtypes.StateOrSlimAccount) bool {
+	return bool(extras.StateAccount.Get(s))
+}
+
+func HeaderExtras(h *Header) *HeaderExtra {
+	return extras.Header.Get(h)
+}
+
+func WithHeaderExtras(h *Header, extra *HeaderExtra) *Header {
```
qdm12, line 25: rename to `WithHeaderExtra`?

core/types/header_ext.go
```diff
@@ -0,0 +1,197 @@
...
+
+	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/libevm/common/hexutil"
+	ethtypes "github.com/ava-labs/libevm/core/types"
+	"github.com/ava-labs/libevm/rlp"
+)
+
+// HeaderExtra is a struct that contains extra fields used by Avalanche
+// in the block header.
+// This type uses HeaderSerializable to encode and decode the extra fields
```
qdm12, line 18: nit for godoc automated links

```suggestion
// This type uses [HeaderSerializable] to encode and decode the extra fields
```

core/types/block.go
```diff
@@ -249,6 +129,10 @@ func NewBlock(
 // CopyHeader creates a deep copy of a block header.
 func CopyHeader(h *Header) *Header {
 	cpy := *h
+	extras.Header.Set(&cpy, &HeaderExtra{})
+	cpyExtra := HeaderExtras(&cpy)
+	*cpyExtra = *HeaderExtras(h)
```
qdm12, line 134: nit a bit less convoluted I think:

```suggestion
	extra := HeaderExtras(h)
	cpyExtra := &HeaderExtra{
		ExtDataHash: extra.ExtDataHash,
	}
	cpy = *WithHeaderExtras(&cpy, cpyExtra)
```

You could also define `hExtra` instead of `extra` here to save one line below

accounts/abi/bind/bind_test.go
```diff
@@ -2179,6 +2179,11 @@ func golangBindings(t *testing.T, overload bool) {
 	if out, err := replacer.CombinedOutput(); err != nil {
 		t.Fatalf("failed to replace binding test dependency to current source tree: %v\n%s", err, out)
 	}
+	replacer = exec.Command(gocmd, "mod", "edit", "-x", "-require", "github.com/ava-labs/libevm@v0.0.0", "-replace", "github.com/ava-labs/libevm=github.com/ava-labs/libevm@v0.0.0-20250113230013-d10c4a3d1ff2")
+	replacer.Dir = pkg
+	if out, err := replacer.CombinedOutput(); err != nil {
+		t.Fatalf("failed to replace binding test dependency to current source tree: %v\n%s", err, out)
+	}
```
qdm12, line 2186: A comment would be nice 😄 
Maybe we can leave this commented with a comment explaining what it does for future work, before this gets merged / new libevm release.

core/types/header_ext.go
qdm12: No, rlpgen requires exported fields and is coming from geth.

core/types/header_ext.go
qdm12: this is a geth comment, leave it as is.

core/types/header_ext.go
```diff
@@ -5,32 +5,201 @@ package types
...
+	"github.com/ava-labs/libevm/common/hexutil"
 	ethtypes "github.com/ava-labs/libevm/core/types"
 	"github.com/ava-labs/libevm/rlp"
 )
 
+func GetHeaderExtra(h *Header) *HeaderExtra {
+	return extras.Header.Get(h)
+}
+
+func WithHeaderExtra(h *Header, extra *HeaderExtra) *Header {
```
qdm12, line 20: These needs a comment + tiny unit test

qdm12 (commented):
Self review

core/types/block_ext_test.go
```diff
@@ -0,0 +1,188 @@
...
+	"unicode"
+	"unsafe"
+
+	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/libevm/libevm/pseudo"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCopyHeader(t *testing.T) {
```
qdm12, line 19: Even when using libevm's `CopyHeader` next (#759), this test should be kept to ensure our PostCopy hook is coded correctly. `CopyHeader` can have subtle yet important changes and this test would help debunk some imo.

core/types/block.go
```diff
@@ -258,16 +144,20 @@ func CopyHeader(h *Header) *Header {
...
+		cpyExtra.BlockGasCost = new(big.Int).Set(hExtra.BlockGasCost)
 	}
 	if len(h.Extra) > 0 {
 		cpy.Extra = make([]byte, len(h.Extra))
 		copy(cpy.Extra, h.Extra)
 	}
+	if h.WithdrawalsHash != nil {
+		cpy.WithdrawalsHash = new(common.Hash)
+		*cpy.WithdrawalsHash = *h.WithdrawalsHash
+	}
```
qdm12, line 160: This is added for now to have the TestCopyHeader pass, but will be removed shortly with #759

nativeasset/contract.go
```diff
@@ -141,7 +141,7 @@ func (c *NativeAssetCall) Run(accessibleState contract.AccessibleState, caller c
 	stateDB.SubBalanceMultiCoin(caller, assetID, assetAmount)
 	stateDB.AddBalanceMultiCoin(to, assetID, assetAmount)
 
-	ret, remainingGas, err = accessibleState.Call(to, callData, remainingGas, new(uint256.Int), vm.WithUNSAFECallerAddressProxying())
+	ret, err = accessibleState.Call(to, callData, remainingGas, new(uint256.Int), vm.WithUNSAFECallerAddressProxying())
```
ARR4N, line 144: Why is this change in this PR?

params/config_extra.go
```diff
@@ -59,10 +59,10 @@ func SetEthUpgrades(c *ChainConfig) {
 }
 
 func GetExtra(c *ChainConfig) *extras.ChainConfig {
-	ex := payloads.FromChainConfig(c)
+	ex := payloads.ChainConfig.Get(c)
```
ARR4N, line 62: Didn't I review this in another PR? This suggests (along with the `nativeasset` change) that something's wrong in the commit history here.

precompile/contract/interfaces.go
```diff
@@ -61,7 +61,7 @@ type AccessibleState interface {
 	GetBlockContext() BlockContext
 	GetSnowContext() *snow.Context
 	GetChainConfig() precompileconfig.ChainConfig
-	Call(addr common.Address, input []byte, gas uint64, value *uint256.Int, opts ...vm.CallOption) (ret []byte, gasRemaining uint64, _ error)
+	Call(addr common.Address, input []byte, gas uint64, value *uint256.Int, opts ...vm.CallOption) (ret []byte, _ error)
```
ARR4N, line 64: As above re commit history. Feel free to just resolve this without comment, but I think it's worth noting the points at which this is happening.

scripts/eth-allowed-packages.txt
```diff
@@ -25,6 +25,8 @@
 "github.com/ava-labs/libevm/ethdb/pebble"
 "github.com/ava-labs/libevm/event"
 "github.com/ava-labs/libevm/libevm"
+"github.com/ava-labs/libevm/libevm/legacy"
+"github.com/ava-labs/libevm/libevm/pseudo"
```
ARR4N, line 29: Why is there a need for `pseudo` access? This is concerning and suggests I did something wrong in `libevm`.

scripts/tests.e2e.sh
```diff
@@ -45,6 +45,7 @@ git checkout -B "test-${AVALANCHE_VERSION}" "${AVALANCHE_VERSION}"
 
 echo "updating coreth dependency to point to ${CORETH_PATH}"
 go mod edit -replace "github.com/ava-labs/coreth=${CORETH_PATH}"
+go mod edit -replace "github.com/ava-labs/libevm=github.com/ava-labs/libevm@v0.0.0-20250214081829-bd9b506d6610"
```
ARR4N, line 48: Why not `0.2.0.rc.3`?

ARR4N (commented):
Partial review while I now go through `core/types/*_ext*.go`. Please don't force push as it will clear the history of what I've viewed and won't allow use of "commits since last review".

scripts/eth-allowed-packages.txt
qdm12: It's used in a `TestCopyHeader` test with reflection, I removed the import by replacing it with a reflect type string check. I may have been a bit far on the testing of copy header but it did help me debunk some problems 1-2 weeks ago.

core/types/block.go
```diff
@@ -249,6 +129,12 @@ func NewBlock(
 // CopyHeader creates a deep copy of a block header.
 func CopyHeader(h *Header) *Header {
```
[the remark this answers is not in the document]
qdm12: It's the next PR 😉

core/types/block_ext.go
```diff
@@ -28,7 +28,7 @@ func (b *Block) setExtData(data []byte, recalc bool) {
 	b.extdata = &_data
 	copy(*b.extdata, data)
 	if recalc {
-		b.header.ExtDataHash = CalcExtDataHash(*b.extdata)
+		GetHeaderExtra(b.header).ExtDataHash = CalcExtDataHash(*b.extdata)
```
ceyonur, line 31: nit: can we use a variable for extraHeader and then assign extDataHash?

ceyonur (approved):
lgtm, just a nit

scripts/eth-allowed-packages.txt
ARR4N: I think that's perfectly valid and it's absolutely cleaner to do that instead of using a reflect string. In case you're unaware, if a type implements https://pkg.go.dev/github.com/ava-labs/libevm/libevm/pseudo#EqualityChecker then you can use `pseudo.Type.Equal()` and if you're using `cmp.Diff()` then you can use:

```go
opt := cmp.Comparer((*pseudo.Type).Equal)
if diff := cmp.Diff(want, got, opt); diff != "" {
```

If you're testing, for example `HeaderExtra` then you can even implement `func (*HeaderExtra) Equal(*HeaderExtra) bool` in the _test_ file so it doesn't pollute the `types` package.

I'll implement `package pseudotest` to make this easier and so we can swap it out in the list of allowed packages.

core/types/header_ext_test.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+	assert.Equal(t, wantEth, decodedEth)
+
+	wantExtra := &HeaderExtra{
+		ExtDataHash: common.Hash{2},
+	}
+	assert.Equal(t, wantExtra, decodedExtra)
+}
+
+func TestHeaderSerializable_updates(t *testing.T) {
```
ARR4N, line 107: (subjective) I'm not a big fan of tests that target implementation details instead of the end goal, which is RLP and JSON round-tripping. I think these should be collapsed into one of the above tests, but that's just my opinion.

Rationale: if the implementation changes (which I strongly believe it should, in a move to `rlp.Fields` like with `Body` instead of a complete override) then this test becomes obsolete and the change is more dangerous. If the test is limited to the actual code path of interest then the test can remain unmodified and therefore provide far greater confidence. Effectively, refocusing "test behaviour, not implementation".

core/types/header_ext_test.go
```diff
@@ -0,0 +1,244 @@
...
+	t.Parallel()
+
+	t.Run("from", func(t *testing.T) {
+		t.Parallel()
+
+		eth := &ethtypes.Header{
```
ARR4N, line 113:
```suggestion
		header := &ethtypes.Header{
```

core/types/header_ext_test.go
```diff
@@ -0,0 +1,244 @@
...
+			BlockGasCost:   big.NewInt(23),
+		}
+
+		got := new(HeaderSerializable)
+		got.updateFromEth(eth)
+		got.updateFromExtras(extras)
+
+		allFieldsAreSet(t, got)
+
+		want := &HeaderSerializable{
```
ARR4N, line 147: This approach is error-prone in that it requires a reader to very closely inspect every single field and value (it's nice that the values increment, but the field order is still problematic). If you test via RLP (and/or JSON) round-tripping then you can simply assert that you get the exact input-struct values back. It will also cover both "to" and "from" sub tests.

core/types/block_ext_test.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+
+		headerExtra = &HeaderExtra{}
+		extras.Header.Set(want, headerExtra)
+
+		assert.Equal(t, want, cpy)
+	})
+
+	t.Run("filled_header", func(t *testing.T) {
+		h := &Header{
```
ARR4N, line 39:
```suggestion
		input := &Header{
```

See later suggestion for the rationale.

core/types/block_ext_test.go
```diff
@@ -0,0 +1,190 @@
...
+			BlockGasCost:   big.NewInt(23),
+		}
+		extras.Header.Set(h, headerExtra)
+
+		allFieldsAreSet(t, h)
+		allFieldsAreSet(t, headerExtra)
+
+		cpy := CopyHeader(h)
+
+		want := &Header{
```
ARR4N, line 73:
```suggestion
		got := CopyHeader(h)
		want := input
```

Now there's no way to make a mistake in the construction of `want`.

core/types/block_ext_test.go
```diff
@@ -0,0 +1,190 @@
...
+		fieldsAreDeepCopied(t, h, cpy)
+		fieldsAreDeepCopied(t, GetHeaderExtra(h), GetHeaderExtra(cpy))
+	})
+}
+
+func ptrTo[T any](x T) *T { return &x }
+
+func allFieldsAreSet(t *testing.T, x any) {
+	t.Helper()
+	require.Equal(t, reflect.Ptr.String(), reflect.TypeOf(x).Kind().String(), "x must be a pointer")
```
ARR4N, line 114:
```suggestion
	require.Equal(t, reflect.Pointer, reflect.TypeOf(x).Kind(), "x must be a pointer")
```

core/types/block_ext_test.go
```diff
@@ -0,0 +1,190 @@
...
+	v := reflect.ValueOf(x).Elem()
+	typ := v.Type()
+	require.Equal(t, reflect.Struct.String(), typ.Kind().String())
```
ARR4N, line 117:
```suggestion
	require.Equal(t, reflect.Struct, typ.Kind())
```

core/types/block_ext_test.go
```diff
@@ -0,0 +1,190 @@
...
+	for i := 0; i < v.NumField(); i++ {
+		field := v.Field(i)
+		fieldName := typ.Field(i).Name
+		fieldValue := field
+		if unicode.IsLower(rune(fieldName[0])) { // unexported
+			if field.Kind() == reflect.Ptr {
+				require.Falsef(t, field.IsNil(), "field %q is nil", fieldName)
+			}
+			field = reflect.NewAt(field.Type(), unsafe.Pointer(field.UnsafeAddr())).Elem() //nolint:gosec
```
ARR4N, line 126: This trick is too obscure to be used without commentary (and IMO to be used at all outside of tests and even then only when absolutely necessary).

[In other words](https://tenor.com/en-GB/view/april-ludgate-thats-gross-i-love-it-parks-and-recreation-gif-11076075)

core/types/block_ext_test.go
```diff
@@ -0,0 +1,190 @@
...
+			fieldValue = field
+		}
+		if field.Kind() == reflect.Pointer {
+			require.NotNilf(t, field.Interface(), "field %q is nil", fieldName)
```
ARR4N, line 130:
```suggestion
			require.Falsef(t, field.IsNil(), "field %q is nil", fieldName)
```

core/types/block_ext_test.go
```diff
@@ -0,0 +1,190 @@
...
+			fieldValue = field.Elem()
+		}
+		isSet := fieldValue.IsValid() && !fieldValue.IsZero()
```
ARR4N, line 133: Why is the call to `IsValid()` necessary?

core/types/block_ext_test.go
```diff
@@ -0,0 +1,190 @@
...
+	t.Helper()
+	require.Equal(t, reflect.Ptr.String(), reflect.TypeOf(x).Kind().String(), "x must be a pointer")
+	v := reflect.ValueOf(x).Elem()
+	typ := v.Type()
+	require.Equal(t, reflect.Struct.String(), typ.Kind().String())
+	for i := 0; i < v.NumField(); i++ {
+		field := v.Field(i)
+		fieldName := typ.Field(i).Name
+		fieldValue := field
+		if unicode.IsLower(rune(fieldName[0])) { // unexported
```
ARR4N, line 122:
```suggestion
		if !typ.Field(i).IsExported() {
```

core/types/block_ext_test.go
```diff
@@ -0,0 +1,190 @@
...
+			fieldValue = field.Elem()
+		}
+		isSet := fieldValue.IsValid() && !fieldValue.IsZero()
+		require.True(t, isSet, "field %q is not set", fieldName)
+	}
+}
+
+func fieldsAreDeepCopied(t *testing.T, original, cpy any) {
+	t.Helper()
+	require.Equal(t, reflect.Ptr.String(), reflect.TypeOf(original).Kind().String(), "original must be a pointer")
```
ARR4N, line 140: No need for calling `String()` on `Kind`s.

core/types/block_ext_test.go
```diff
@@ -0,0 +1,190 @@
...
+	require.Equal(t, reflect.Ptr.String(), reflect.TypeOf(cpy).Kind().String(), "cpy must be a pointer")
+
+	v := reflect.ValueOf(original).Elem()
+	for i := 0; i < v.NumField(); i++ {
+		field := v.Field(i)
+		fieldName := v.Type().Field(i).Name
+		isUnexported := unicode.IsLower(rune(fieldName[0]))
+		if isUnexported {
```
ARR4N, line 148:
```suggestion
		if !v.Type().Field(i).IsExported() {
```

core/types/block_ext_test.go
```diff
@@ -0,0 +1,190 @@
...
+			field = reflect.NewAt(field.Type(), unsafe.Pointer(field.UnsafeAddr())).Elem() //nolint:gosec
+		}
+		var originalField any
+		switch field.Kind() {
```
ARR4N, line 152: You don't need to mutate values to show that they're deep copied. For example, you can directly compare pointers to show that they address different parts of memory.

ARR4N:
The tests were bugging me so I [refactored](https://github.com/ava-labs/coreth/commit/ff71176ab71a612993f62b41142472bd4d1933f5) them into something much similar. I'm not opening a PR because this branch gets rebased a lot, but feel free to cherry-pick if you like it.

core/types/block_ext.go
```diff
@@ -28,7 +28,8 @@ func (b *Block) setExtData(data []byte, recalc bool) {
 	b.extdata = &_data
 	copy(*b.extdata, data)
 	if recalc {
-		b.header.ExtDataHash = CalcExtDataHash(*b.extdata)
+		ex := GetHeaderExtra(b.header)
+		ex.ExtDataHash = CalcExtDataHash(*b.extdata)
```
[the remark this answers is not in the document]
qdm12: That was a suggestion from https://github.com/ava-labs/coreth/pull/746#discussion_r1965676601 😄

plugin/evm/header/base_fee_test.go
```diff
@@ -24,6 +24,7 @@ func TestBaseFee(t *testing.T) {
 		name      string
 		upgrades  extras.NetworkUpgrades
 		parent    *types.Header
+		parentExt *types.HeaderExtra
```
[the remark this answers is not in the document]
qdm12: I was trying to lower the diffs with indentation 😄 
But sure changing it to `parentExtra`

plugin/evm/header/dynamic_fee_windower.go
```diff
@@ -199,6 +200,7 @@ func feeWindow(
 		// still calculated using the AP4 step. This is different than the
 		// actual BlockGasCost calculation used for the child block. This
 		// behavior is kept to preserve the original behavior of this function.
+		parent := types.GetHeaderExtra(parent)
```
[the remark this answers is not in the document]
qdm12: Indeed, changed to `parentExtra`. I was focusing on lower git diffs, but that seems a bit silly now.

plugin/evm/header/extra_test.go
```diff
@@ -230,6 +241,9 @@ func TestExtraPrefix(t *testing.T) {
 			config := &extras.ChainConfig{
 				NetworkUpgrades: test.upgrades,
 			}
+			if test.parentExt != nil {
```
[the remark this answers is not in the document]
qdm12: `types.HeaderWithExtra` is replaced by `types.SetHeaderExtra`, I think that's fine as it is? An alternative would be to just return the header from `types.SetHeaderExtra` if the extra is nil, what do you think?

core/types/header_ext_test.go
```diff
@@ -0,0 +1,244 @@
...
+		ParentHash: common.Hash{1},
+		// Required json fields
+		Difficulty: big.NewInt(2),
+		Number:     big.NewInt(3),
+	}
+	extra := &HeaderExtra{
+		ExtDataHash: common.Hash{2},
+	}
+
+	encoded, err := extra.EncodeJSON(eth)
```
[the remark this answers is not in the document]
qdm12: Totally, this and RLP test are both changed to set the extra in the header and encode/decode it, there is no point encoding/decoding each separately I think.

core/types/block_ext_test.go
qdm12: Call me paranoid, but - especially in a copy function - the point is to not rely on the input for the wanted result in case the input gets mutated.

core/types/block_ext_test.go
qdm12: Actually that's on purpose, since otherwise it just prints a confusing integer (enum). It's much better to print out the string to know what's wrong.

core/types/block_ext_test.go
qdm12: Actually that's on purpose, since otherwise it just prints a confusing integer (enum). It's much better to print out the string to know what's wrong.

core/types/block_ext_test.go
qdm12: The point is to print out the kind as a string instead of an enum, a lot more useful 😉

core/types/block_ext_test.go
qdm12: `isUnexported` is used at the end of the function as well, but I changed it to use

```go
isUnexported := !v.Type().Field(i).IsExported()
```

core/types/header_ext_test.go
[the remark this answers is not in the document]
qdm12: Good point, thanks for adding this in your commit! 💯

core/types/block_ext_test.go
qdm12: Indeed, done in your modified commit 💯

core/types/block_ext_test.go
qdm12: Yeah kind of ugly way to check unexported fields. The latest code using your commit doesn't check unexported fields anymore, but I suppose that's fine for the time being. Should we however worry about future addition of unexported fields and us forgetting to deep copy them?

core/types/block_ext_test.go
qdm12: No longer used with newer code, resolving this.

core/types/block_ext_test.go
qdm12: No longer used with newer code, resolving this.

core/types/block_ext_test.go
qdm12: No longer used with newer code, resolving this.

core/types/header_ext_test.go
qdm12: Good point, done in last commit 👍

core/types/header_ext_test.go
qdm12: No longer used with newer code, resolving this.

core/types/header_ext_test.go
```diff
@@ -0,0 +1,144 @@
...
+	for i := range v.Type().NumField() {
+		field := v.Type().Field(i)
+		if !field.IsExported() {
+			continue
+		}
+
+		t.Run(field.Name, func(t *testing.T) {
+			switch f := v.Field(i).Interface().(type) {
+			case common.Hash, common.Address, BlockNonce, Bloom, uint64:
+				assert.NotZero(t, f)
```
ARR4N, line 128: (subjective rant) I very strongly dislike these "fancier" assert/require functions (including `IsType`) when there's a native language construct for achieving the same in an obviously correct manner. I had to go review the implementation of `NotZero()` to convince myself that it achieves exactly what we want.

I'm not entirely convinced that it's appropriate for `big.Int`.

ARR4N (changes requested):
The change to the test makes it possibly tautological so it's important to revert it.

core/types/block_ext_test.go
ARR4N: > Call me paranoid, but - especially in a copy function - the point is to not rely on the input for the wanted result in case the input gets mutated.

Absolutely agree. That's why I implemented it as a function that returns a new instance of the same data.

core/types/block_ext_test.go
ARR4N: https://google.github.io/styleguide/go/decisions#assert `*cough*` 😛

core/types/block_ext_test.go
ARR4N: TL;DR no

If something unexported is added by geth then it won't have a direct effect on RLP nor JSON so, from the perspective of a libevm consumer, it's not something that needs to be worried about. `CopyHeader` will use the upstream code anyway, so only the `HeaderExtra` copy needs to be tested.

core/types/block_ext_test.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+// assertDifferentPointers asserts that `a` and `b` are both non-nil
+// pointers of the same type and point to different memory locations.
+func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
+	t.Helper()
+	require.NotNilf(t, a, "a (%T) cannot be nil", a)
+	require.NotNilf(t, b, "b (%T) cannot be nil", b)
+	require.IsType(t, a, b, "a and b must be of the same type")
```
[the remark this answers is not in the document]
qdm12: Added prefix `test is badly written: `

core/types/header_ext_test.go
```diff
@@ -0,0 +1,144 @@
...
+	require.NoError(t, err, "decode")
+
+	want := headerWithNonZeroFields()
+	want.WithdrawalsHash = nil
+	assert.Equal(t, want, got)
+
+	return b
+}
+
+func Test_headerWithNonZeroFields(t *testing.T) {
```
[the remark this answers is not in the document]
qdm12: Actually, that's the only point in using a underscore: to specify case sensitive function names for a test, given you cannot use `TestheaderWithNonZeroFields`. Ok in this case it's not that useful, but, if you have MyFunc and myFunc, and a test for each, how would you differentiate them? The best I came up with is `TestMyFunc` and `Test_myFunc` 🤔 WDYT?

core/types/header_ext_test.go
qdm12: I get your points, but:

- one case was already using `assert.NotEmpty` it fell natural to use other assert.* functions
- assert.NotZero using `any` (not great) allows to combine switch cases together (great)
- your function `assertNonZero` felt like something that can be replaced with a lib-already-in-use-function

Regarding *big.Int, your line was `assert.NotEqual(t, 0, f.Cmp(big.NewInt(0)))` but what we really want is just for it to be non-nil, leaving it to `0` is fine in that case (not the zero value despite being `0` 😄)

For the require.IsType, yes it's not amazing (maybe one day we'll get generics with this lib, probably not 😄)  but it's an explicit one-liner, better than a switch case with generic 😉

core/state_processor_test.go
```diff
@@ -383,8 +383,9 @@ func GenerateBadBlock(parent *types.Block, engine consensus.Engine, txs types.Tr
 		BaseFee:   baseFee,
 	}
 	if configExtra.IsApricotPhase4(header.Time) {
-		header.BlockGasCost = big.NewInt(0)
-		header.ExtDataGasUsed = big.NewInt(0)
+		ex := types.GetHeaderExtra(header)
```
darioush, line 386: nit: not that I'm against this `ex` name, but not sure how I feel about it. I'd say let's go with `headerExtra` or `extra` for consistency

core/types/block_ext.go
darioush: was this done?

core/types/block_ext_test.go
darioush: agree about keeping the test, but we may need to find a better home for it soon.

core/types/header_ext.go
```diff
@@ -5,32 +5,210 @@ package types
...
+	h.BlockGasCost = extras.BlockGasCost
+}
+
+func (h *HeaderSerializable) updateToExtras(extras *HeaderExtra) {
+	extras.ExtDataHash = h.ExtDataHash
+	extras.ExtDataGasUsed = h.ExtDataGasUsed
+	extras.BlockGasCost = h.BlockGasCost
+}
+
+//go:generate go run github.com/fjl/gencodec -type HeaderSerializable -field-override headerMarshaling -out gen_header_json.go
```
darioush, line 146: I wonder if we should call the output file something else? like gen_header_serializable_{json/rlp}.go ?

core/types/gen_header_json.go
```diff
@@ -67,86 +68,86 @@ func (h Header) MarshalJSON() ([]byte, error) {
...
-		BaseFee          *hexutil.Big    `json:"baseFeePerGas" rlp:"optional"`
-		ExtDataGasUsed   *hexutil.Big    `json:"extDataGasUsed" rlp:"optional"`
-		BlockGasCost     *hexutil.Big    `json:"blockGasCost" rlp:"optional"`
-		BlobGasUsed      *hexutil.Uint64 `json:"blobGasUsed" rlp:"optional"`
-		ExcessBlobGas    *hexutil.Uint64 `json:"excessBlobGas" rlp:"optional"`
-		ParentBeaconRoot *common.Hash    `json:"parentBeaconBlockRoot" rlp:"optional"`
-	}
-	var dec Header
+// DecodeJSON unmarshals from JSON.
+func (h *HeaderSerializable) DecodeJSON(input []byte) error {
```
darioush, line 72: did we change the JSON gen in libevm to output methods called Encode/DecodeJSON?

core/types/header_ext.go
```diff
@@ -5,32 +5,210 @@ package types
...
+	ExtDataGasUsed *hexutil.Big
+	BlockGasCost   *hexutil.Big
+	Hash           common.Hash `json:"hash"` // adds call to Hash() in MarshalJSON
+	BlobGasUsed    *hexutil.Uint64
+	ExcessBlobGas  *hexutil.Uint64
+}
+
+// Hash returns the block hash of the header, which is simply the keccak256 hash of its
+// RLP encoding.
+func (h *HeaderSerializable) Hash() common.Hash {
```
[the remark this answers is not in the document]
darioush: probably should add a comment so it doesn't cause further confusion. I've been confused about this before too

scripts/lint_allowed_eth_imports.sh
```diff
@@ -11,7 +11,7 @@ set -o pipefail
 # 4. Print out the difference between the search results and the list of specified allowed package imports from libevm.
 libevm_regexp='"github.com/ava-labs/libevm/.*"'
 allow_named_imports='eth\w\+ "'
-extra_imports=$(find . -type f \( -name "*.go" \) ! -path "./core/main_test.go" \
+extra_imports=$(find . -type f \( -name "*.go" \) ! -path "./core/main_test.go" ! -path "gen_*.go" \
```
darioush, line 14:
```suggestion
extra_imports=$(find . -type f \( -name "*.go" \) ! -path "./core/main_test.go" ! -name "gen_*.go" \
```

scripts/eth-allowed-packages.txt
```diff
@@ -10,6 +10,7 @@
 "github.com/ava-labs/libevm/common/math"
 "github.com/ava-labs/libevm/common/prque"
 "github.com/ava-labs/libevm/core/asm"
+"github.com/ava-labs/libevm/core/types"
```
darioush, line 13: I think you can remove this if you apply the suggestion in lint_allowed_eth_imports.sh
Since we still have the types package, it's better to be explicit and require libevm imports as ethtypes.

core/types/header_ext_test.go
```diff
@@ -0,0 +1,146 @@
...
+	t.Parallel()
+
+	testHeaderEncodings(t, json.Marshal, json.Unmarshal)
+}
+
+func testHeaderEncodings(t *testing.T, encode func(any) ([]byte, error), decode func([]byte, any) error) []byte {
+	t.Helper()
+
+	input, _ := headerWithNonZeroFields() // the Header carries the HeaderExtra so we can ignore it
+	b, err := encode(input)
```
darioush, line 50: nit: could we call this `bytes`?

core/types/header_ext_test.go
```diff
@@ -0,0 +1,146 @@
...
+	assert.Equal(t, "0x"+wantHashHex, gotHashHex, "Header.Hash()")
+}
+
+func TestHeaderJSON(t *testing.T) {
+	t.Parallel()
+
+	testHeaderEncodings(t, json.Marshal, json.Unmarshal)
+}
+
+func testHeaderEncodings(t *testing.T, encode func(any) ([]byte, error), decode func([]byte, any) error) []byte {
```
darioush, line 46: nit: how about `testHeaderEncodeDecode`?

core/types/header_ext_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	header, _ := headerWithNonZeroFields()
+	gotHashHex := header.Hash().Hex()
+	assert.Equal(t, "0x"+wantHashHex, gotHashHex, "Header.Hash()")
+}
+
+func TestHeaderJSON(t *testing.T) {
+	t.Parallel()
+
+	testHeaderEncodings(t, json.Marshal, json.Unmarshal)
```
darioush, line 43: nit: can we explicitly ignore the return value and mention we aren't testing the JSON vs a gold standard or similar?

core/types/block_ext_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+		empty := &Header{}
+
+		headerExtra := &HeaderExtra{}
+		extras.Header.Set(empty, headerExtra)
+
+		cpy := CopyHeader(empty)
+
+		want := &Header{
+			Difficulty: new(big.Int),
```
darioush, line 31: wow this comes to me as a surprise. I think we should fix this in OverrideNewEVMArgs so it sets difficulty to 0 not nil!

core/types/block_ext_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+func exportedFieldsDeepCopied[T interface {
+	Header | HeaderExtra
+}](t *testing.T, original, cpy *T) {
+	t.Helper()
+
+	v := reflect.ValueOf(*original)
+	typ := v.Type()
+	cp := reflect.ValueOf(*cpy)
+	for i := 0; i < v.NumField(); i++ {
```
darioush, line 66: nit: apparently this is a thing now:
`	for i := range v.NumField() {`

core/types/block_ext_test.go
```diff
@@ -0,0 +1,105 @@
...
+		if !field.IsExported() {
+			continue
+		}
+		switch field.Type.Kind() {
+		case reflect.Array, reflect.Uint64:
+			// Not pointers, but using explicit Kinds for safety
+			continue
+		}
+
+		t.Run(field.Name, func(t *testing.T) {
```
darioush, line 77: is it worth calling t.Run here?

core/types/block_ext_test.go
```diff
@@ -0,0 +1,105 @@
...
+		wantHeader, wantExtra := headerWithNonZeroFields()
+		assert.Equal(t, wantHeader, gotHeader)
+		assert.Equal(t, wantExtra, gotExtra)
+
+		exportedFieldsDeepCopied(t, header, gotHeader)
+		exportedFieldsDeepCopied(t, GetHeaderExtra(header), gotExtra)
+	})
+}
+
+func exportedFieldsDeepCopied[T interface {
```
darioush, line 58: should we rename this? `exportedFieldsPointToDifferentMemory`?

core/types/block_ext_test.go
darioush: https://github.com/ava-labs/coreth/pull/832

core/types/block_ext_test.go
qdm12: Indeed, indeed 😄 
Just FYI, careful when using intrange it behaves slightly differently when you use the loop variable AFTER the loop, see https://github.com/ckaznocha/intrange/issues/33

core/types/block_ext_test.go
qdm12: That's @ARR4N's technique, such that the field name is in the test name when an error occurs, and you don't have to add the field name to each `t.Errorf` / `assert.*` calls below. It's pretty good, with the only disadvantage it does add quite a few log lines when running tests with the `-v` flag.

core/types/block_ext_test.go
qdm12: Sure 😄 That was the original name written by ARR4N (well `fieldsPointToDifferentMemory`) and thought it was too many syllables - changed to `exportedFieldsPointToDifferentMemory`

core/types/header_ext_test.go
ARR4N: I originally called it `buf`, which is a common name in the standard lib for a buffer of bytes, but am happy with another word. I agree that `b` is too short for a variable used over 15 lines with a lot else going on.

core/types/header_ext_test.go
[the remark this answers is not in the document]
ARR4N: We don't have `HeaderWithNonZeroFields` so it's unnecessarily jarring IMO. And even when there are both I would assess each case (e.g. I assume that most of the time the unexported one can and should be tested via the exported one).

ARR4N (approved):

core/types/header_ext_test.go
qdm12: Should we test against a gold standard? 🤔

core/types/header_ext_test.go
qdm12: Renamed to `encoded` (named the return as well because the signature of the test helper wasn't really helpful knowing what bytes were returned)

core/types/imports.go
qdm12: Done in #800

core/types/imports.go
qdm12: Done in #800

core/types/block_ext.go
qdm12: I did change and then revert it 🤣 
Are we in a tie??!
let's leave it as is, I am personally 50% shared between the two approaches, I don't think it matters much.

core/types/header_ext.go
qdm12: That's a good point, and it would conflict less when doing a geth sync on the coreth side

core/types/header_ext_test.go
darioush: Up to you, I don't think it is strictly necessary like RLP. But also it is not harmful.

core/types/header_ext.go
qdm12: Actually both generators don't support type aliases, I created:

- https://github.com/fjl/gencodec/pull/25 to add support for it for JSON generation
- https://github.com/ava-labs/libevm/pull/154 to add support for it for RLP generation

I did re-generate them using local modified versions of the generators for this to unblock the situation.

core/types/gen_header_json.go
qdm12: Actually not, I did re-generate the JSON, and it produces Marshal/Unmarshal method names, which is fine here.

core/types/header_ext.go
qdm12: The files were generated before we used imports (i.e. `Bloom = ethttypes.Bloom`) and I did not re-generate them since then. Now both generators fail due to the type aliases, but with those 2 features/fixes mentioned above, I could re-generate them fine.

darioush: Let's open a coreth issue (and add a TODO comment at the codegen site) that notes this and continue IMO, what do you think?

core/types/header_ext_test.go
qdm12: Reverted to using a local `assertNonZero` for cases in the switch except for []uint8

qdm12 (dismissed):
Self approving 🪃 !

core/types/block_ext_test.go
```diff
@@ -93,13 +92,17 @@ func exportedFieldsDeepCopied[T interface {
...
+	switch {
+	case a == nil:
+		t.Errorf("a (%T) cannot be nil", a)
+	case b == nil:
+		t.Errorf("b (%T) cannot be nil", b)
+	case a == b:
 		t.Errorf("pointers to same memory")
 	}
+	// Note: no need to check `b` is of the same type as `a`, otherwise
+	// the memory address would be different as well.
```
ARR4N, line 107: (fyi) Not technically true, but sufficient for these purposes so no need to change it.

core/types/header_ext.go
```diff
@@ -143,8 +143,11 @@ func (h *HeaderSerializable) updateToExtras(extras *HeaderExtra) {
 	extras.BlockGasCost = h.BlockGasCost
 }
 
-//go:generate go run github.com/fjl/gencodec -type HeaderSerializable -field-override headerMarshaling -out gen_header_json.go
-//go:generate go run github.com/ava-labs/libevm/rlp/rlpgen -type HeaderSerializable -out gen_header_rlp.go
+// NOTE: both generators currently do not support type aliases.
+// We are using custom versions of these programs for now to support type aliases,
```
ARR4N, line 147: What do you mean by custom versions? These aren't captured by a change to, for example, the `go.sum` entry for `gencodec`.

core/types/header_ext_test.go
```diff
@@ -40,18 +40,23 @@ func TestHeaderRLP(t *testing.T) {
...
+	// Note we ignore the returned encoded bytes because we don't
+	// need to compare them to a JSON gold standard.
+	_ = testHeaderEncodeDecode(t, json.Marshal, json.Unmarshal)
 }
 
-func testHeaderEncodings(t *testing.T, encode func(any) ([]byte, error), decode func([]byte, any) error) []byte {
+func testHeaderEncodeDecode(t *testing.T,
+	encode func(any) ([]byte, error),
+	decode func([]byte, any) error) (
+	encoded []byte) {
```
ARR4N, line 51:
```suggestion
func testHeaderEncodeDecode(
	t *testing.T,
	encode func(any) ([]byte, error),
	decode func([]byte, any) error,
) (encoded []byte) {
```

Open and close parentheses, braces, and brackets at the same level of indentation. There is currently no visual cue to the reader w.r.t. a delineation between input args, output args, and implementation.

core/types/block_ext_test.go
qdm12: If `a` is not nil so it has a memory address, can there be another-type variable `b` at that same memory address? 🤔  How? Why?

core/types/header_ext.go
qdm12: See #833 
For now fjl/gencodec is not pinned to a version at all, and libevm is pinned to the version in go.mod.
I did modify them locally to generate with type aliases, until these are properly merged/released.

scripts/eth-allowed-packages.txt
```diff
@@ -10,7 +10,6 @@
 "github.com/ava-labs/libevm/common/math"
 "github.com/ava-labs/libevm/common/prque"
 "github.com/ava-labs/libevm/core/asm"
-"github.com/ava-labs/libevm/core/types"
```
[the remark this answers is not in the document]
qdm12: See https://github.com/ava-labs/coreth/pull/746#discussion_r1968787485 - thoughts on this?

core/types/block_ext_test.go
qdm12: What do you mean by better home 🏠 ? 🤔

core/types/block_ext_test.go
ARR4N: Memory can be interpreted however one wishes; e.g. this [valid use](https://pkg.go.dev/unsafe#Pointer:~:text=(1)%20Conversion%20of%20a%20*T1%20to%20Pointer%20to%20*T2.) of `unsafe.Pointer`. The chance of that happening here is negligible, hence why I said "sufficient for these purposes", but the reason I originally put a check is to be explicit instead of relying on assumptions.

core/types/header_ext.go
ARR4N: > I did modify them locally to generate with type aliases, until these are properly merged/released.

Code needs to be reproducible from the repo, not rely on some local changes held on a single developer's machine.

> See https://github.com/ava-labs/coreth/issues/833

LGTM and sufficient to rectify the above as long as it's addressed ASAP.

qdm12: I'll work on it in parallel while next PRs are undergoing reviews, and wrap it up before merging libevm -> master

ARR4N (approved):
Noting that [force push was a no-op](https://github.com/ava-labs/coreth/compare/8ca5efd0a4fe02ed26e2f8df742870707c9d4d09..57a67347b0e9b1b9eb2133c5d4925e620ade5ecc).

qdm12 (approved):
Re-approving (given no-op force push)

core/types/header_ext_test.go
darioush: I'm a big fan of one arg per line if all args don't fit in one line.

PR #749, opened 2025-01-16
chore(core/types): remove `Blocks`'s `Timestamp()` method

qdm12:
## Why this should be merged

In `core/types`, the `Block` method `Timestamp` does the same as the method `Time` and **is not present upstream** (this causes small difficulties in #750)

## How this works

- Define `core.BlockContext` implementing `contract.ConfigurationBlockContext` (same as subnet-evm)
- Use `core.NewBlockContext` to convert `types.Block` and `types.Header` for calls to `core.ApplyUpgrades`

## How this was tested

Existing CI passing

## Need to be documented?

No

No

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -246,7 +246,8 @@ func (g *Genesis) toBlock(db ethdb.Database, triedb *triedb.Database) *types.Blo
 	}
 
 	// Configure any stateful precompiles that should be enabled in the genesis.
-	err = ApplyPrecompileActivations(g.Config, nil, types.NewBlockWithHeader(head), statedb)
+	blockContext := NewBlockContext(head.Number, head.Time)
+	err = ApplyPrecompileActivations(g.Config, nil, blockContext, statedb)
 	if err != nil {
 		panic(fmt.Sprintf("unable to configure precompiles in genesis block: %v", err))
 	}
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -82,7 +82,8 @@ func (p *StateProcessor) Process(block *types.Block, parent *types.Header, state
 	)
 
 	// Configure any upgrades that should go into effect during this block.
-	err := ApplyUpgrades(p.config, &parent.Time, block, statedb)
+	blockContext := NewBlockContext(block.Number(), block.Time())
+	err := ApplyUpgrades(p.config, &parent.Time, blockContext, statedb)
 	if err != nil {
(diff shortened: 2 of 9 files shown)


precompile/contract/interfaces.go
```diff
@@ -73,6 +68,7 @@ type Configurator interface {
 		chainConfig precompileconfig.ChainConfig,
 		precompileconfig precompileconfig.Config,
 		state StateDB,
-		blockContext ConfigurationBlockContext,
+		blockNumber *big.Int,
+		blockTime uint64,
```
darioush, line 72: Nice! Simple types >> interface that returns stuff

qdm12: Also using an interface might suggest it could return a different block number or timestamp depending when it's called - that's why you would have methods. So just injecting dumb types makes it more obvious it doesn't change 😉

ceyonur (dismissed):

core/genesis.go
```diff
@@ -246,7 +246,8 @@ func (g *Genesis) toBlock(db ethdb.Database, triedb *triedb.Database) *types.Blo
 	}
 
 	// Configure any stateful precompiles that should be enabled in the genesis.
-	err = ApplyPrecompileActivations(g.Config, nil, types.NewBlockWithHeader(head), statedb)
+	block := types.NewBlockWithHeader(head)
```
darioush, line 249: nit: seems we can use `head`'s Number/Time directly without making a block

darioush (dismissed):

darioush (dismissed):

precompile/contract/interfaces.go
qdm12, line 72: Do we need this blockTime at the end? It doesn't seem used anywhere in subnet-evm. Waiting for a reply on https://github.com/ava-labs/subnet-evm/pull/1429/

darioush: Let's keep them for symmetry, previously exposed interfaces, and possible future use, imo.
(same comments here https://github.com/ava-labs/subnet-evm/pull/1429#discussion_r1927232609)

precompile/contract/interfaces.go
```diff
@@ -67,7 +67,7 @@ type AccessibleState interface {
 // ConfigurationBlockContext defines the interface required to configure a precompile.
 type ConfigurationBlockContext interface {
 	Number() *big.Int
-	Timestamp() uint64
+	Time() uint64
```
qdm12, line 70: Note this is non-breaking given the types.Block already implements `Time()`

darioush (approved):

ceyonur (approved):

PR #750, opened 2025-01-16
chore(core/types): use libevm Block

qdm12:


plugin/evm/vm_test.go
```diff
@@ -3909,7 +3909,7 @@ func TestParentBeaconRootBlock(t *testing.T) {
 				nil,
 				nil,
 				new(trie.Trie),
-				ethBlock.ExtData(),
+				*types.BlockExtras(ethBlock).Extdata,
```
darioush, line 3912: Could we add a function like `types.BlockExtData(ethBlock) []byte`?
It seems we could keep the nil check and `*types.BlockExtras(ethBlock).Extdata` seems repeated a few times.

plugin/evm/block.go
```diff
@@ -120,7 +120,7 @@ type Block struct {
 // newBlock returns a new Block wrapping the ethBlock type and implementing the snowman.Block interface
 func (vm *VM) newBlock(ethBlock *types.Block) (*Block, error) {
 	isApricotPhase5 := vm.chainConfigExtra().IsApricotPhase5(ethBlock.Time())
-	atomicTxs, err := atomic.ExtractAtomicTxs(ethBlock.ExtData(), isApricotPhase5, atomic.Codec)
+	atomicTxs, err := atomic.ExtractAtomicTxs(*types.BlockExtras(ethBlock).Extdata, isApricotPhase5, atomic.Codec)
```
darioush, line 123: same comment about `types.BlockExtData(ethBlock)`

plugin/evm/vm_test.go
```diff
@@ -2450,7 +2450,7 @@ func TestEmptyBlock(t *testing.T) {
 		false,
 	)
 
-	if len(emptyEthBlock.ExtData()) != 0 || types.HeaderExtras(emptyEthBlock.Header()).ExtDataHash != (common.Hash{}) {
+	if len(*types.BlockExtras(emptyEthBlock).Extdata) != 0 || types.HeaderExtras(emptyEthBlock.Header()).ExtDataHash != (common.Hash{}) {
```
darioush, line 2453: same comment about types.BlockExtData(ethBlock)

core/types/block_ext.go
```diff
@@ -4,50 +4,124 @@
...
+// This type uses BlockSerializable to encode and decode the extra fields
+// along with the upstream type for compatibility with existing network blocks.
+type BlockExtra struct {
+	Version uint32
+	Extdata *[]byte
+
+	// Fields removed from geth:
+	//  - withdrawals  Withdrawals
+	//  - ReceivedAt   time.Time
+	//  - ReceivedFrom interface{}
```
qdm12, line 26: note to self: remove these eventually

core/types/block_ext.go
```diff
@@ -4,50 +4,124 @@
...
-func (b *Block) setExtData(data []byte, recalc bool) {
+func setExtData(b *Block, data []byte, recalc bool) {
 	_data := make([]byte, len(data))
-	b.extdata = &_data
-	copy(*b.extdata, data)
+	BlockExtras(b).Extdata = &_data
+	copy(*BlockExtras(b).Extdata, data)
 	if recalc {
-		HeaderExtras(b.header).ExtDataHash = CalcExtDataHash(*b.extdata)
+		HeaderExtras(b.Header()).ExtDataHash = CalcExtDataHash(_data)
```
qdm12, line 105: this doesn't modify the block header, use `SetHeader` instead!

core/types/block_ext.go
```diff
@@ -61,7 +135,19 @@ func NewBlockWithExtData(
...
+func WrapWithTimestamp(b *Block) *BlockWithTimestamp {
+	return &BlockWithTimestamp{Block: b}
+}
+
+type BlockWithTimestamp struct {
+	*Block
+}
+
+func (b *BlockWithTimestamp) Timestamp() uint64 {
+	return b.Time()
```
qdm12, line 152: remove this once #749 is merged

core/types/imports.go
```diff
@@ -60,6 +62,10 @@ var (
 	SlimAccountRLP       = ethtypes.SlimAccountRLP
 	FullAccount          = ethtypes.FullAccount
 	FullAccountRLP       = ethtypes.FullAccountRLP
+	CalcUncleHash        = ethtypes.CalcUncleHash
+	NewBlock             = ethtypes.NewBlock
+	NewBlockWithHeader   = ethtypes.NewBlockWithHeader
+	CopyHeader           = ethtypes.CopyHeader
```
qdm12, line 68: Is it ok if I sort these alphabetically? To have a consistent order between coreth and subnet-evm? Or I can do a separate PR that works as well.

core/types/libevm.go
```diff
@@ -26,3 +29,16 @@ func WithHeaderExtras(h *Header, extra *HeaderExtra) *Header {
 	extras.Header.Set(h, extra)
 	return h
 }
+
+func BlockExtras(b *Block) *BlockExtra {
```
qdm12, line 33: This fits with the `HeaderExtras` function naming, but really it's itching me a bit: it returns a single `BlockExtra`, but it's a plural `BlockExtras`... could we have it as `GetBlockExtra` instead? 🤔

core/types/libevm.go
```diff
@@ -26,3 +29,16 @@ func WithHeaderExtras(h *Header, extra *HeaderExtra) *Header {
...
+	return extras.Block.Get(b)
+}
+
+func WithBlockExtras(b *Block, version uint32, extdata *[]byte) *Block {
```
qdm12, line 37: similar question here, would it be ok to rename to `WithBlockExtra`?

plugin/evm/vm_test.go
qdm12: Done, also with `types.BlockVersion`, and also unexported both extra fields `Version` and `ExtData` since they should not need to be exported anyway with those exported functions on the block.

core/types/libevm.go
darioush: We can rename it if it bothers you, subjectively I slightly prefer the less verbose one, but I acknowledge the inconsistency.
We can also take a final pass at renaming all these once it's put together.

qdm12:
Closed since it has to be done together with Body RLP hooks in #760 given both Body and Block share the same extras.

PR #753, opened 2025-01-16
put back Time_ (use PrecompileBlockContext to wrap BlockContext) 

darioush:
- vm.BlockContext is an upstream type, let's put it back
- since the reason for the Timestamp() and Number() methods to implement a precompile interface (controlled by us), we can wrap the upstream type and have a Time() method.

(Or it can be put back to Timestamp())


qdm12 (approved):
Totally awesome!! 💯 
Nice catch!

PR #754, opened 2025-01-17
chore: make function comments match function names

piguagua:
## Why this should be merged

make function comments match function names


qdm12:
Hi piguagua and thank you for your contribution!

The code you modified is a code originating from [go-ethereum](https://github.com/ethereum/go-ethereum). I suggest you fix it the go-ethereum repository "upstream", and we can incorporate it here in a future sync of go-ethereum automagically. Closing this since it's not the right place, but good catch anyway!

PR #756, opened 2025-01-17
use concrete returns / callsite interfaces where possible

darioush:
To follow the best practices of golang, interfaces should be defined by callsites and concrete types should be returned.
I think the interfaces packages makes the code more complex than it need be.
At the time when we were writing this code (and in some other places for example in avalanchego), sometimes we returned interfaces as a definition of behavior the client can expect, however exported/unexported methods should suffice to do this and the defining package returning an interface should really only be used when necessary.

Additionally, interfaces are necessary for situations where there are multiple implementations but here we only have a single implementation which is also mostly used as is in tests.


plugin/evm/atomic/sync/atomic_sync_extender.go
```diff
@@ -9,22 +9,43 @@ import (
...
-	"github.com/ava-labs/coreth/plugin/evm/atomic/state/interfaces"
 	"github.com/ava-labs/coreth/plugin/evm/message"
 	"github.com/ava-labs/coreth/plugin/evm/sync"
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
 )
 
 var _ sync.Extender = (*AtomicSyncExtender)(nil)
 
+type AtomicBackend interface {
```
darioush, line 20: IMO this can also be a concrete type, as it seems acceptable to me for sync to import state package.

ceyonur (approved):

PR #758, opened 2025-01-20
Atomic vm wrapper

ceyonur:
## Why this should be merged

This adds a way to extend the inner EVM and wraps it with atomic VM.  For demonstration it sets the network codec with atomic codec.

## How this works

This pull request introduces significant changes to the `plugin/evm` package to improve the extensibility and modularity of the EVM (Ethereum Virtual Machine) implementation. The primary focus is on creating a new `VM` wrapper and updating initialization procedures to support network codec management. Here are the most important changes:

### Unexported Shared EVM

* Renamed and unexported the Shared EVM (plugin/evm/vm.go) and provided constructors.

### New VM Wrapper and Initialization

* [`plugin/evm/atomic/vm/vm.go`](diffhunk://#diff-26a9954b366b563cb620189ad070df2b73b355692248686a6b86487fba21b449R1-R78): Introduced a new `VM` wrapper that implements multiple interfaces (`secp256k1fx.VM`, `block.ChainVM`, `block.BuildBlockWithContextChainVM`, `block.StateSyncableVM`). Added the `WrapVM` function to wrap an `InnerVM` and the `Initialize` method to set up the network codec.

### Factory Updates

* [`plugin/evm/factory.go`](diffhunk://#diff-324641d7c94bf2d58b9e247d895ad9d83b29cf4217f51011cb5c9d929ef51ac7R8-R12): Modified the `Factory` struct to return the new `VM` wrapper in the `New` and `NewPlugin` methods, allowing the EVM to be used as a plugin with the appropriate network codec setup. [[1]](diffhunk://#diff-324641d7c94bf2d58b9e247d895ad9d83b29cf4217f51011cb5c9d929ef51ac7R8-R12) [[2]](diffhunk://#diff-324641d7c94bf2d58b9e247d895ad9d83b29cf4217f51011cb5c9d929ef51ac7R22-R30)

### Codec Management

* [`plugin/evm/vm.go`](diffhunk://#diff-3be5ea9687f8af021d9baaa6584d174a7975b181b604e1a759e4ad2e40136912L161-L162): Removed the global `networkCodec` variable and added a `networkCodec` field to the `VM` struct. Implemented the `SetNetworkCodec` method to manage the codec. Updated various methods to use the `networkCodec` field instead of the global variable. [[1]](diffhunk://#diff-3be5ea9687f8af021d9baaa6584d174a7975b181b604e1a759e4ad2e40136912L161-L162) [[2]](diffhunk://#diff-3be5ea9687f8af021d9baaa6584d174a7975b181b604e1a759e4ad2e40136912L204-L210) [[3]](diffhunk://#diff-3be5ea9687f8af021d9baaa6584d174a7975b181b604e1a759e4ad2e40136912R221-R223) [[4]](diffhunk://#diff-3be5ea9687f8af021d9baaa6584d174a7975b181b604e1a759e4ad2e40136912R311-R318) [[5]](diffhunk://#diff-3be5ea9687f8af021d9baaa6584d174a7975b181b604e1a759e4ad2e40136912L552-R554) [[6]](diffhunk://#diff-3be5ea9687f8af021d9baaa6584d174a7975b181b604e1a759e4ad2e40136912L720-R722) [[7]](diffhunk://#diff-3be5ea9687f8af021d9baaa6584d174a7975b181b604e1a759e4ad2e40136912L1249-R1251) [[8]](diffhunk://#diff-3be5ea9687f8af021d9baaa6584d174a7975b181b604e1a759e4ad2e40136912L1503-R1505)

### Test Updates

* [`plugin/evm/vm_warp_test.go`](diffhunk://#diff-816005c76cbcf9a7b2e47bb3b79f791507ae3b21594cfbbad80318ce065d2e2dR15): Added initialization of the `networkCodec` variable in tests to ensure proper codec setup during test execution. [[1]](diffhunk://#diff-816005c76cbcf9a7b2e47bb3b79f791507ae3b21594cfbbad80318ce065d2e2dR15) [[2]](diffhunk://#diff-816005c76cbcf9a7b2e47bb3b79f791507ae3b21594cfbbad80318ce065d2e2dR66-R75)

### Main Function Update

* [`plugin/main.go`](diffhunk://#diff-0f251a0cc92f4617f492756a75ea58c88fd11c94bcdebf64f3557c6bb05f8537L32-R33): Updated the `main` function to use the new `Factory` struct for creating the EVM plugin, ensuring the proper setup of the network codec.

## How this was tested

Existing UTs should cover, but might be refactored.

## Need to be documented?

No

No

### DIFF
--- a/plugin/evm/atomic/vm/vm.go
+++ b/plugin/evm/atomic/vm/vm.go
@@ -22,12 +22,12 @@ var (
 	_ block.StateSyncableVM              = &VM{}
 )
 
-type ExtandableEVM interface {
+type ExtensibleEVM interface {
 	SetNetworkCodec(codec codec.Manager) error
 }
 
 type InnerVM interface {
-	ExtandableEVM
+	ExtensibleEVM
 	avalanchecommon.VM
 	secp256k1fx.VM
 	block.ChainVM
--- a/plugin/evm/atomic/vm/vm.go
+++ b/plugin/evm/atomic/vm/vm.go
@@ -16,10 +16,10 @@ import (
 )
 
 var (
-	_ secp256k1fx.VM                     = &VM{}
-	_ block.ChainVM                      = &VM{}
-	_ block.BuildBlockWithContextChainVM = &VM{}
-	_ block.StateSyncableVM              = &VM{}
+	_ secp256k1fx.VM                     = (*VM)(nil)
+	_ block.ChainVM                      = (*VM)(nil)
+	_ block.BuildBlockWithContextChainVM = (*VM)(nil)
+	_ block.StateSyncableVM              = (*VM)(nil)
 )
 
 type ExtensibleEVM interface {
@@ -36,7 +36,7 @@ type InnerVM interface {
 }
 
 type VM struct {
-	InnerVM // Inner EVM
+	InnerVM
 }
 
 func WrapVM(vm InnerVM) *VM {
--- a/plugin/evm/atomic/vm/vm.go
+++ b/plugin/evm/atomic/vm/vm.go
@@ -4,10 +4,10 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/ava-labs/avalanchego/codec"
 	avalanchedatabase "github.com/ava-labs/avalanchego/database"
 	avalanchecommon "github.com/ava-labs/avalanchego/snow/engine/common"
 	"github.com/ava-labs/coreth/plugin/evm/atomic/sync"
+	"github.com/ava-labs/coreth/plugin/evm/extension"
 	"github.com/ava-labs/coreth/plugin/evm/message"
 
 	"github.com/ava-labs/avalanchego/snow"
@@ -22,12 +22,7 @@ var (
 	_ block.StateSyncableVM              = (*VM)(nil)
 )
 
-type ExtensibleEVM interface {
-	SetNetworkCodec(codec codec.Manager) error
-}
-
 type InnerVM interface {
-	ExtensibleEVM
 	avalanchecommon.VM
 	secp256k1fx.VM
 	block.ChainVM
@@ -37,6 +32,16 @@ type InnerVM interface {
 
 type VM struct {
 	InnerVM
+}
+
+func NewAtomicExtensionConfig() (extension.ExtensionConfig, error) {
+	codec, err := message.NewCodec(sync.AtomicSyncSummary{})
+	if err != nil {
+		return extension.ExtensionConfig{}, fmt.Errorf("failed to create codec manager: %w", err)
+	}
+	return extension.ExtensionConfig{
+		NetworkCodec: codec,
+	}, nil
 }
 
 func WrapVM(vm InnerVM) *VM {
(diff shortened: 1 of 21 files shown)


plugin/evm/vm.go
```diff
@@ -201,17 +199,10 @@ func init() {
...
-	networkCodec, err = message.NewCodec(atomicsync.AtomicSyncSummary{})
-	if err != nil {
-		panic(fmt.Errorf("failed to create codec manager: %w", err))
-	}
 }
 
-// VM implements the snowman.ChainVM interface
-type VM struct {
+// sharedEvm implements the snowman.ChainVM interface
+type sharedEvm struct {
```
darioush, line 205: Could we put this back? "VM" seems sufficient and as the wrapped VM is in another package it seems that we can leave this.

plugin/evm/atomic/vm/vm.go
```diff
@@ -0,0 +1,78 @@
...
+type VM struct {
+	InnerVM // Inner EVM
+}
+
+func WrapVM(vm InnerVM) *VM {
+	return &VM{InnerVM: vm}
+}
+
+// Initialize implements the snowman.ChainVM interface
+func (vm *VM) Initialize(
```
darioush, line 47: possibly instead of `plugin/evm/atomic/vm`, we can have `plugin/evm/coreth` which defines the VM for use in avalanchego.

plugin/evm/block.go
```diff
@@ -112,12 +112,12 @@ func readMainnetBonusBlocks() (map[uint64]ids.ID, error) {
...
 	id        ids.ID
 	ethBlock  *types.Block
-	vm        *VM
+	vm        *sharedEvm
 	atomicTxs []*atomic.Tx
 }
 
 // newBlock returns a new Block wrapping the ethBlock type and implementing the snowman.Block interface
-func (vm *VM) newBlock(ethBlock *types.Block) (*Block, error) {
+func (vm *sharedEvm) newBlock(ethBlock *types.Block) (*Block, error) {
```
darioush, line 120: I assume some further changes to `block.go` is needed to fully have the atomic vm separated from the "base" evm that we can share with subnet-evm?

plugin/evm/vm.go
ceyonur: I dont want to export this as without default extensions it will break. we can check if required extensions are nil in initialize to make sure it won't break if we want to. but I still slightly think we should not directly expose vm.

plugin/evm/vm.go
```diff
@@ -314,14 +308,41 @@ type VM struct {
 	rpcHandlers []interface{ Stop() }
 }
 
+// NewDefaultEVM returns a new instance of the VM with default extensions
+// This should not be called if the VM is being extended
+func NewDefaultEVM() *sharedEvm {
```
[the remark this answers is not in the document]
ceyonur: I moved it to test for now, alternatively we can check these extension points in `Initialize` if they're nil, and if so we can provide some defaults in the `vm.Initialize()` directly.

plugin/evm/factory.go
```diff
@@ -16,8 +19,13 @@ var (
 	_ vms.Factory = &Factory{}
 )
 
+// TODO: either move this from plugin or move the VM itself
 type Factory struct{}
```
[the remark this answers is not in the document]
ceyonur: Normally We implement `vms.Factory` from avalanchego 
https://github.com/ava-labs/coreth/blob/master/plugin/evm/factory.go#L16

which is used here:
https://github.com/ava-labs/avalanchego/blob/add-l1-validators-api/node/node.go#L1240

but yea this can just be a function

plugin/evm/vm.go
darioush: I don't think `sharedEvm` is a good name here (what is it shared between? runtime? compile time?) Perhaps we can use a more descriptive name or `vm`?

 I don't disagree that unexported is better, but renaming this seems like it should be its own PR as it is not necessary for this PR and makes this PR difficult to follow.

plugin/evm/vm.go
```diff
@@ -314,14 +308,26 @@ type VM struct {
 	rpcHandlers []interface{ Stop() }
 }
 
+func newExtensibleEVM(isPlugin bool) *sharedEvm {
```
darioush, line 311: could we have a function like newVM that takes the `codec.Manager` and drop the SetNetworkCodec function below? 
It seems this type is not fully initialized even if it is constructed with a "new" function.

PR #759, opened 2025-01-21
chore(core/types): `Header` libevm `PostCopy` hook (5)

qdm12:
- Implement `PostCopy` header hook
- Remove `CopyHeader` and use imported one from libevm

➡️  [header copy hooks libevm PR](https://github.com/ava-labs/libevm/pull/106/files)

### DIFF
--- a/core/types/header_ext.go
+++ b/core/types/header_ext.go
@@ -84,21 +84,16 @@ func (h *HeaderExtra) DecodeJSON(eth *ethtypes.Header, input []byte) error {
 }
 
 func (h *HeaderExtra) PostCopy(dst *ethtypes.Header) {
-	extraCopy := &HeaderExtra{
+	cp := &HeaderExtra{
 		ExtDataHash: h.ExtDataHash,
 	}
-
 	if h.BlockGasCost != nil {
-		extraCopy.BlockGasCost = big.NewInt(0)
-		extraCopy.BlockGasCost.SetBytes(h.BlockGasCost.Bytes())
+		cp.BlockGasCost = new(big.Int).Set(h.BlockGasCost)
 	}
-
 	if h.ExtDataGasUsed != nil {
-		extraCopy.ExtDataGasUsed = big.NewInt(0)
-		extraCopy.ExtDataGasUsed.SetBytes(h.ExtDataGasUsed.Bytes())
+		cp.ExtDataGasUsed = new(big.Int).Set(h.ExtDataGasUsed)
 	}
-
-	SetHeaderExtra(dst, extraCopy)
+	SetHeaderExtra(dst, cp)
 }
 
 func (h *HeaderSerializable) updateFromEth(eth *ethtypes.Header) {
--- a/core/types/header_ext.go
+++ b/core/types/header_ext.go
@@ -84,7 +84,21 @@ func (h *HeaderExtra) DecodeJSON(eth *ethtypes.Header, input []byte) error {
 }
 
 func (h *HeaderExtra) PostCopy(dst *ethtypes.Header) {
-	panic("not implemented")
+	extraCopy := &HeaderExtra{
+		ExtDataHash: h.ExtDataHash,
+	}
+
+	if h.BlockGasCost != nil {
+		extraCopy.BlockGasCost = big.NewInt(0)
+		extraCopy.BlockGasCost.SetBytes(h.BlockGasCost.Bytes())
+	}
+
+	if h.ExtDataGasUsed != nil {
+		extraCopy.ExtDataGasUsed = big.NewInt(0)
+		extraCopy.ExtDataGasUsed.SetBytes(h.ExtDataGasUsed.Bytes())
+	}
+
+	SetHeaderExtra(dst, extraCopy)
 }
 
 func (h *HeaderSerializable) updateFromEth(eth *ethtypes.Header) {
--- a/core/types/header_ext_test.go
+++ b/core/types/header_ext_test.go
@@ -168,56 +168,6 @@ func assertNonZero[T interface {
 	}
 }
 
-func TestHeaderExtraPostCopy(t *testing.T) {
-	t.Parallel()
-
-	t.Run("empty_extra", func(t *testing.T) {
-		t.Parallel()
-
-		got := &ethtypes.Header{
-			ParentHash: common.Hash{1},
-		}
-		extra := &HeaderExtra{}
-
-		extra.PostCopy(got)
-
-		want := &ethtypes.Header{
-			ParentHash: common.Hash{1},
-		}
-		wantExtra := &HeaderExtra{}
-		SetHeaderExtra(want, wantExtra)
-
-		assert.Equal(t, want, got)
-		gotExtra := GetHeaderExtra(got)
-		assert.Equal(t, wantExtra, gotExtra)
-	})
-
-	t.Run("filled_extra", func(t *testing.T) {
-		got := &ethtypes.Header{
-			ParentHash: common.Hash{1},
-		}
-		extra := &HeaderExtra{
-			ExtDataHash:    common.Hash{2},
-			ExtDataGasUsed: big.NewInt(3),
-			BlockGasCost:   big.NewInt(4),
-		}
-
-		extra.PostCopy(got)
-
-		want := &ethtypes.Header{
-			ParentHash: common.Hash{1},
-		}
-		wantExtra := &HeaderExtra{
-			ExtDataHash:    common.Hash{2},
-			ExtDataGasUsed: big.NewInt(3),
-			BlockGasCost:   big.NewInt(4),
-		}
-		_ = WithHeaderExtra(want, wantExtra)
-
-		assert.Equal(t, want, got)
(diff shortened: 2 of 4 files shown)


darioush (approved):

core/types/header_ext.go
```diff
@@ -67,6 +67,25 @@ func (h *HeaderExtra) UnmarshalJSON(eth *ethtypes.Header, input []byte) error {
...
+		extraCopy.BlockGasCost = big.NewInt(0)
+		extraCopy.BlockGasCost.SetBytes(h.BlockGasCost.Bytes())
+	}
+
+	if h.ExtDataGasUsed != nil {
+		extraCopy.ExtDataGasUsed = big.NewInt(0)
+		extraCopy.ExtDataGasUsed.SetBytes(h.ExtDataGasUsed.Bytes())
+	}
+
+	cpy := ethtypes.CopyEthHeader(header)
```
qdm12, line 85: I chose to export the original eth `CopyHeader` as `CopyEthHeader`, otherwise we would have to re-implement here (I tried to unexport it as `copyHeader` before this change).

core/types/imports.go
```diff
@@ -60,6 +60,7 @@ var (
 	SlimAccountRLP       = ethtypes.SlimAccountRLP
 	FullAccount          = ethtypes.FullAccount
 	FullAccountRLP       = ethtypes.FullAccountRLP
+	CopyHeader           = ethtypes.CopyHeader
```
qdm12, line 63: nit move above alphabetically-ish

core/types/header_ext.go
```diff
@@ -86,7 +86,21 @@ func (h *HeaderExtra) DecodeJSON(eth *ethtypes.Header, input []byte) error {
...
+		extraCopy.BlockGasCost = big.NewInt(0)
+		extraCopy.BlockGasCost.SetBytes(h.BlockGasCost.Bytes())
+	}
+
+	if h.ExtDataGasUsed != nil {
+		extraCopy.ExtDataGasUsed = big.NewInt(0)
+		extraCopy.ExtDataGasUsed.SetBytes(h.ExtDataGasUsed.Bytes())
+	}
+
+	_ = WithHeaderExtra(dst, extraCopy)
```
ARR4N, line 103:
```suggestion
	cp := &HeaderExtra{
		ExtDataHash: h.ExtDataHash,
	}
	if h.BlockGasCost != nil {
		cp.BlockGasCost = new(big.Int).Set(h.BlockGasCost)
	}
	if h.ExtDataGasUsed != nil {
		cp.ExtDataGasUsed = new(big.Int).Set(h.ExtDataGasUsed)
	}
	WithHeaderExtra(dst, cp)
```

ARR4N, line 103: Probably beyond the scope of this PR, but this pattern suggests that `WithHeaderExtra` is incorrectly named as it should be `SetHeaderExtra`. The `With` prefix implies that it's returning a value but not _necessarily_ that it's setting it on the argument.

core/types/header_ext_test.go
```diff
@@ -242,3 +242,55 @@ func TestHeaderSerializable_updates(t *testing.T) {
...
+
+	t.Run("empty_extra", func(t *testing.T) {
+		t.Parallel()
+
+		got := &ethtypes.Header{
+			ParentHash: common.Hash{1},
+		}
+		extra := &HeaderExtra{}
+
+		extra.PostCopy(got)
```
ARR4N, line 257:
```suggestion
		got := CopyHeader(&ethtypes.Header{
			ParentHash: common.Hash{1},
		})
```

Not only does this simplify the test but it also ensures proper integration for the desired behaviour.

core/types/header_ext_test.go
```diff
@@ -242,3 +242,55 @@ func TestHeaderSerializable_updates(t *testing.T) {
...
+
+		want := &ethtypes.Header{
+			ParentHash: common.Hash{1},
+		}
+		wantExtra := &HeaderExtra{}
+		_ = WithHeaderExtra(want, wantExtra)
```
ARR4N, line 263: What's the intention behind this line? I think it's a no-op.

core/types/header_ext_test.go
```diff
@@ -242,3 +242,55 @@ func TestHeaderSerializable_updates(t *testing.T) {
...
+		got := &ethtypes.Header{
+			ParentHash: common.Hash{1},
+		}
+		extra := &HeaderExtra{
+			ExtDataHash:    common.Hash{2},
+			ExtDataGasUsed: big.NewInt(3),
+			BlockGasCost:   big.NewInt(4),
+		}
+
+		extra.PostCopy(got)
```
ARR4N, line 280:
```suggestion
		got := CopyHeader(WithHeaderExtra(
			&ethtypes.Header{
				ParentHash: common.Hash{1},
			},
			&HeaderExtra{
				ExtDataHash:    common.Hash{2},
				ExtDataGasUsed: big.NewInt(3),
				BlockGasCost:   big.NewInt(4),
			},
		))
```

Same rationale as above.

ARR4N (changes requested):

core/types/header_ext.go
ARR4N: This comment was only meant to be on the final line.

core/types/header_ext_test.go
```diff
@@ -242,3 +242,55 @@ func TestHeaderSerializable_updates(t *testing.T) {
...
+
+		want := &ethtypes.Header{
+			ParentHash: common.Hash{1},
+		}
+		wantExtra := &HeaderExtra{
+			ExtDataHash:    common.Hash{2},
+			ExtDataGasUsed: big.NewInt(3),
+			BlockGasCost:   big.NewInt(4),
+		}
+		_ = WithHeaderExtra(want, wantExtra)
```
ARR4N, line 290: As above. This doesn't appear to have any impact on the test (nor should it).

core/types/header_ext_test.go
ARR4N: I just tested locally and you'll also need to add in zero values for `Difficulty` and `Number` in the `want` because they'll be copied.

```go
			Difficulty: big.NewInt(0),
			Number:     big.NewInt(0),
```

Same applies to the other change I recommended for the test.

core/types/header_ext_test.go
ARR4N: Nevermind. I wasn't expecting `assert.Equal()` to compare unexported values but now I see why it's necessary. It should, instead be refactored and I'll submit a suggestion in a moment.

ARR4N, line 263:
```suggestion
		wantExtra := &HeaderExtra{}
		want := WithHeaderExtra(
			&ethtypes.Header{
				ParentHash: common.Hash{1},
				Number: big.NewInt(0),
				Difficulty: big.NewInt(0),
			},
			wantExtra,
		)
```

This addresses:

1. The need for the explicit zeroes due to using `CopyHeader()`.
2. Refactoring so the `WithHeaderExtra()` makes more sense.

core/types/header_ext.go
qdm12: Indeed changed in base branch from WithHeaderExtra -> SetHeaderExtra  0ecd53e973f12591adc0e1007f535898f285ae7d

core/types/header_ext_test.go
qdm12: In this case there is already `TestCopyHeader` which pretty much does the same and more, should we then remove this test?

ARR4N: It doesn't test that the `HeaderExtra` is empty so I think it's worth keeping.

ARR4N (changes requested):
Changing status back to changes requested. See my response re keeping the tests.

core/types/header_ext_test.go
qdm12: It does, given libevm's CopyHeader does call PostCopy or am I missing something?
Maybe a bit risky to rely on CopyHeader keeping PostCopy in the future though 🤔

ARR4N: Oh I thought you meant an upstream test of `CopyHeader`, not the one in `block_ext_test.go`. Yes, that's sufficient.

core/types/header_ext_test.go
qdm12: Test removed in favor of `TestCopyHeader`

core/types/header_ext_test.go
qdm12: Test removed in favor of `TestCopyHeader`

ceyonur (approved):

ARR4N (approved):
Delete all the things!

core/types/header_ext_test.go
```diff
@@ -168,4 +168,6 @@ func assertNonZero[T interface {
 	}
 }
 
+// Note [TestCopyHeader] tests the [HeaderExtra.PostCopy] method.
```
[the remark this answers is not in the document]
qdm12: I would prefer to keep it here, given PostCopy is defined in header_ext.go and a developer might be looking for PostCopy tests in header_ext_test.go more than in `block_ext_test.go`

PR #760, opened 2025-01-21
chore(core/types): remove block.go and import `Block` and `Body` from libevm (6)

qdm12:
## Why this should be merged

Remove core/types/block.go and one step closer to remove core/rawdb entirely

## How this works

- `BlockBodyExtra` extra payload defined for both `Block` and `Body`
  - hook `Copy` to be used in `Block` methods `Body`, `WithSeal`, `WithBody` and `WithWithdrawals`
  - hooks `BlockRLPFieldsForEncoding` and `BlockRLPFieldsForDecoding` for RLP codec of `Block`
  - hooks `BodyRLPFieldsForEncoding` and `BodyRLPFieldsForDecoding` for RLP codec of `Body`
- `SetBlockExtra` to set an extra payload `BlockBodyExtra` on a `Block`
- `SetBodyExtra` to set an extra payload `BlockBodyExtra` on a `Body`
- Block getter pure functions for extra fields: `BlockExtDataGasUsed`, `BlockGasCost`, `BlockExtData`, `BlockVersion`, `BlockExtDataGasUsed`, `BlockGasCost`
- New import from libevm's `core/types`: `Block`, `Blocks`, `Body`, `CalcUncleHash`, `NewBlock` and `NewBlockWithHeader`
- remove now unneeded `core/types/block.go`
- add test dependency github.com/google/go-cmp v0.7.0

## How this was tested

Existing and new tests

## Need to be documented?

No

No


core/types/imports.go
```diff
@@ -13,6 +13,7 @@ type (
 	AccessList        = ethtypes.AccessList
 	AccessTuple       = ethtypes.AccessTuple
 	AccessListTx      = ethtypes.AccessListTx
+	Body              = ethtypes.Body
```
qdm12, line 16: move down a bit so it fits alphabetically

core/types/body_ext.go
```diff
@@ -0,0 +1,77 @@
...
+// along with the upstream type for compatibility with existing network blocks.
+type BodyExtra struct {
+	Version uint32
+	ExtData *[]byte
+
+	// Fields removed from geth:
+	//  - withdrawals  Withdrawals
+}
+
+func (b *BodyExtra) EncodeRLP(eth *ethtypes.Body, writer io.Writer) error {
```
qdm12, line 25: Add test for this and `DecodeRLP`

core/types/libevm.go
```diff
@@ -26,3 +29,12 @@ func WithHeaderExtras(h *Header, extra *HeaderExtra) *Header {
 	extras.Header.Set(h, extra)
 	return h
 }
+
+func GetBodyExtra(b *Body) *BodyExtra {
```
qdm12, line 33: Add test for this and `WithBodyExtra`

core/types/body_ext.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"github.com/ava-labs/libevm/rlp"
+)
+
+func GetBodyExtra(b *Body) *BodyExtra {
+	return extras.Body.Get(b)
+}
+
+func WithBodyExtra(b *Body, extra *BodyExtra) *Body {
```
qdm12, line 14: Needs simple test

core/types/body_ext.go
```diff
@@ -0,0 +1,51 @@
...
+// along with the upstream type for compatibility with existing network blocks.
+type BodyExtra struct {
+	Version uint32
+	ExtData *[]byte
+
+	// Fields removed from geth:
+	//  - withdrawals  Withdrawals
+}
+
+func (b *BodyExtra) RLPFieldsForEncoding(body *Body) *rlp.Fields {
```
qdm12, line 31: Needs RLP encode/decode test

core/types/block.go
```diff
@@ -152,7 +143,15 @@ func (b *Block) EncodeRLP(w io.Writer) error {
 // Body returns the non-header content of the block.
 // Note the returned data is not an independent copy.
 func (b *Block) Body() *Body {
-	return &Body{b.transactions, b.uncles, b.version, b.extdata}
+	body := &Body{
```
qdm12, line 146: Needs test

core/types/body_ext.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+}
+
+func (b *BodyExtra) RLPFieldPointersForDecoding(body *Body) *rlp.Fields {
+	return &rlp.Fields{
+		Required: []any{
+			&body.Transactions,
+			&body.Uncles,
+			&b.Version,
+			&b.ExtData,
```
qdm12, line 48: Maybe a dumb remark, but note this doesn't use rlp.Nillable, I believe this is what we want correct?

core/rawdb/accessors_chain.go
```diff
@@ -522,7 +522,19 @@ func ReadBlock(db ethdb.Reader, hash common.Hash, number uint64) *types.Block {
...
+	extras := types.GetBlockExtra(block)
+	extras.Version = bodyExtra.Version
+	if bodyExtra.ExtData != nil {
+		extras.ExtData = common.PointerTo(make([]byte, len(*bodyExtra.ExtData)))
+		copy(*extras.ExtData, *bodyExtra.ExtData)
+	} else {
+		extras.ExtData = common.PointerTo(make([]byte, 0))
+	}
+
+	return block
```
qdm12, line 537: Note:

1. this is about to be thrown away and replaced by hooks in #772 
2. most of the logic comes from `WithExtData` which had to be changed because we cannot set block fields (header field) directly anymore, we need to use NewBlockWithHeader.

ethclient/ethclient.go
```diff
@@ -265,7 +265,10 @@ func (ec *client) getBlock(ctx context.Context, method string, args ...interface
 		}
 		txs[i] = tx.tx
 	}
-	return types.NewBlockWithHeader(head).WithBody(txs, uncles).WithExtData(body.Version, (*[]byte)(body.BlockExtraData)), nil
+	block := types.NewBlockWithHeader(head)
+	block = block.WithBody(types.Body{Transactions: txs, Uncles: uncles})
+	block = types.WithBlockExtra(block, body.Version, (*[]byte)(body.BlockExtraData), false)
+	return block, nil
```
qdm12, line 271: we could have this into one line but it would be even more confusing imo, especially since it would start with `types.WithBlockExtras(types.NewBlockWithHeader(head), ...)`

core/types/block.go
```diff
@@ -1,301 +0,0 @@
-// (c) 2019-2020, Ava Labs, Inc.
```
darioush, line 1: goodbye, old friend :)

core/types/block_ext.go
```diff
@@ -7,50 +7,117 @@ import (
...
-		return
+// Copy deep copies the [BlockBodyExtra] `b` and returns it.
+// It is notably used in the following functions:
+// - [Block]'s Body() method
+// - [Block]'s With*() methods
+func (b *BlockBodyExtra) Copy() *BlockBodyExtra {
+	cpy := *b
+	if b.ExtData == nil {
+		data := make([]byte, 0)
+		cpy.ExtData = &data
```
darioush, line 32: why do we copy nil as empty slice?

sync/client/client_test.go
```diff
@@ -266,7 +266,7 @@ func TestGetBlocks(t *testing.T) {
 
 				return responseBytes
 			},
-			expectedErr: "failed to unmarshal response: rlp: expected input list for types.extblock",
+			expectedErr: "failed to unmarshal response: rlp: expected List and ctx error context canceled",
```
darioush, line 269: where does `error context canceled` come from?

ethclient/ethclient.go
```diff
@@ -265,7 +265,7 @@ func (ec *client) getBlock(ctx context.Context, method string, args ...interface
 		}
 		txs[i] = tx.tx
 	}
-	return types.NewBlockWithHeader(head).WithBody(txs, uncles).WithExtData(body.Version, (*[]byte)(body.BlockExtraData)), nil
+	return types.NewBlockWithHeader(head).WithBody(types.Body{Transactions: txs, Uncles: uncles}), nil
```
darioush, line 268: why don't we use body here? seems we'd miss `WithExtData(body.Version, (*[]byte)(body.BlockExtraData))`

core/types/block_ext_test.go
```diff
@@ -106,3 +111,410 @@ func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
 	// Note: no need to check `b` is of the same type as `a`, otherwise
 	// the memory address would be different as well.
 }
+
+func TestBlockExtraGetWith(t *testing.T) {
```
darioush, line 115: nit: not sure I understand what "GetWith" means here?

core/types/block_ext_test.go
```diff
@@ -106,3 +111,410 @@ func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
...
+	// Check transactions in the following because unexported fields
+	// size and time are set by the RLP decoder.
+	require.Len(t, gotBody.Transactions, 1)
+	gotTx := gotBody.Transactions[0]
+	gotTxJSON, err := gotTx.MarshalJSON()
+	require.NoError(t, err)
+	wantTxJSON, err := testTx.MarshalJSON()
+	require.NoError(t, err)
+	assert.Equal(t, string(wantTxJSON), string(gotTxJSON))
+	gotBody.Transactions = nil
```
darioush, line 193: this seems like a bit of a weird way to check that Transactions and Uncles are equal.
Is there a reflect.DeepEqual or can we just check both Transactions and Uncles separately?

core/types/block_ext.go
```diff
@@ -7,50 +7,117 @@ import (
...
+	"github.com/ava-labs/libevm/rlp"
 )
 
-func (b *Block) WithExtData(version uint32, extdata *[]byte) *Block {
-	b.version = version
-	b.setExtDataHelper(extdata, false)
-	return b
+// BlockBodyExtra is a struct containing extra fields used by Avalanche
+// in the [Block] and [Body].
+type BlockBodyExtra struct {
```
darioush, line 16: nit: should we call this `BlockExtra`?

core/types/block_ext_test.go
```diff
@@ -106,3 +111,410 @@ func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
...
+		Uncles:       []*Header{uncle},
+		Withdrawals:  []*ethtypes.Withdrawal{{Index: 7}}, // ignored
+	}
+	extra := &BlockBodyExtra{
+		Version: 1,
+		ExtData: ptrTo([]byte{2}),
+	}
+	extras.Body.Set(body, extra)
+
+	b, err := rlp.EncodeToBytes(body)
```
darioush, line 177: nit: I still dislike this `b` name

core/types/block_ext_test.go
```diff
@@ -106,3 +111,410 @@ func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
...
+		Transactions: []*Transaction{tx},
+		Uncles:       []*Header{wantUncle},
+	}
+	wantBodyExtra := &BlockBodyExtra{
+		Version: 8,
+		ExtData: ptrTo([]byte{9}),
+	}
+	extras.Body.Set(wantBody, wantBodyExtra)
+
+	body := block.Body()
```
darioush, line 329:
```suggestion
	body := block.Body() // this is expected to perform a deep copy
```
or something like that

core/types/block_ext_test.go
```diff
@@ -106,3 +111,410 @@ func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
...
+			},
+		},
+		{
+			name: "with_recalc",
+			header: func() *Header {
+				header := &Header{}
+				extra := &HeaderExtra{
+					ExtDataHash: common.Hash{1}, // should be overwritten
+				}
+				SetHeaderExtra(header, extra)
```
darioush, line 436: I wonder if we should have a pattern like we have for the chain config like `HeaderAndExtras(&Header{}, &Extra{}) *Header`
as this is likely to be used in other tests as well.

core/types/block_ext.go
```diff
@@ -7,50 +7,117 @@ import (
...
+// - [Block]'s Body() method
+// - [Block]'s With*() methods
+func (b *BlockBodyExtra) Copy() *BlockBodyExtra {
+	cpy := *b
+	if b.ExtData == nil {
+		data := make([]byte, 0)
+		cpy.ExtData = &data
+	} else {
+		data := make([]byte, len(*b.ExtData))
+		copy(data, *b.ExtData)
```
ARR4N, line 35:
```suggestion
		data := slices.Clone(*b.ExtData)
```

core/types/block_ext.go
```diff
@@ -62,7 +129,16 @@ func NewBlockWithExtData(
...
-	b := NewBlock(header, txs, uncles, receipts, hasher)
-	b.setExtData(extdata, recalc)
-	return b
+	if recalc {
+		headerExtra := GetHeaderExtra(header)
+		headerExtra.ExtDataHash = CalcExtDataHash(extdata)
+	}
+	block := NewBlock(header, txs, uncles, receipts, hasher)
+	extdataCopy := make([]byte, len(extdata))
+	copy(extdataCopy, extdata)
```
ARR4N, line 138:
```suggestion
	extdataCopy := slices.Clone(extdata)
```

core/types/block_ext.go
```diff
@@ -62,7 +129,16 @@ func NewBlockWithExtData(
...
 	hasher TrieHasher, extdata []byte, recalc bool,
 ) *Block {
-	b := NewBlock(header, txs, uncles, receipts, hasher)
-	b.setExtData(extdata, recalc)
-	return b
+	if recalc {
+		headerExtra := GetHeaderExtra(header)
+		headerExtra.ExtDataHash = CalcExtDataHash(extdata)
+	}
+	block := NewBlock(header, txs, uncles, receipts, hasher)
```
ARR4N, line 136: (nit) I would move this to after the definition of `extra` as it's unused until then so there's no point in forcing the reader to parse and remember it right now.

```go
// ...
extra := {
    // ...
}
block := NewBlock(/*...*/)
extras.Block.Set(block, extra)
return block
```

core/types/block_test.go
```diff
@@ -113,8 +113,8 @@ func TestEIP1559BlockEncoding(t *testing.T) {
 	check("Time", block.Time(), uint64(1426516743))
 	check("Size", block.Size(), uint64(len(blockEnc)))
 	check("BaseFee", block.BaseFee(), new(big.Int).SetUint64(1_000_000_000))
-	check("ExtDataGasUsed", block.ExtDataGasUsed(), (*big.Int)(nil))
-	check("BlockGasCost", block.BlockGasCost(), (*big.Int)(nil))
+	check("ExtDataGasUsed", BlockExtDataGasUsed(&block), (*big.Int)(nil))
```
ARR4N, line 116: (optional) Having to always address it with `&block` suggests that it should be declared as a pointer in the first place. This would be in keeping with the values returned by constructors as well as all method receivers.

ethclient/ethclient.go
ARR4N: > seems we'd miss `WithExtData(body.Version, (*[]byte)(body.BlockExtraData))`

I agree. The extra needs to be set on the newly constructed `Block`. There's no point setting it on the ephemeral `Body` being passed to `WithBody` as that will just be cloned.

ARR4N (changes requested):
Partial review, covering everything except the large test file.

> [!CAUTION]
> Please do not force push to this branch as it will make it impossible to review only the diff between reviews.

core/rawdb/accessors_chain.go
```diff
@@ -522,7 +522,7 @@ func ReadBlock(db ethdb.Reader, hash common.Hash, number uint64) *types.Block {
 	if body == nil {
 		return nil
 	}
-	return types.NewBlockWithHeader(header).WithBody(body.Transactions, body.Uncles).WithExtData(body.Version, body.ExtData)
+	return types.NewBlockWithHeader(header).WithBody(*body)
```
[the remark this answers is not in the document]
ARR4N: It's only safe to rely on this pattern if the old one is as shown here, sourcing the `WithExtData()` fields from the `Body`. There is another point in the code where this wasn't the case and they were dropped.

core/types/block_ext_test.go
ARR4N, line 115: This test and the `Body` equivalent aren't really necessary as (a) they're fundamental guarantees of `libevm` and (b) they're implicitly tested by the RLP round-trip tests.

core/types/block_ext_test.go
ARR4N: I agree. I recommend using `go-cmp` and options like `cmp.Comparer` as [done here in `libevm`](https://github.com/ava-labs/libevm/blob/9f410ecc7ebeb68890e272d2c2b7fe7f37a82f9d/core/types/backwards_compat.libevm_test.go#L107). I'd use `Transaction.Hash()` and `Header.Hash()` as the equality checks.

core/types/block_ext_test.go
```diff
@@ -106,3 +111,410 @@ func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
...
+	// Check transactions in the following because unexported fields
+	// size and time are set by the RLP decoder.
+	gotTxs := gotBlock.Transactions()
+	require.Len(t, gotTxs, 1)
+	gotTx := gotTxs[0]
+	gotTxBin, err := gotTx.MarshalBinary()
+	require.NoError(t, err)
+	wantTxBin, err := testTx.MarshalBinary()
+	require.NoError(t, err)
+	assert.Equal(t, wantTxBin, gotTxBin)
```
ARR4N, line 263: See my [comment on the equivalent code above](https://github.com/ava-labs/coreth/pull/760/files#r1970522666).

core/types/block_ext_test.go
```diff
@@ -106,3 +111,410 @@ func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
...
+	extras.Body.Set(b, extra)
+
+	extra = extras.Body.Get(b)
+	wantExtra := &BlockBodyExtra{
+		Version: 1,
+	}
+	assert.Equal(t, wantExtra, extra)
+}
+
+func TestBodyExtraRLP(t *testing.T) {
```
ARR4N, line 157: This and the `Block` equivalent both need `coreth` golden data. An RLP round trip is only part of the story because the _exact_ encoding format is necessary.

core/types/block_ext_test.go
```diff
@@ -106,3 +111,410 @@ func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
...
+
+// TestBlockBody tests the [BlockBodyExtra.Copy] method is implemented correctly.
+func TestBlockBody(t *testing.T) {
+	t.Parallel()
+
+	header := &Header{ParentHash: common.Hash{1}}
+	tx := NewTransaction(0, common.Address{1}, big.NewInt(2), 3, big.NewInt(4), []byte{5})
+	txs := []*Transaction{tx}
+	uncles := []*Header{{ParentHash: common.Hash{6}}}
+	receipts := []*Receipt{{PostState: []byte{7}}}
```
ARR4N, line 303: If this is only a test of `BlockBodyExtra.Copy` then the rest of these fields are irrelevant and only distract from the test. The method doesn't have access to the `Block` so there's no concern about it mutating the fields.

core/types/block_ext_test.go
```diff
@@ -106,3 +111,410 @@ func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
...
+			assert.Equal(t, test.wantExtData, extData)
+
+			version := BlockVersion(block)
+			assert.Equal(t, test.wantVersion, version)
+
+			extDataGasUsed := BlockExtDataGasUsed(block)
+			assert.Equal(t, test.wantExtDataGasUsed, extDataGasUsed)
+
+			blockGasCost := BlockGasCost(block)
+			assert.Equal(t, test.wantBlockGasCost, blockGasCost)
```
ARR4N, line 391:
```suggestion
			assert.Equal(t, test.wantExtData, extData, "BlockExtData()")

			version := BlockVersion(block)
			assert.Equal(t, test.wantVersion, version, "BlockVersion()")

			extDataGasUsed := BlockExtDataGasUsed(block)
			assert.Equal(t, test.wantExtDataGasUsed, extDataGasUsed, "BlockExtDataGasUsed()")

			blockGasCost := BlockGasCost(block)
			assert.Equal(t, test.wantBlockGasCost, blockGasCost, "BlockGasCost()")
```

core/types/block_ext_test.go
```diff
@@ -106,3 +111,410 @@ func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
...
+		})
+	}
+}
+
+func TestNewBlockWithExtData(t *testing.T) {
+	t.Parallel()
+
+	// This transaction is generated beforehand because of its unexported time field being set
+	// on creation.
+	testTx := NewTransaction(0, common.Address{1}, big.NewInt(2), 3, big.NewInt(4), []byte{5})
```
ARR4N, line 401: (optional) If you use `cmp.Diff()` then you can fine-tune your comparisons instead of the one-size-fits-all `assert.Equal()` and `reflect.DeepEqual()`.

core/types/block_ext.go
```diff
@@ -7,50 +7,117 @@ import (
...
-	b.setExtDataHelper(extdata, false)
-	return b
+// BlockBodyExtra is a struct containing extra fields used by Avalanche
+// in the [Block] and [Body].
+type BlockBodyExtra struct {
+	Version uint32
+	ExtData *[]byte
+
+	// RLP block and body field removed from geth:
+	// - Withdrawals []*Withdrawal
```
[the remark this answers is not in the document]
qdm12: moved to `BodyRLPFieldsForEncoding` and `BlockRLPFieldsForEncoding` methods. I think it's enough to mention it on encoding functions, no need to mention it on decoding functions as well.

core/types/block_ext_test.go
```diff
@@ -106,3 +111,410 @@ func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
...
+				}
+				SetHeaderExtra(header, extra)
+				return header
+			},
+			extdata: []byte{2},
+			recalc:  true,
+			wantBlock: func() *Block {
+				header := &Header{}
+				extra := &HeaderExtra{
+					ExtDataHash: common.HexToHash("0xf2ee15ea639b73fa3db9b34a245bdfa015c260c598b211bf05a1ecc4b3e3b4f2"),
```
ARR4N, line 444:
```suggestion
					ExtDataHash: CalcExtDataHash([]byte{2}),
```

Otherwise it's just a magic number.

ARR4N (changes requested):

core/rawdb/accessors_chain.go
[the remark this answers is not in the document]
qdm12: Good point! Yes we shouldn't really document this, it's more of a safety-net and low-diff feature than something we should rely on otherwise.

core/types/block_ext.go
```diff
@@ -5,52 +5,132 @@ package types
...
+			&b.Version,
+			&b.ExtData,
+		},
+	}
+}
+
+// BlockRLPFieldPointersForEncoding returns the fields that should be encoded
+// for the [Block] and [BlockBodyExtra].
+// Note the following fields are added (+) and removed (-) compared to geth:
+// - (-) [ethtypes.Body] `Withdrawals` field
```
qdm12, line 77:
```suggestion
// - (-) [ethtypes.Block] `Withdrawals` field
```

core/types/block_ext_test.go
qdm12: Sorry that was due to late-evening-i-was-tired change removing unused exported methods such as GetBlockExtra and SetBlockExtra. No longer needed indeed.

core/types/block_ext.go
qdm12: counter-nit: I see your point, although defining block then extra kind of makes it a bit more natural given the call `extras.Block.Set(block, extra)` 🤔  Or maybe that's just my 🧠 being oddly twisted 😄

core/types/block_ext_test.go
qdm12: Test should be removed indeed (Arr4n is on it)

core/types/block_ext_test.go
qdm12: I think this gets computed internally with RLP, so you cannot set what you want really.

core/types/block_ext_test.go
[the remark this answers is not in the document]
qdm12: Nice, will review shortly!

> You have a failing test w.r.t. a nil vs empty ExtData slice, which I haven't debugged.

TIN (today I noticed, not tax number): slices.Clone clones a nil slice as a nil slice, not the same as


```go
	extdataCopy := make([]byte, len(extdata))
	copy(extdataCopy, extdata)
```

Now does it matter? 🔍

core/types/block_ext_test.go
qdm12: Tests refactored and this is no longer needed, resolving this!

core/types/block_ext_test.go
qdm12: Tests refactored and this is no longer needed, resolving this!

core/types/block_ext_test.go
qdm12: Tests refactored and this is no longer needed, resolving this!

core/types/block_ext_test.go
qdm12: Tests refactored and this is no longer needed, resolving this!

core/types/block_ext_test.go
qdm12: Tests refactored and this is no longer needed, resolving this!

core/types/block_test.go
qdm12: This is geth code, so not changing this 😉 But otherwise good point

core/types/block_ext_test.go
qdm12: tests refactored, no need for this anymore.

core/types/block_ext_test.go
qdm12: tests refactored, this is included.

core/types/block_ext_test.go
qdm12: tests refactored, this is included.

core/types/block_ext_test.go
qdm12: this is removed

core/types/block_ext.go
qdm12: Let's keep it as it is and limit possibility of problems introduced in the libevm branch.
We can eventually refactor later, but anyway changing this probably doesn't really matter i.e. performance-wise.

core/types/block_ext_test.go
darioush: why would that be the case? we could write:

```
header := WithHeaderExtra(&Header{}, &HeaderExtra{ExtDataHash: common.Hash{1})
```

I think it's okay to add production code that simplifies many tests, the idea is a valid idea (adding too much production code to help testing is a smell) but if most packages would need to redefine this then it seems valid.

An alternative is to put it in a test package.

In any case this is a side note not a blocker for this PR.

core/types/block_ext_test.go
qdm12: Added for both body and block (and checked in master branch)

core/types/block_ext.go
```diff
@@ -7,50 +7,117 @@ import (
...
+		Required: []any{
+			&block.Header,
+			&block.Txs,
+			&block.Uncles,
+			&b.Version,
+			&b.ExtData,
+		},
+	}
+}
+
```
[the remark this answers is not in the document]
qdm12: I think this is done right?

core/types/block_ext_test.go
qdm12: There was a conversation on #746 and one point ARR4N made was that it was a bit obscure to have a With* function that actually modifies the passed header, although some other With functions tend to deep copy the passed element to return it.
We could deep copy (maybe) the header/block/body to have With*Extra functions consistent with other With functions, but I think just having it as Set*Extra makes more sense 🤔

core/types/block_ext_test.go
```diff
@@ -112,108 +115,194 @@ func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
...
+	return block, extra
 }
 
-func TestBodyExtraGetWith(t *testing.T) {
+func TestBlockWithNonZeroFields(t *testing.T) {
 	t.Parallel()
 
-	b := &Body{}
+	block, extra := blockWithNonZeroFields()
+	t.Run("Block", func(t *testing.T) { allExportedFieldsSet(t, block, "ReceivedAt", "ReceivedFrom") })
```
darioush, line 194: nit: maybe 
```suggestion
	allowedExceptions := []string{"ReceivedAt", "ReceivedFrom")
	t.Run("Block", func(t *testing.T) { allExportedFieldsSet(t, block, allowedExceptions...) })
```

core/types/block_ext_test.go
darioush: I think this will make tests that set extra parameters in headers very verbose. But again, seems fine for now.

ethclient/ethclient.go
darioush: I'm not sure what the difference between calling .WithBody(body) and the code here is?
body already includes the extras, and calling WithBody also clones the extras.
Since this is shared code I would err on the side of matching upstream (as it will be in future versions)

darioush (approved):
added some optional suggestions

ethclient/ethclient.go
qdm12: The `body` variable is an `rpcBlock` type 😉 (easy to miss out! - maybe a bit of a bad var name, but that's out of scope I'd say)

core/types/block_ext.go
ARR4N: We'll be re-running the full chain history and also running a canary for a while to confirm.

core/types/block_ext_test.go
ARR4N: > one point ARR4N made was that it was a bit obscure to have a With* function that actually modifies the passed header

I think you've misunderstood as I'm with Darioush on this and suggested it at least a couple of times:

1. https://github.com/ava-labs/coreth/pull/746#discussion_r1965354704
2. https://github.com/ava-labs/coreth/pull/746#discussion_r1965914109

The thing I think we shouldn't do is use `With*` as a setter:

```go
hdr := &Header{...}
ex := &HeaderExtra{...}
WithHeaderExtra(hdr, ex) // reads weirdly as it suggests a dropped return argument
SetHeaderExtra(hdr, ex)  // reads naturally as an action performed on the arguments

hdr = WithHeaderExtra(
    &Header{...},
    &HeaderExtra{...},
) // reads naturally
```

I see no problem with both existing:

```go
func SetHeaderExtra(h *Header, e *HeaderExtra) {
    extras.Header.Set(h, e)
}

func WithHeaderExtra(h *Header, e *HeaderExtra) *Header {
  SetHeaderExtra(h, e)
  return h
}

File: core/types/block_ext_test.go [hunk 35]
```diff
@@ -112,108 +115,194 @@ func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
...
+	gotBody := new(Body)
+	err = rlp.DecodeBytes(encoded, gotBody)
+	require.NoError(t, err, "decode")
+	gotExtra := extras.Body.Get(gotBody)
 
-	extra = &BlockBodyExtra{
-		Version: 1,
+	wantBody, wantExtra := bodyWithNonZeroFields()
+
+	wantBody.Withdrawals = nil // this should be ignored
```
ARR4N, line 132:
```suggestion
	wantBody.Withdrawals = nil // this should be dropped by the RLP round trip
```

core/types/block_ext_test.go
```diff
@@ -112,108 +115,194 @@ func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
...
+
+	receipts := []*Receipt{{PostState: []byte{11}}}
+
+	block := NewBlock(header, txs, uncles, receipts, stubHasher{})
+	extra := &BlockBodyExtra{
+		Version: 12,
+		ExtData: &[]byte{13},
 	}
-	assert.Equal(t, wantExtra, extra)
+	extras.Block.Set(block, extra)
```
ARR4N, line 186:
```suggestion
	SetBlockExtra(block, extra)
```

Since it exists and future devs might look at tests as the expected way to do this.

core/types/block_ext_test.go
ARR4N: I agree. I could infer what they meant but there's a tiny bit of uncertainty without the variable explaining it.

core/types/block_ext_test.go
```diff
@@ -112,108 +115,194 @@ func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
...
+}
+
+// bodyWithNonZeroFields returns a [Body] and a [BlockBodyExtra],
+// each with all fields set to non-zero values.
+// The [BlockBodyExtra] extra payload is set in the [Body] via `extras.Body.Set`.
+//
+// NOTE: They can be used to demonstrate that RLP round-trip encoding
+// can recover all fields, but not that the encoded format is correct. This is
+// very important as the RLP encoding of a [Body] defines its hash.
+func bodyWithNonZeroFields() (*Body, *BlockBodyExtra) {
```
ARR4N, line 220: This whole function can be reduced to this:

```go
block, extra := blockWithNonZeroFields()
return block.Body(), extra
```

And then `TestBodyWithNonZeroFields()` can also be removed.

core/types/block_ext_test.go
```diff
@@ -240,98 +329,55 @@ func TestBlockExtraRLP(t *testing.T) {
...
-	tx := NewTransaction(0, common.Address{1}, big.NewInt(2), 3, big.NewInt(4), []byte{5})
-	txs := []*Transaction{tx}
-	uncles := []*Header{{ParentHash: common.Hash{6}}}
-	receipts := []*Receipt{{PostState: []byte{7}}}
-	block := NewBlock(header, txs, uncles, receipts, stubHasher{})
 	blockExtras := &BlockBodyExtra{
-		Version: 8,
-		ExtData: ptrTo([]byte{9}),
+		Version: 1,
+		ExtData: &[]byte{2},
```
ARR4N, line 367: These were deliberately made to be constants so that the reader of the code doesn't need to compare each field. Please revert the change.

This is an obviously correct declaration with only local information:
```go
_ = HeaderExtra{
  Version: version,
  ExtData: []byte{extDataByte},
}
```

Whereas this forces the reader to compare fields across the declarations (i.e. non-local information), opening up to errors:
```go
_ = HeaderExtra{
  Version: 1,
  ExtData: []byte{2},
}
```

core/types/block_ext_test.go
```diff
@@ -112,108 +115,194 @@ func assertDifferentPointers[T any](t *testing.T, a *T, b any) {
 	// the memory address would be different as well.
 }
 
-func TestBlockExtraGetWith(t *testing.T) {
+func TestBodyRLP(t *testing.T) {
```
ARR4N, line 118: Most of this is now a duplicate of `TestBodyExtraRLP()`.

There only needs to be `TestBodyRLP()` and `TestBlockRLP()` that both perform (a) RLP round-trip recovery; and (b) comparison to golden data.

While it's feasible to split them both into (a) and (b) components I think that adds unnecessary boilerplate when the golden-data comparison adds literally 2 lines (const and assertion).

ARR4N (changes requested):

core/types/block_ext_test.go
qdm12: I thought it would be easier to read a shorter code and that it was trivial to compare 😄 
I did revert it but with a version constant + `extData` variable set to `&[]byte{2}` - otherwise the reader would still have to check the expression is the same for both ExtData fields imo.

core/types/block_ext_test.go
qdm12: Done 👍

core/types/header_ext_test.go
```diff
@@ -117,22 +118,35 @@ func headerWithNonZeroFields() (*Header, *HeaderExtra) {
 	return header, extra
 }
 
-func allExportedFieldsSet[T interface {
+func allFieldsSet[T interface {
 	Header | HeaderExtra | Block | Body | BlockBodyExtra
 }](t *testing.T, x *T, ignoredFields ...string) {
+	t.Helper()
```
ARR4N, line 124: I think `t.Helper()` here isn't very useful because it will report all errors as being at 78/79 instead of where they're actually occurring.

core/types/header_ext_test.go
```diff
@@ -117,22 +118,35 @@ func headerWithNonZeroFields() (*Header, *HeaderExtra) {
...
 		t.Run(field.Name, func(t *testing.T) {
-			switch f := v.Field(i).Interface().(type) {
+			fieldValue := v.Field(i)
+			if !field.IsExported() {
+				// Note: we need to check unexported fields especially for [Block].
+				if fieldValue.Kind() == reflect.Ptr {
+					require.Falsef(t, fieldValue.IsNil(), "field %q is nil", field.Name)
+				}
+				fieldValue = reflect.NewAt(fieldValue.Type(), unsafe.Pointer(fieldValue.UnsafeAddr())).Elem() //nolint:gosec
+				assert.NotZero(t, fieldValue.Interface())
```
ARR4N, line 145: (optional) You could just let it run through the switch like the exported ones.

core/types/block_ext_test.go
```diff
@@ -179,72 +146,46 @@ func blockWithNonZeroFields() (*Block, *BlockBodyExtra) {
 	receipts := []*Receipt{{PostState: []byte{11}}}
 
 	block := NewBlock(header, txs, uncles, receipts, stubHasher{})
+	withdrawals := []*ethtypes.Withdrawal{{Index: 12}}
```
ARR4N, line 149: IIUC you were incrementing to make the values easier to correlate when debugging, so you should bump the ones in the extra.

core/types/block_ext_test.go
ARR4N: Yeah it's of negligible size and relies entirely on existing code so might as well.

core/types/block_ext_test.go
```diff
@@ -262,22 +203,7 @@ func headerHashComparer() cmp.Option {
...
-		Transactions: []*Transaction{tx},
-		Uncles:       []*Header{uncle},
-		Withdrawals:  []*ethtypes.Withdrawal{{Index: 7}}, // ignored
-	}
-	extra := &BlockBodyExtra{
-		Version: 1,
-		ExtData: &[]byte{2},
-	}
-	extras.Body.Set(body, extra)
+	body, wantExtra := bodyWithNonZeroFields()
```
ARR4N, line 206: I would re-call `bodyWithNonZeroFields()` later to generate a fresh copy of the two objects. It's a simple code change that gives strong guarantees in return.

core/types/block_ext_test.go
```diff
@@ -299,39 +225,22 @@ func TestBodyExtraRLP(t *testing.T) {
...
-
-	withdrawals := []*ethtypes.Withdrawal{{Index: 9}} // ignored
-	block = block.WithWithdrawals(withdrawals)
-
-	extra := &BlockBodyExtra{
-		Version: 10,
-		ExtData: &[]byte{11},
-	}
-	extras.Block.Set(block, extra)
+	block, wantExtra := blockWithNonZeroFields()
```
ARR4N, line 243: As above re calling it again for guarantees.

ARR4N (dismissed):

core/types/header_ext_test.go
qdm12: Removed since it was actually ineffective given we run subtests in this function (TIL/TIN this!)

core/types/block_ext.go
qdm12: Let's YOLO it then, changed to just

```go
	if b.ExtData != nil {
		data := slices.Clone(*b.ExtData)
		cpy.ExtData = &data
	}
```

core/types/block_ext_test.go
qdm12: Noted, I'll do a PR after this one to add With* functions for header/block/body and use it in tests where needed.

qdm12:
[Last force push](https://github.com/ava-labs/coreth/compare/a770faa6e3ba84cec31effc325694b7768bab055..37797f0a5f79f028def67a1be952ee32711838a4) adding missing commit pgp signatures on a few commits, sorry about that.

darioush (approved):

qdm12:
FYI auto-merge is enabled, so don't approve if you would like some changes 😉

ARR4N (approved):

PR #765, opened 2025-01-24
fix panic if config duration is set to 0

darioush:
Avoids a panic if the node is not configured correctly.

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -1120,6 +1120,11 @@ func (vm *VM) initBlockBuilding() error {
 	vm.builder = vm.NewBlockBuilder(vm.toEngine)
 	vm.builder.awaitSubmittedTxs()
 
+	var p2pValidators p2p.ValidatorSet = &validatorSet{}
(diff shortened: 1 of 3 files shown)


qdm12 (approved):
Looks good

ceyonur (approved):

StephenButtolph (approved):
lgtm

StephenButtolph:
missed it by that much :(

plugin/evm/vm.go
```diff
@@ -1120,6 +1120,11 @@ func (vm *VM) initBlockBuilding() error {
 	vm.builder = vm.NewBlockBuilder(vm.toEngine)
 	vm.builder.awaitSubmittedTxs()
 
+	var p2pValidators p2p.ValidatorSet = &validatorSet{}
```
joshua-kim, line 1123: This will work but might leave confusing logs in `DEBUG` as we log when we drop messages because it's not a validator. An alternative would be to use `p2p.NoOpHandler`

PR #770, opened 2025-01-28
Add F-Upgrade upgrades.Config support

StephenButtolph:
## Why this should be merged

As we are starting to prepare for the F-upgrade, we need to add handling of the fork time.

## How this works

Basically just looked at every instance of the Etna usage and added corresponding FUpgrade support.

## How this was tested

CI

## Need to be documented?

No.

No.

### DIFF
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -1105,6 +1105,10 @@ func overrideConfig(original *params.ChainConfig, override *params.ChainConfig) 
 		copy.EtnaTimestamp = timestamp
 		canon = false
 	}
+	if timestamp := override.FUpgradeTimestamp; timestamp != nil {
+		copy.FUpgradeTimestamp = timestamp
+		canon = false
+	}
 	if timestamp := override.CancunTime; timestamp != nil {
 		copy.CancunTime = timestamp
 		canon = false
--- a/params/network_upgrades.go
+++ b/params/network_upgrades.go
@@ -204,7 +204,7 @@ func (n *NetworkUpgrades) Description() string {
 	banner += fmt.Sprintf(" - Cortina Timestamp:                @%-10v (https://github.com/ava-labs/avalanchego/releases/tag/v1.10.0)\n", ptrToString(n.CortinaBlockTimestamp))
 	banner += fmt.Sprintf(" - Durango Timestamp:                @%-10v (https://github.com/ava-labs/avalanchego/releases/tag/v1.11.0)\n", ptrToString(n.DurangoBlockTimestamp))
 	banner += fmt.Sprintf(" - Etna Timestamp:                   @%-10v (https://github.com/ava-labs/avalanchego/releases/tag/v1.12.0)\n", ptrToString(n.EtnaTimestamp))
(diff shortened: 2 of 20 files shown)


eth/tracers/api.go
```diff
@@ -1105,6 +1105,10 @@ func overrideConfig(original *params.ChainConfig, override *params.ChainConfig)
 		copy.EtnaTimestamp = timestamp
 		canon = false
 	}
+	if timestamp := override.FUpgradeTimestamp; timestamp != nil {
+		copy.FUpgradeTimestamp = timestamp
+		canon = false
+	}
```
StephenButtolph, line 1111: Should this be after the Cancun check (because Cancun was ~mostly activated with Etna)

params/network_upgrades.go
```diff
@@ -190,7 +203,8 @@ func (n *NetworkUpgrades) Description() string {
 	banner += fmt.Sprintf(" - Banff Timestamp:                  @%-10v (https://github.com/ava-labs/avalanchego/releases/tag/v1.9.0)\n", ptrToString(n.BanffBlockTimestamp))
 	banner += fmt.Sprintf(" - Cortina Timestamp:                @%-10v (https://github.com/ava-labs/avalanchego/releases/tag/v1.10.0)\n", ptrToString(n.CortinaBlockTimestamp))
 	banner += fmt.Sprintf(" - Durango Timestamp:                @%-10v (https://github.com/ava-labs/avalanchego/releases/tag/v1.11.0)\n", ptrToString(n.DurangoBlockTimestamp))
-	banner += fmt.Sprintf(" - Etna Timestamp:               @%-10v (https://github.com/ava-labs/avalanchego/releases/tag/v1.12.0)\n", ptrToString(n.EtnaTimestamp))
+	banner += fmt.Sprintf(" - Etna Timestamp:                   @%-10v (https://github.com/ava-labs/avalanchego/releases/tag/v1.12.0)\n", ptrToString(n.EtnaTimestamp))
```
StephenButtolph, line 206: Minor spacing fix here as well to align Etna with the other upgrades.

eth/tracers/api.go
darioush: Let's keep our upgrades separate from the eth ones.

darioush (approved):

params/network_upgrades.go
```diff
@@ -177,6 +184,12 @@ func (n *NetworkUpgrades) IsEtna(time uint64) bool {
 	return isTimestampForked(n.EtnaTimestamp, time)
 }
 
+// IsFUpgrade returns whether [time] represents a block
```
qdm12, line 187: brackets won't work in that case, use backtickets or nothing to not mislead the docs reader / dev trying to right click it.

```suggestion
// IsFUpgrade returns whether `time` represents a block
```

qdm12 (approved):

params/network_upgrades.go
StephenButtolph: I think it probably makes sense to fix `IsFUpgrade`, `IsEtna`, ect all in the same PR. After talking with @darioush it seems like this makes the most sense to do after the libevm migration.

PR #772, opened 2025-01-29
chore(core/rawdb): use libevm core/rawdb

qdm12:
## Why this should be merged

Gets rid of most of core/rawdb geth code using libevm.

You might want to review the PR commit by commit, given historically, work for this PR was done as:

- First stage (see commits) was to get code as close to geth code (libevm main branch) as possible.
- Second stage was identifying the differences: `ReadBlock`'s use of extras and `InspectDatabase`
- Third stage was implement libevm hooks and other change and use libevm core/rawdb (last commit)

## How this works

- Move our own code to `_ext.go` files
- Import symbols from libevm in imports.go
- Use Block and Body hooks previously implemented

## How this was tested

CI passing

## Need to be documented?

No

No


core/rawdb/accessors_metadata.go
```diff
@@ -171,108 +185,15 @@ func UpdateUncleanShutdownMarker(db ethdb.KeyValueStore) {
...
-// the node has run with pruning disabled at some pooint.
-func HasPruningDisabled(db ethdb.KeyValueStore) (bool, error) {
-	return db.Has(pruningDisabledKey)
-}
-
-// DeletePruningDisabled deletes the marker indicating that the node has
-// run with pruning disabled.
-func DeletePruningDisabled(db ethdb.KeyValueStore) error {
-	return db.Delete(pruningDisabledKey)
-}
```
qdm12, line 257: This is unused in the entire codebase, should it be removed?

core/rawdb/accessors_chain_test.go
```diff
@@ -276,17 +378,12 @@ func TestBlockReceiptStorage(t *testing.T) {
 	receipts := []*types.Receipt{receipt1, receipt2}
 
 	// Check that no receipt entries are in a pristine database
-	header := &types.Header{Number: big.NewInt(0), Extra: []byte("test header")}
-	hash := header.Hash()
+	hash := common.BytesToHash([]byte{0x03, 0x14})
```
qdm12, line 381: This is original geth code, testing less things - is this ok to remove the additional checks we had added? 🤔 I suppose worth it anyway, but I wanted to double check still.

darioush: seems fine, if in doubt we can replicate the tests in an `ext` package.

core/rawdb/accessors_metadata_ext.go
```diff
@@ -0,0 +1,112 @@
...
+
+import (
+	"time"
+
+	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/libevm/ethdb"
+	"github.com/ava-labs/libevm/rlp"
+)
+
+// writeTimeMarker writes a marker of the current time in the db at [key]
```
darioush, line 12: nit:
```suggestion
// writeTimeMarker writes a marker of the current time in the db at `key`
```
and in other places

core/rawdb/accessors_metadata_ext.go
```diff
@@ -0,0 +1,112 @@
...
+
+// ReadAcceptorTip reads the hash of the last accepted block that was fully processed.
+// If there is no value present (the index is being initialized for the first time), then the
+// empty hash is returned.
+func ReadAcceptorTip(db ethdb.KeyValueReader) (common.Hash, error) {
+	has, err := db.Has(acceptorTipKey)
+	if err != nil {
+		return common.Hash{}, err
+	} else if !has {
+		// If the index is not present on disk, the [acceptorTipKey] index has not been initialized yet.
```
darioush, line 104:
```suggestion
		// If the index is not present on disk, the `acceptorTipKey` index has not been initialized yet.
```

core/rawdb/database_ext.go
```diff
@@ -0,0 +1,71 @@
+// (c) 2025, Ava Labs, Inc.
```
darioush, line 1: seems should match the 2 line format for the copyright header
(also in other places)

darioush (approved):

core/rawdb/database_ext.go
qdm12: Good point!
We definitely should setup goheader in our CI (later, probably after the subnet-evm/coreth merge!)

core/rawdb/accessors_metadata_ext.go
qdm12: Fixed - also here's another weekend project: a linter linting wrong uses of squared brackets for non-resolvable godoc links in comments!!

qdm12:
This PR is further improved/cleaned up with #851

core/rawdb/accessors_metadata_ext.go
```diff
@@ -0,0 +1,114 @@
...
+	var lastRun uint64
+	if err := rlp.DecodeBytes(data, &lastRun); err != nil {
+		return time.Time{}, err
+	}
+
+	return time.Unix(int64(lastRun), 0), nil
+}
+
+// deleteTimeMarker deletes any value stored at `key`
+func deleteTimeMarker(db ethdb.KeyValueStore, key []byte) error {
```
ceyonur, line 39: do we need this? It's fine if we want to keep CRUD but seems it just calls delete on the db.

core/rawdb/accessors_metadata_ext.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"time"
+
+	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/libevm/ethdb"
+	"github.com/ava-labs/libevm/rlp"
+)
+
+// writeTimeMarker writes a marker of the current time in the db at `key`
+func writeTimeMarker(db ethdb.KeyValueStore, key []byte) error {
```
ceyonur, line 15: nit: `writeCurrentTimeMarker`

core/rawdb/accessors_metadata_ext.go
```diff
@@ -0,0 +1,114 @@
...
+
+	return time.Unix(int64(lastRun), 0), nil
+}
+
+// deleteTimeMarker deletes any value stored at `key`
+func deleteTimeMarker(db ethdb.KeyValueStore, key []byte) error {
+	return db.Delete(key)
+}
+
+// WriteOfflinePruning writes a marker of the last attempt to run offline pruning
```
ceyonur, line 43: nit: can we specify "time" marker in comments?

core/rawdb/accessors_metadata_ext.go
qdm12: Definitely does not look like we need this, removed.

core/rawdb/accessors_metadata_ext.go
qdm12: Done!

core/rawdb/accessors_metadata_ext.go
qdm12: Done, indeed good point!

core/rawdb/database_ext.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/ava-labs/libevm/common"
+	ethrawdb "github.com/ava-labs/libevm/core/rawdb"
+	"github.com/ava-labs/libevm/ethdb"
+)
+
+// InspectDatabase traverses the entire database and checks the size
+// of all different categories of data.
+func InspectDatabase(db ethdb.Database, keyPrefix, keyStart []byte) error {
+	var (
```
ARR4N, line 17: Isn't this going to conflict with #850? Perhaps just add that code here?

core/rawdb/accessors_state_sync.go
```diff
@@ -191,3 +191,32 @@ func GetLatestSyncPerformed(db ethdb.Iteratee) uint64 {
 	}
 	return latestSyncPerformed
 }
+
+// clearPrefix removes all keys in db that begin with prefix and match an
+// expected key length. `keyLen` should include the length of the prefix.
```
ARR4N, line 196:
```suggestion
// expected key length. `keyLen` must include the length of the prefix.
```

core/rawdb/accessors_state_sync.go
```diff
@@ -191,3 +191,32 @@ func GetLatestSyncPerformed(db ethdb.Iteratee) uint64 {
...
+func clearPrefix(db ethdb.KeyValueStore, prefix []byte, keyLen int) error {
+	it := db.NewIterator(prefix, nil)
+	defer it.Release()
+
+	batch := db.NewBatch()
+	for it.Next() {
+		key := common.CopyBytes(it.Key())
+		if len(key) != keyLen {
+			// avoid deleting keys that do not match the expected length
```
ARR4N, line 205:
```suggestion
```

It's already stated in the function comment and is obvious from context.

core/rawdb/accessors_metadata_ext.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return time.Time{}, err
+	}
+
+	var lastRun uint64
+	if err := rlp.DecodeBytes(data, &lastRun); err != nil {
+		return time.Time{}, err
+	}
+
+	return time.Unix(int64(lastRun), 0), nil
```
ARR4N, line 35:
```suggestion
	var unix uint64
	if err := rlp.DecodeBytes(data, &unix); err != nil {
		return time.Time{}, err
	}

	return time.Unix(int64(unix), 0), nil
```

`lastRun` only makes sense in the context of the call sites, not this function in isolation.

core/rawdb/accessors_metadata_ext.go
```diff
@@ -0,0 +1,109 @@
...
+
+// ReadAcceptorTip reads the hash of the last accepted block that was fully processed.
+// If there is no value present (the index is being initialized for the first time), then the
+// empty hash is returned.
+func ReadAcceptorTip(db ethdb.KeyValueReader) (common.Hash, error) {
+	has, err := db.Has(acceptorTipKey)
+	if err != nil {
+		return common.Hash{}, err
+	} else if !has {
+		// If the index is not present on disk, the [acceptorTipKey] index has not been initialized yet.
```
ARR4N, line 101:
```suggestion
		return common.Hash{}, err
	}
	if !has {
		// If the index is not present on disk, the [acceptorTipKey] index has not been initialized yet.
```

The `else` is redundant while sending a subtle signal to the reader that they should remember the original condition.

core/rawdb/accessors_metadata_ext.go
```diff
@@ -0,0 +1,109 @@
...
+		return common.Hash{}, nil
+	}
+	h, err := db.Get(acceptorTipKey)
+	if err != nil {
+		return common.Hash{}, err
+	}
+	return common.BytesToHash(h), nil
```
ARR4N, line 108: I think you should check the length of `h`. I've been bitten many times by false assumptions (of my own and of others) when using the `common` converters.

core/rawdb/accessors_snapshot_ext.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/libevm/ethdb"
+	"github.com/ava-labs/libevm/log"
+)
+
+// ReadSnapshotBlockHash retrieves the hash of the block whose state is contained in
+// the persisted snapshot.
+func ReadSnapshotBlockHash(db ethdb.KeyValueReader) common.Hash {
+	data, _ := db.Get(snapshotBlockHashKey)
```
ARR4N, line 15: Why not check and return the error?

core/rawdb/accessors_snapshot_ext.go
```diff
@@ -0,0 +1,43 @@
...
+// invalid.
+func DeleteSnapshotBlockHash(db ethdb.KeyValueWriter) {
+	if err := db.Delete(snapshotBlockHashKey); err != nil {
+		log.Crit("Failed to remove snapshot block hash", "err", err)
+	}
+}
+
+// IterateAccountSnapshots returns an iterator for walking all of the accounts in the snapshot
+func IterateAccountSnapshots(db ethdb.Iteratee) ethdb.Iterator {
+	return NewKeyLengthIterator(db.NewIterator(SnapshotAccountPrefix, nil), len(SnapshotAccountPrefix)+common.HashLength)
```
ARR4N, line 42:
```suggestion
	return NewKeyLengthIterator(
		db.NewIterator(SnapshotAccountPrefix, nil),
		len(SnapshotAccountPrefix)+common.HashLength,
	)
```

There are a lot of parentheses and it's difficult to parse how they pair up. I found the second argument very confusing.

ARR4N (changes requested):

core/rawdb/database_ext.go
qdm12: That's fine, I'll take this into account and handle it accordingly depending which gets merged first 😉

core/rawdb/accessors_snapshot_ext.go
qdm12: Totally agreed, but I think this uses the same approach as i.e. geth's `ReadSnapshotRoot` - return an empty hash in case of error. Should we leave it as is then? 🤔

core/rawdb/accessors_state_sync.go
qdm12: Thanks for the suggestion, bundled in 1ccca3b0bdaf5fc3516a95d60596678a3ee78738.
Note these are simply code movements of existing code, but why not change them whilst we're at it indeed 👍

core/rawdb/accessors_state_sync.go
qdm12: Thanks for the suggestion, bundled in 1ccca3b0bdaf5fc3516a95d60596678a3ee78738.
Note these are simply code movements of existing code, but why not change them whilst we're at it indeed 👍

core/rawdb/accessors_metadata_ext.go
qdm12: Thanks for the suggestion, bundled in 1ccca3b0bdaf5fc3516a95d60596678a3ee78738.
Note these are simply code movements of existing code, but why not change them whilst we're at it indeed 👍

core/rawdb/accessors_metadata_ext.go
qdm12: Hmm interesting, I do like the `else if` form to save a line personally, but sure, I can change it 🤷 
Bundled in 1ccca3b0bdaf5fc3516a95d60596678a3ee78738.
Note these are simply code movements of existing code, but why not change them whilst we're at it indeed 👍

core/rawdb/accessors_metadata_ext.go
qdm12: Bundled in 1ccca3b0bdaf5fc3516a95d60596678a3ee78738.
Note these are simply code movements of existing code, but why not change them whilst we're at it indeed 👍

core/rawdb/accessors_snapshot_ext.go
qdm12: Bundled in 1ccca3b0bdaf5fc3516a95d60596678a3ee78738.
Note these are simply code movements of existing code, but why not change them whilst we're at it indeed 👍

ARR4N (approved):

darioush (approved):

PR #774, opened 2025-01-30
Cleanup dynamic fee windower logic

StephenButtolph:
## Why this should be merged

The existing windowing logic is overly complex.

1. The format of the window is baked into `CalcBaseFee`.
2. The implementation supports arbitrary sized window elements even though we never use this. Specifically how we have `rollWindow` and `rollLongWindow` even though all we ever use are `uint64`s.

## How this works

Moves all the windower logic into a new type `DynamicFeeWindow`

## How this was tested

Expanded unit tests (and existing dynamic fees tests)

## Need to be documented?

no.

no.

### DIFF
--- a/consensus/dummy/dynamic_fee_window.go
+++ b/consensus/dummy/dynamic_fee_window.go
@@ -15,6 +15,11 @@ import (
 
 var ErrInsufficientDynamicFeeWindowLength = errors.New("insufficient length for dynamic fee window")
 
+// DynamicFeeWindow is a window of the last [params.RollupWindow] seconds of gas
+// usage.
+//
+// Index 0 is the oldest entry, and [params.RollupWindow-1] is the current
+// entry.
 type DynamicFeeWindow [params.RollupWindow]uint64
 
 func ParseDynamicFeeWindow(bytes []byte) (DynamicFeeWindow, error) {
--- a/consensus/dummy/dynamic_fee_window.go
+++ b/consensus/dummy/dynamic_fee_window.go
@@ -1,0 +1,87 @@
+// (c) 2019-2020, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package dummy
+
+import (
+	"encoding/binary"
+	"errors"
+	"fmt"
+
+	"github.com/ava-labs/avalanchego/utils/wrappers"
+	"github.com/ava-labs/coreth/params"
+	"github.com/ethereum/go-ethereum/common/math"
+)
+
+var ErrInsufficientDynamicFeeWindowLength = errors.New("insufficient length for dynamic fee window")
+
+type DynamicFeeWindow [params.RollupWindow]uint64
+
+func ParseDynamicFeeWindow(bytes []byte) (DynamicFeeWindow, error) {
+	if len(bytes) < params.DynamicFeeExtraDataSize {
+		return DynamicFeeWindow{}, fmt.Errorf("%w expected at least %d, but found %d",
+			ErrInsufficientDynamicFeeWindowLength,
+			params.DynamicFeeExtraDataSize,
+			len(bytes),
+		)
+	}
+
+	var window DynamicFeeWindow
+	for i := range window {
+		offset := i * wrappers.LongLen
+		window[i] = binary.BigEndian.Uint64(bytes[offset:])
+	}
+	return window, nil
+}
+
+// Add adds the amount to the most recent entry in the window.
+//
+// If the most recent entry overflows, it is set to [math.MaxUint64].
+func (w *DynamicFeeWindow) Add(amount uint64) {
+	const lastIndex uint = params.RollupWindow - 1
+	(*w)[lastIndex] = add(w[lastIndex], amount)
+}
+
+// Shift removes the oldest amount entries from the window and adds amount new
+// empty entries.
+func (w *DynamicFeeWindow) Shift(amount uint64) {
+	if amount >= params.RollupWindow {
+		*w = DynamicFeeWindow{}
+		return
+	}
+
+	var newWindow DynamicFeeWindow
+	copy(newWindow[:], w[amount:])
+	*w = newWindow
+}
+
+// Sum returns the sum of all the entries in the window.
+//
+// If the sum overflows, [math.MaxUint64] is returned.
+func (w *DynamicFeeWindow) Sum() uint64 {
+	var sum uint64
+	for _, v := range w {
+		sum = add(sum, v)
+	}
+	return sum
+}
+
+func (w *DynamicFeeWindow) Bytes() []byte {
+	bytes := make([]byte, len(w)*wrappers.LongLen)
+	for i, v := range w {
+		offset := i * wrappers.LongLen
+		binary.BigEndian.PutUint64(bytes[offset:], v)
+	}
+	return bytes
+}
+
+func add(sum uint64, values ...uint64) uint64 {
+	var overflow bool
+	for _, v := range values {
+		sum, overflow = math.SafeAdd(sum, v)
+		if overflow {
+			return math.MaxUint64
+		}
+	}
+	return sum
+}
--- a/consensus/dummy/dynamic_fee_window.go
+++ b/consensus/dummy/dynamic_fee_window.go
@@ -47,16 +47,16 @@ func (w *DynamicFeeWindow) Add(amounts ...uint64) {
 	w[lastIndex] = add(w[lastIndex], amounts...)
 }
 
-// Shift removes the oldest amount entries from the window and adds amount new
-// empty entries.
-func (w *DynamicFeeWindow) Shift(amount uint64) {
-	if amount >= params.RollupWindow {
+// Shift removes the oldest n entries from the window and adds n new empty
+// entries.
+func (w *DynamicFeeWindow) Shift(n uint64) {
+	if n >= params.RollupWindow {
 		*w = DynamicFeeWindow{}
 		return
 	}
 
 	var newWindow DynamicFeeWindow
-	copy(newWindow[:], w[amount:])
+	copy(newWindow[:], w[n:])
 	*w = newWindow
 }
 
--- a/consensus/dummy/dynamic_fee_window.go
+++ b/consensus/dummy/dynamic_fee_window.go
@@ -18,7 +18,7 @@ var ErrInsufficientDynamicFeeWindowLength = errors.New("insufficient length for 
 // DynamicFeeWindow is a window of the last [params.RollupWindow] seconds of gas
 // usage.
 //
-// Index 0 is the oldest entry, and [params.RollupWindow-1] is the current
+// Index 0 is the oldest entry, and [params.RollupWindow]-1 is the current
 // entry.
 type DynamicFeeWindow [params.RollupWindow]uint64
 
--- a/consensus/dummy/dynamic_fee_window.go
+++ b/consensus/dummy/dynamic_fee_window.go
@@ -72,7 +72,7 @@ func (w *DynamicFeeWindow) Sum() uint64 {
 }
 
 func (w *DynamicFeeWindow) Bytes() []byte {
-	bytes := make([]byte, len(w)*wrappers.LongLen)
+	bytes := make([]byte, params.DynamicFeeExtraDataSize)
 	for i, v := range w {
 		offset := i * wrappers.LongLen
 		binary.BigEndian.PutUint64(bytes[offset:], v)
--- a/consensus/dummy/dynamic_fee_window.go
+++ b/consensus/dummy/dynamic_fee_window.go
@@ -1,4 +1,4 @@
-// (c) 2019-2020, Ava Labs, Inc. All rights reserved.
+// (c) 2019-2025, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
 package dummy
--- a/consensus/dummy/dynamic_fee_window.go
+++ b/consensus/dummy/dynamic_fee_window.go
@@ -39,12 +39,12 @@ func ParseDynamicFeeWindow(bytes []byte) (DynamicFeeWindow, error) {
 	return window, nil
 }
 
-// Add adds the amount to the most recent entry in the window.
+// Add adds the amounts to the most recent entry in the window.
 //
 // If the most recent entry overflows, it is set to [math.MaxUint64].
-func (w *DynamicFeeWindow) Add(amount uint64) {
+func (w *DynamicFeeWindow) Add(amounts ...uint64) {
 	const lastIndex uint = params.RollupWindow - 1
-	(*w)[lastIndex] = add(w[lastIndex], amount)
+	(*w)[lastIndex] = add(w[lastIndex], amounts...)
 }
 
 // Shift removes the oldest amount entries from the window and adds amount new
--- a/consensus/dummy/dynamic_fee_window.go
+++ b/consensus/dummy/dynamic_fee_window.go
@@ -1,4 +1,4 @@
-// (c) 2019-2025, Ava Labs, Inc. All rights reserved.
+// (c) 2025, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
 package dummy
@@ -13,7 +13,7 @@ import (
 	"github.com/ethereum/go-ethereum/common/math"
 )
 
-var ErrInsufficientDynamicFeeWindowLength = errors.New("insufficient length for dynamic fee window")
+var ErrDynamicFeeWindowInsufficientLength = errors.New("insufficient length for dynamic fee window")
 
 // DynamicFeeWindow is a window of the last [params.RollupWindow] seconds of gas
 // usage.
@@ -24,8 +24,8 @@ type DynamicFeeWindow [params.RollupWindow]uint64
 
 func ParseDynamicFeeWindow(bytes []byte) (DynamicFeeWindow, error) {
 	if len(bytes) < params.DynamicFeeExtraDataSize {
-		return DynamicFeeWindow{}, fmt.Errorf("%w expected at least %d, but found %d",
-			ErrInsufficientDynamicFeeWindowLength,
+		return DynamicFeeWindow{}, fmt.Errorf("%w: expected at least %d bytes but got %d bytes",
+			ErrDynamicFeeWindowInsufficientLength,
 			params.DynamicFeeExtraDataSize,
 			len(bytes),
 		)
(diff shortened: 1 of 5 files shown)


consensus/dummy/dynamic_fees.go
```diff
@@ -83,65 +69,56 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header, timestamp uin
...
 
-		// Only add the [blockGasCost] to the gas used if it isn't AP5
-		if !isApricotPhase5 {
-			addedGas, overflow = math.SafeAdd(addedGas, blockGasCost)
-			if overflow {
-				addedGas = math.MaxUint64
-			}
-		}
+	// Compute the new state of the gas rolling window.
+	addedGas := add(parent.GasUsed, parentExtraStateGasUsed, blockGasCost)
```
StephenButtolph, line 105: This isn't a change in behavior, but it feels odd to me that we added the `blockGasCost` to the parent's `addedGas` rather than the current blocks.

consensus/dummy/dynamic_fees.go
```diff
@@ -83,65 +69,56 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header, timestamp uin
...
+	dynamicFeeWindow.Add(addedGas)
 
-		slot := params.RollupWindow - 1 - roll
-		start := slot * wrappers.LongLen
-		updateLongWindow(newRollupWindow, start, addedGas)
```
StephenButtolph, line 137: This slot management and surrounding `if` was only needed because we decided to shift at the start of the function rather than later in the function. By moving the shift, we can avoid all these checks.

consensus/dummy/dynamic_fees.go
```diff
@@ -214,76 +191,6 @@ func selectBigWithinBounds(lowerBound, value, upperBound *big.Int) *big.Int {
...
-// [1, 2, 3, 4]
-// Roll = 1
-// [2, 3, 4, 0]
-// Roll = 2
-// [3, 4, 0, 0]
-// Roll = 3
-// [4, 0, 0, 0]
-// Roll >= 4
-// [0, 0, 0, 0]
-// Assumes that [roll] is greater than or equal to 0
```
StephenButtolph, line 232: The new code uses a `uint64` to avoid any assumptions (which actually flows more naturally anyways, because we actually cast from a `uint64 -> int` and then assume the `int >= 0` rather than just using the `uint64`.

consensus/dummy/dynamic_fees.go
```diff
@@ -214,76 +191,6 @@ func selectBigWithinBounds(lowerBound, value, upperBound *big.Int) *big.Int {
...
-// If an overflow occurs, while summing the contents, the maximum uint64 value is returned.
-func sumLongWindow(window []byte, numLongs int) uint64 {
-	var (
-		sum      uint64 = 0
-		overflow bool
-	)
-	for i := 0; i < numLongs; i++ {
-		// If an overflow occurs while summing the elements of the window, return the maximum
-		// uint64 value immediately.
-		sum, overflow = math.SafeAdd(sum, binary.BigEndian.Uint64(window[wrappers.LongLen*i:]))
```
StephenButtolph, line 266: It seems like there was a decent amount of work to avoid parsing the window into a `[]uint64`, but that's actually what we are doing here anyways... So assuming overflow is infrequent (which it should never happen in reality) just using a `[...]uint64` should be just as efficient but avoid the `[]byte` handling.

consensus/dummy/dynamic_fees.go
```diff
@@ -83,65 +69,56 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header, timestamp uin
...
+		// may not have a populated [ExtDataGasUsed].
+		if parent.ExtDataGasUsed != nil {
+			parentExtraStateGasUsed = parent.ExtDataGasUsed.Uint64()
 		}
+	default:
+		blockGasCost = ApricotPhase3BlockGasFee
+	}
 
-		// Only add the [blockGasCost] to the gas used if it isn't AP5
-		if !isApricotPhase5 {
```
StephenButtolph, line 128: This check isn't needed, as `blockGasCost` is `0` if `isApricotPhase5` is true.

consensus/dummy/dynamic_fee_window.go
```diff
@@ -0,0 +1,92 @@
...
+		window[i] = binary.BigEndian.Uint64(bytes[offset:])
+	}
+	return window, nil
+}
+
+// Add adds the amount to the most recent entry in the window.
+//
+// If the most recent entry overflows, it is set to [math.MaxUint64].
+func (w *DynamicFeeWindow) Add(amount uint64) {
+	const lastIndex uint = params.RollupWindow - 1
```
StephenButtolph, line 46: This was something fun I discovered working on the canoto stuff, if `params.RollupWindow` is modified to make this code panic, by defining it as a `const ... uint`, this code actually won't compile.

Not really important for this code tbh... But I like it 🤷

consensus/dummy/dynamic_fee_window_test.go
```diff
@@ -0,0 +1,246 @@
...
+
+func TestDynamicFeeWindow_Add(t *testing.T) {
+	tests := []struct {
+		name     string
+		window   DynamicFeeWindow
+		amount   uint64
+		expected DynamicFeeWindow
+	}{
+		{
+			name: "normal addition",
```
darioush, line 23: nit: seems preferable to use `_` in the names passed to t.Run, since spaces are replaced with `_`, then we can search any failing test by string in the code.

darioush (approved):
nice cleanup

consensus/dummy/dynamic_fee_window.go
```diff
@@ -0,0 +1,92 @@
+// (c) 2019-2025, Ava Labs, Inc. All rights reserved.
```
qdm12, line 1: nit shouldn't this be:

```suggestion
// (c) 2025, Ava Labs, Inc. All rights reserved.
```

since it's a new file

consensus/dummy/dynamic_fee_window.go
```diff
@@ -0,0 +1,92 @@
...
+	"encoding/binary"
+	"errors"
+	"fmt"
+
+	"github.com/ava-labs/avalanchego/utils/wrappers"
+	"github.com/ava-labs/coreth/params"
+	"github.com/ethereum/go-ethereum/common/math"
+)
+
+var ErrInsufficientDynamicFeeWindowLength = errors.New("insufficient length for dynamic fee window")
```
qdm12, line 16: nit starting the err name with `ErrDynamicFeeWindow` would be better when searching for errors related to the dynamic fee window, so you could rename this to `ErrDynamicFeeWindowInsufficientLength`

consensus/dummy/dynamic_fee_window.go
```diff
@@ -0,0 +1,92 @@
...
+func ParseDynamicFeeWindow(bytes []byte) (DynamicFeeWindow, error) {
+	if len(bytes) < params.DynamicFeeExtraDataSize {
+		return DynamicFeeWindow{}, fmt.Errorf("%w expected at least %d, but found %d",
+			ErrInsufficientDynamicFeeWindowLength,
+			params.DynamicFeeExtraDataSize,
+			len(bytes),
+		)
+	}
+
+	var window DynamicFeeWindow
```
qdm12, line 34: supernit you could save a line if using named returns

```suggestion
func ParseDynamicFeeWindow(bytes []byte) (window DynamicFeeWindow, err error) {
	if len(bytes) < params.DynamicFeeExtraDataSize {
		return DynamicFeeWindow{}, fmt.Errorf("%w expected at least %d, but found %d",
			ErrInsufficientDynamicFeeWindowLength,
			params.DynamicFeeExtraDataSize,
			len(bytes),
		)
	}

```

consensus/dummy/dynamic_fee_window.go
```diff
@@ -0,0 +1,92 @@
...
+// DynamicFeeWindow is a window of the last [params.RollupWindow] seconds of gas
+// usage.
+//
+// Index 0 is the oldest entry, and [params.RollupWindow]-1 is the current
+// entry.
+type DynamicFeeWindow [params.RollupWindow]uint64
+
+func ParseDynamicFeeWindow(bytes []byte) (DynamicFeeWindow, error) {
+	if len(bytes) < params.DynamicFeeExtraDataSize {
+		return DynamicFeeWindow{}, fmt.Errorf("%w expected at least %d, but found %d",
```
qdm12, line 27: nit just for clarity

```suggestion
		return DynamicFeeWindow{}, fmt.Errorf("%w: expected at least %d bytes but got %d bytes",
```

consensus/dummy/dynamic_fee_window_test.go
```diff
@@ -0,0 +1,246 @@
...
+	"strconv"
+	"testing"
+
+	"github.com/ava-labs/coreth/params"
+	"github.com/ethereum/go-ethereum/common/math"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDynamicFeeWindow_Add(t *testing.T) {
+	tests := []struct {
```
qdm12, line 16: supernit you could use `map[string]struct{...}` to have the subtest name as the map key - it formats nicely imo the test cases. The only drawback (or not) is that subtests run without order.

consensus/dummy/dynamic_fee_window_test.go
```diff
@@ -0,0 +1,246 @@
+// (c) 2019-2025, Ava Labs, Inc. All rights reserved.
```
qdm12, line 1:
```suggestion
// (c) 2025, Ava Labs, Inc. All rights reserved.
```

consensus/dummy/dynamic_fees.go
```diff
@@ -83,65 +69,55 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header, timestamp uin
...
-			}
-		}
+	// Compute the new state of the gas rolling window.
+	dynamicFeeWindow.Add(parent.GasUsed, parentExtraStateGasUsed, blockGasCost)
 
-		slot := params.RollupWindow - 1 - roll
-		start := slot * wrappers.LongLen
-		updateLongWindow(newRollupWindow, start, addedGas)
+	if timestamp < parent.Time {
+		return nil, nil, fmt.Errorf("cannot calculate base fee for timestamp (%d) prior to parent timestamp (%d)", timestamp, parent.Time)
```
qdm12, line 108: nit why do we use parentheses for the timestamp numbers?

consensus/dummy/dynamic_fees.go
```diff
@@ -83,65 +69,55 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header, timestamp uin
...
-				ApricotPhase4MaxBlockGasCost,
-				ApricotPhase4BlockGasCostStep,
-				parent.BlockGasCost,
-				parent.Time, timestamp,
-			).Uint64()
+	// Add in parent's consumed gas
+	var blockGasCost, parentExtraStateGasUsed uint64
+	switch {
+	case isApricotPhase5:
+		// [blockGasCost] has been removed in AP5, so it is left as 0.
```
qdm12, line 76: brackets won't work for local variables

```suggestion
		// `blockGasCost` has been removed in AP5, so it is left as 0.
```

consensus/dummy/dynamic_fees.go
```diff
@@ -83,65 +69,55 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header, timestamp uin
...
+			parentExtraStateGasUsed = parent.ExtDataGasUsed.Uint64()
 		}
-
-		// Compute the new state of the gas rolling window.
-		addedGas, overflow := math.SafeAdd(parent.GasUsed, parentExtraStateGasUsed)
-		if overflow {
-			addedGas = math.MaxUint64
+	case isApricotPhase4:
+		// The [blockGasCost] is paid by the effective tips in the block using
+		// the block's value of [baseFee].
```
qdm12, line 85: brackets won't work for local variables

```suggestion
		// The `blockGasCost` is paid by the effective tips in the block using
		// the block's value of `baseFee`.
```

consensus/dummy/dynamic_fees.go
```diff
@@ -83,65 +69,55 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header, timestamp uin
...
 
-			// On the boundary of AP3 and AP4 or at the start of a new network, the parent
-			// may not have a populated [ExtDataGasUsed].
-			if parent.ExtDataGasUsed != nil {
-				parentExtraStateGasUsed = parent.ExtDataGasUsed.Uint64()
-			}
-		default:
-			blockGasCost = ApricotPhase3BlockGasFee
+		// At the start of a new network, the parent
+		// may not have a populated [ExtDataGasUsed].
```
qdm12, line 79: nit Those brackets won't work since it's not a global symbol, prefer using backticks

```suggestion
		// may not have a populated `ExtDataGasUsed`.
```

consensus/dummy/dynamic_fees.go
```diff
@@ -83,65 +69,55 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header, timestamp uin
...
+			ApricotPhase4TargetBlockRate,
+			ApricotPhase4MinBlockGasCost,
+			ApricotPhase4MaxBlockGasCost,
+			ApricotPhase4BlockGasCostStep,
+			parent.BlockGasCost,
+			parent.Time, timestamp,
+		).Uint64()
+
+		// On the boundary of AP3 and AP4 or at the start of a new network, the parent
+		// may not have a populated [ExtDataGasUsed].
```
qdm12, line 96: nit Those brackets won't work since it's not a global symbol, prefer using backticks

```suggestion
		// may not have a populated `ExtDataGasUsed`.
```

qdm12 (dismissed):

consensus/dummy/dynamic_fees.go
StephenButtolph: 🤷 that's just how it was. I removed the parens.

consensus/dummy/dynamic_fee_window_test.go
StephenButtolph: I generally find it easier to debug / work with a test where the order is consistent... So I always use the struct version rather than the map approach... But If people feel strongly I can switch to this.

consensus/dummy/dynamic_fee_window.go
StephenButtolph: I typically avoid using named returns other than for exceptional cases... It doesn't seem like named returns are very common in coreth, so I think it makes sense to keep the line rather than using named returns.

consensus/dummy/dynamic_fee_window_test.go
darioush: I also prefer the consistent order of arrays here

darioush (dismissed):

consensus/dummy/dynamic_fee_window.go
```diff
@@ -0,0 +1,88 @@
...
+		return DynamicFeeWindow{}, fmt.Errorf("%w: expected at least %d bytes but got %d bytes",
+			ErrDynamicFeeWindowInsufficientLength,
+			params.DynamicFeeExtraDataSize,
+			len(bytes),
+		)
+	}
+
+	var window DynamicFeeWindow
+	for i := range window {
+		offset := i * wrappers.LongLen
```
ARR4N, line 36: (subjective) I think `wrappers.LongLen` adds very little while requiring the reader to check its value. We're extracting a `uint64` from bytes, so a raw `8` is an acceptable magic number IMO.

consensus/dummy/dynamic_fee_window.go
```diff
@@ -0,0 +1,88 @@
...
+
+// Add adds the amounts to the most recent entry in the window.
+//
+// If the most recent entry overflows, it is set to [math.MaxUint64].
+func (w *DynamicFeeWindow) Add(amounts ...uint64) {
+	const lastIndex uint = params.RollupWindow - 1
+	w[lastIndex] = add(w[lastIndex], amounts...)
+}
+
+// Shift removes the oldest amount entries from the window and adds amount new
```
ARR4N, line 50:
```suggestion
// Shift removes the oldest `amount` entries from the window and adds `amount` new
```

Or rename it to `n` (it's a short function so that's fine, and `n` is self-explanatory). Being an English word it was very confusing to read "removes the oldest amount" (valid sentence thus far) only to then realise that it's a variable.

https://en.wikipedia.org/wiki/Garden-path_sentence

consensus/dummy/dynamic_fee_window.go
```diff
@@ -0,0 +1,88 @@
...
+//
+// If the sum overflows, [math.MaxUint64] is returned.
+func (w *DynamicFeeWindow) Sum() uint64 {
+	return add(0, w[:]...)
+}
+
+func (w *DynamicFeeWindow) Bytes() []byte {
+	bytes := make([]byte, params.DynamicFeeExtraDataSize)
+	for i, v := range w {
+		offset := i * wrappers.LongLen
```
ARR4N, line 73: As above re `wrappers.LongLen`.

consensus/dummy/dynamic_fee_window_test.go
```diff
@@ -0,0 +1,246 @@
...
+			},
+		},
+		{
+			amount: 9,
+			expected: DynamicFeeWindow{
+				10,
+			},
+		},
+		{
+			amount:   params.RollupWindow,
```
ARR4N, line 128: I'd use a raw 10 here (or as another test). What's important is that this is shifting by one more than the last test (i.e. the edge case), not that it's by the actual value of `params.RollupWindow`, even though they're the same thing.

consensus/dummy/dynamic_fee_window_test.go
```diff
@@ -0,0 +1,246 @@
...
+		},
+		{
+			name:   "zero_window",
+			bytes:  make([]byte, params.DynamicFeeExtraDataSize),
+			window: DynamicFeeWindow{},
+		},
+		{
+			name: "truncate_bytes",
+			bytes: []byte{
+				params.DynamicFeeExtraDataSize: 1,
```
ARR4N, line 198: What is this sorcery?!!!

TIL about this syntax for slice literals 🤯

ARR4N (approved):
Readability LGTM

consensus/dummy/dynamic_fee_window.go
StephenButtolph: Hmmm I feel we should keep using `wrappers.LongLen` because that aligns with the definition of:
```golang
const DynamicFeeExtraDataSize = RollupWindow * wrappers.LongLen
```

PR #775, opened 2025-01-30
Implement ACP-176

StephenButtolph:


StephenButtolph (commented):
Will need to look into the mempool filtering to make sure that having a block with a low capacity doesn't throw out the txs from the mempool

miner/worker.go
```diff
@@ -199,7 +190,7 @@ func (w *worker) commitNewWork(predicateContext *precompileconfig.PredicateConte
 		return nil, fmt.Errorf("failed to create new current environment: %w", err)
 	}
 	if header.ParentBeaconRoot != nil {
-		context := core.NewEVMBlockContext(header, w.chain, nil)
+		context := core.NewEVMBlockContextWithPredicateResults(header, w.chain, nil, predicate.NewResults())
```
StephenButtolph, line 193: TODO: Need to document this better

StephenButtolph:
See #815

PR #777, opened 2025-01-31
[WIP] Use libevm instead of go-ethereum

michaelkaplan13:
## Why this should be merged
`libevm` is fork of `go-ethereum` that is currently at the same upstream version as we use here (`v1.13.14`). In case should want or need to back-port upstream changes from `go-ethereum` in the future without needing to fully update to a newer upstream version, we should import the `libevm` packages and remove the direct dependency on `go-ethereum` entirely.

## How this works
Self-explanatory

## How this was tested
CI

## Need to be documented in RELEASES.md?
No


qdm12:
I think this should be closed once #820 gets merged, given go-ethereum is no longer imported anywhere in the libevm branch.

PR #783, opened 2025-02-07
Implement ACP-176 math

StephenButtolph:
## Why this should be merged

This PR implements all of the formulas specified by [ACP-176](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md).

## How this works

Adds the implementation and extensive tests.

## How this was tested

1. Unit tests
2. Utilized this in a local network to handle fees.

## Need to be documented?

No.

No.


plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,196 @@
...
+	return gas.Gas(sort.Search(maxTargetExcess, func(targetExcessGuess int) bool {
+		state := State{
+			TargetExcess: gas.Gas(targetExcessGuess),
+		}
+		return state.Target() >= desiredTarget
+	}))
+}
+
+// fakeExponential approximates factor * e ** (numerator / denominator) using
+// Taylor expansion.
```
darioush, line 152:
```suggestion
// Taylor expansion:
// e^x = 1 + x/1! + x^2/2! + x^3/3! + ...

```

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,196 @@
...
+	change = min(change, MaxTargetExcessDiff)
+	if excess < desired {
+		return excess + change
+	}
+	return excess - change
+}
+
+// modifyExcess scales the excess during gas target modifications to keep the
+// price constant.
+func modifyExcess(
```
darioush, line 181: nit: maybe scaleExcess?

plugin/evm/acp176/acp176.go
StephenButtolph: This is copied from the EIP-4844 code... But that code doesn't export it... So I think we need to copy it (unless we want to modify it to be exported in their code).

plugin/evm/acp176/acp176_test.go
```diff
@@ -0,0 +1,694 @@
...
+		},
+		{
+			name: "max_initial_excess_change",
+			state: State{
+				Gas: gas.State{
+					Excess: 47_286_485, // (MinTargetPerSecond + 977) * ln(3) * TargetToPriceUpdateConversion
+				},
+				TargetExcess: MaxTargetExcessDiff,
+			},
+			skipTestDesiredTargetExcess: true,
```
darioush, line 72: why is this set?

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		priceUpdateConversion = math.MaxUint64
+	}
+
+	bigExcess := new(big.Int).SetUint64(uint64(s.Gas.Excess))
+	bigPriceUpdateFraction := new(big.Int).SetUint64(uint64(priceUpdateConversion))
+	return fakeExponential(minBaseFee, bigExcess, bigPriceUpdateFraction)
+}
+
+// AdvanceTime increases the gas capacity and decreases the gas excess based on
```
darioush, line 75: not sure why this is called Fraction? maybe Denominator?

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,193 @@
...
+func (s *State) MaxCapacity() gas.Gas {
+	targetPerSecond := s.Target()
+	maxCapacity, err := safemath.Mul(targetToMaxCapacity, targetPerSecond)
+	if err != nil {
+		maxCapacity = math.MaxUint64
+	}
+	return maxCapacity
+}
+
+// BaseFee returns the current required fee per gas.
```
darioush, line 62: Could we keep that 1st line of comment?

```suggestion
// BaseFee returns the current required fee per gas.
//
// p = M * e ** (excess / (TargetToPriceUpdateConversion * target))
```

darioush (dismissed):

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,193 @@
...
+func (s *State) BaseFee() *big.Int {
+	target := s.Target()
+	priceUpdateConversion, err := safemath.Mul(TargetToPriceUpdateConversion, target)
+	if err != nil {
+		priceUpdateConversion = math.MaxUint64
+	}
+
+	bigExcess := new(big.Int).SetUint64(uint64(s.Gas.Excess))
+	bigPriceUpdateFraction := new(big.Int).SetUint64(uint64(priceUpdateConversion))
```
darioush, line 71: not sure why this is called Fraction? maybe Denominator?

darioush (dismissed):

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,197 @@
...
+	)
+	for i := 1; accum.Sign() > 0; i++ {
+		output.Add(output, accum)
+
+		accum.Mul(accum, numerator)
+		accum.Div(accum, denominator)
+		accum.Div(accum, big.NewInt(int64(i)))
+	}
+	return output.Div(output, denominator)
+}
```
StephenButtolph, line 167: @ARR4N Specifically interested in your opinion on copying this from the eip-4844 package vs exporting the function in the eip-4844 package.

StephenButtolph: I suppose one alternative could be defining an exported version and just calling the unexported version.

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,197 @@
...
+	)
+}
+
+// DesiredTargetExcess calculates the optimal desiredTargetExcess given the
+// desired target.
+//
+// This could be solved directly by calculating D * ln(desiredTarget / P) using
+// floating point math. However, it introduces inaccuracies. So, we use a binary
+// search to find the closest integer solution.
+func DesiredTargetExcess(desiredTarget gas.Gas) gas.Gas {
```
michaelkaplan13, line 142: IIUC, this calculates $q_{desired}$ as the first value that puts the gas target above the nodes set desired gas target.

It likely doesn't matter either way, but should we err on the side of going with the last value that puts the gas target below their desired level to ensure it never recommends exceeding it? Or potentially going with the value of $q$ that puts the gas target closest to the configured desired level rather than always rounding up.

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,197 @@
...
+	targetToMaxCapacity = TargetToMax * TimeToFillCapacity
+	maxTargetExcess     = 1_024_950_627 // TargetConversion * ln(MaxUint64 / MinTargetPerSecond) + 1
+)
+
+var minBaseFee = big.NewInt(MinBaseFee)
+
+// State represents the current state of the gas pricing and constraints.
+type State struct {
+	Gas          gas.State
+	TargetExcess gas.Gas
```
michaelkaplan13, line 40: `TargetExcess` is the latest value of $q$ right? I think I'm missing something...could you explain the rationale for including "excess" in the naming convention here? It initially had me confused with the excess gas used over the current target.

michaelkaplan13 (approved):
One question on naming and one question on rounding, but otherwise LGTM.

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// ConsumeGas decreases the gas capacity and increases the gas excess by
+// gasUsed + extraGasUsed. If the gas capacity is insufficient, an error is
+// returned.
+func (s *State) ConsumeGas(
+	gasUsed uint64,
+	extraGasUsed *big.Int,
+) error {
+	var err error
```
qdm12, line 106: nit

```suggestion
) (err error) {
```

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,197 @@
...
+			TargetExcess: gas.Gas(targetExcessGuess),
+		}
+		return state.Target() >= desiredTarget
+	}))
+}
+
+// fakeExponential approximates factor * e ** (numerator / denominator) using
+// Taylor expansion:
+// e^x = 1 + x/1! + x^2/2! + x^3/3! + ...
+func fakeExponential(factor, numerator, denominator *big.Int) *big.Int {
```
qdm12, line 154: nit: perhaps rename this to approximateExponential, since it's not totally fake either? 🤔

plugin/evm/acp176/acp176_test.go
```diff
@@ -0,0 +1,694 @@
...
+				Gas: gas.State{
+					Excess: 1_999_999_999, // inverse of max_decrease -1 due to rounding error
+				},
+				TargetExcess: 2 * MaxTargetExcessDiff,
+			},
+		},
+	}
+)
+
+func TestTarget(t *testing.T) {
```
qdm12, line 554: `TestTarget`, `TestMaxCapacity`, `TestBaseFee` and `TestDesiredTargetExcess` could be merged into one test I think. I will also say sharing a global variable of tests is a risky approach in case one test modifies one test case (now or in the future) which would influence other tests.

plugin/evm/acp176/acp176_test.go
```diff
@@ -0,0 +1,694 @@
+// (c) 2019-2025, Ava Labs, Inc. All rights reserved.
```
qdm12, line 1:
```suggestion
// (c) 2025, Ava Labs, Inc. All rights reserved.
```

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,197 @@
...
+	))
+}
+
+// MaxCapacity returns the maximum possible accrued gas capacity.
+func (s *State) MaxCapacity() gas.Gas {
+	targetPerSecond := s.Target()
+	maxCapacity, err := safemath.Mul(targetToMaxCapacity, targetPerSecond)
+	if err != nil {
+		maxCapacity = math.MaxUint64
+	}
```
qdm12, line 60: (maybe dumb question) What's the reasoning behind defaulting the max capacity to the max if the multiplication fails? Same question for BaseFee, AdvanceTime

plugin/evm/acp176/acp176_test.go
```diff
@@ -0,0 +1,694 @@
...
+
+func TestTarget(t *testing.T) {
+	for _, test := range readerTests {
+		t.Run(test.name, func(t *testing.T) {
+			require.Equal(t, test.target, test.state.Target())
+		})
+	}
+}
+
+func BenchmarkTarget(b *testing.B) {
```
qdm12, line 562: What's the objective of these benchmarks? 👀 🤔 
Are we comparing i.e. two implementations?

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,197 @@
...
+	return maxCapacity
+}
+
+// BaseFee returns the current required fee per gas.
+//
+// BaseFee = MinBaseFee * e^(Excess / (Target() * TargetToPriceUpdateConversion))
+func (s *State) BaseFee() *big.Int {
+	target := s.Target()
+	priceUpdateConversion, err := safemath.Mul(TargetToPriceUpdateConversion, target)
```
ARR4N, line 69:
```suggestion
	priceUpdateConversion, err := safemath.Mul(TargetToPriceUpdateConversion, target) // K
```

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,197 @@
...
+type State struct {
+	Gas          gas.State
+	TargetExcess gas.Gas
+}
+
+// Target returns the target gas consumed per second.
+//
+// Target = MinTargetPerSecond * e^(TargetExcess / TargetConversion)
+func (s *State) Target() gas.Gas {
+	return gas.Gas(gas.CalculatePrice(
```
ARR4N, line 47: While it's convenient that the equation for `T` has the same form as the equation for the gas price, it's detrimental to readability to conflate the two. To me it suggests that a common abstraction should exist (something capturing "exponential" and a non-negative exponent). This would also address the confusion that @michaelkaplan13 and I had re the word "excess" in `TargetExcess`.

For `X = A·exp(B/C)` we'd then need names for `A`, `B`, and `C`.

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,197 @@
...
+	bigExcess := new(big.Int).SetUint64(uint64(s.Gas.Excess))
+	bigPriceUpdateConversion := new(big.Int).SetUint64(uint64(priceUpdateConversion))
+	return fakeExponential(minBaseFee, bigExcess, bigPriceUpdateConversion)
+}
+
+// AdvanceTime increases the gas capacity and decreases the gas excess based on
+// the elapsed seconds.
+func (s *State) AdvanceTime(seconds uint64) {
+	targetPerSecond := s.Target()
+	maxPerSecond, err := safemath.Mul(TargetToMax, targetPerSecond)
```
ARR4N, line 83:
```suggestion
	maxPerSecond, err := safemath.Mul(TargetToMax, targetPerSecond) // R
```

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		maxPerSecond = math.MaxUint64
+	}
+	maxCapacity, err := safemath.Mul(TimeToFillCapacity, maxPerSecond)
```
ARR4N, line 87: Why is this different to the `MaxCapacity()` method? They both look like $C$ to me.

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,197 @@
...
+	if extraGasUsed != nil {
+		if !extraGasUsed.IsUint64() {
+			return fmt.Errorf("%w: extraGasUsed (%d) exceeds MaxUint64",
+				gas.ErrInsufficientCapacity,
+				extraGasUsed,
+			)
+		}
+		s.Gas, err = s.Gas.ConsumeGas(gas.Gas(extraGasUsed.Uint64()))
+	}
+	return err
```
ARR4N, line 120:
```suggestion
	if extraGasUsed == nil {
		return nil
	}
	if !extraGasUsed.IsUint64() {
		return fmt.Errorf("%w: extraGasUsed (%d) exceeds MaxUint64",
			gas.ErrInsufficientCapacity,
			extraGasUsed,
		)
	}
	s.Gas, err = s.Gas.ConsumeGas(gas.Gas(extraGasUsed.Uint64()))
	return err
```

Clearer if you reduce the nesting, especially the dangling `return err` that came from inside the `if`.

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,197 @@
...
+
+// ConsumeGas decreases the gas capacity and increases the gas excess by
+// gasUsed + extraGasUsed. If the gas capacity is insufficient, an error is
+// returned.
+func (s *State) ConsumeGas(
+	gasUsed uint64,
+	extraGasUsed *big.Int,
+) error {
+	var err error
+	s.Gas, err = s.Gas.ConsumeGas(gas.Gas(gasUsed))
```
ARR4N, line 107: My old readability mentor (then the head of el Goog readability team) does a much better job than me at explaining why he thinks [this is an antipattern](https://gist.github.com/kylelemons/f9ed27f211dc818abc9b06cc3ff91540).

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,197 @@
...
+func (s *State) BaseFee() *big.Int {
+	target := s.Target()
+	priceUpdateConversion, err := safemath.Mul(TargetToPriceUpdateConversion, target)
+	if err != nil {
+		priceUpdateConversion = math.MaxUint64
+	}
+
+	bigExcess := new(big.Int).SetUint64(uint64(s.Gas.Excess))
+	bigPriceUpdateConversion := new(big.Int).SetUint64(uint64(priceUpdateConversion))
+	return fakeExponential(minBaseFee, bigExcess, bigPriceUpdateConversion)
```
ARR4N, line 76: Isn't this what `gas.CalculateGas()` does? Why do we have another exponential implementation here?

plugin/evm/acp176/acp176.go
alarso16, line 87: Why not use `MaxCapacity()`?

plugin/evm/acp176/acp176.go
ARR4N: I think it's safer to copy (with attribution). 4844 is only the beginning of a lot of work they're doing with respect to sharding so who knows how things might change 🤷

plugin/evm/acp176/acp176.go
StephenButtolph: I felt like people would probably select round numbers. For example: `1.5M gps` or `2M gps`. If we go on the low side, then the actual targets could look like `1,499,998 gps` vs `1,500,001 gps` (or something). I felt like just visually, going slightly over looks nicer...

I don't think this actually matters as far as the protocol is concerned... Also we are able to perfectly calculate:
- 1,000,000
- 1,500,000
- 3,000,000
- 6,000,000
- 10,000,000
- 100,000,000

So for reasonable numbers the difference is nonexistent.

plugin/evm/acp176/acp176.go
StephenButtolph: This function is copied from geth, I don't think we should change the name.

plugin/evm/acp176/acp176.go
StephenButtolph: If the number overflows, it is just capped to maxuint64.

plugin/evm/acp176/acp176_test.go
StephenButtolph: The benchmarks are here to measure the performance of the package. Many of these functions are called during block processing and are therefore on a critical consensus path. They can't be slow.

plugin/evm/acp176/acp176.go
StephenButtolph: We need access to `maxPerSecond` later in this function and calling `Target()` is a non-negligible cost.

plugin/evm/acp176/acp176.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		maxCapacity = math.MaxUint64
+	}
+	return maxCapacity
+}
+
+// BaseFee returns the current required fee per gas.
+//
+// BaseFee = MinBaseFee * e^(Excess / (Target() * TargetToPriceUpdateConversion))
+func (s *State) BaseFee() *big.Int {
```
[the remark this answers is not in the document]
StephenButtolph: I agree that within this package is makes more sense. I do think it may make less sense externally... But we'll handle that as the future PRs come in.

plugin/evm/acp176/acp176.go
StephenButtolph: No longer required

plugin/evm/acp176/acp176.go
StephenButtolph: This is also generating `MaxCapacity` - it just felt wasteful to call `Target()` again (as that is much more expensive than just `safemath.Mul`.

plugin/evm/acp176/acp176.go
StephenButtolph: This made me re-think it. I think we're fine instituting a max gas price at 16 AVAX per gas 😂

StephenButtolph: Decided to remove the usage of `fakeExponential` - thanks!

plugin/evm/acp176/acp176.go
StephenButtolph: I'd prefer to just change the return type to be a `bool`

plugin/evm/acp176/acp176_test.go
StephenButtolph: > I don't think a future developer would know what to think of these benchmarks.

I think a future developer should be capable of looking at the performance difference a PR makes to determine if the change is acceptable or not.

plugin/evm/acp176/acp176_test.go
StephenButtolph: I don't have an opinion here. What would this test look like?
```golang
func TestEvaluate(t *testing.T) {
	for _, test := range readerTests {
		t.Run(test.name, func(t *testing.T) {
			assert := assert.New(t)

			assert.Equal(test.target, test.state.Target(), "incorrect target")
			assert.Equal(test.maxCapacity, test.state.MaxCapacity(), "incorrect maxCapacity")
			assert.Equal(test.gasPrice, test.state.GasPrice(), "incorrect gasPrice")
			if !test.skipTestDesiredTargetExcess {
				assert.Equal(test.state.TargetExcess, DesiredTargetExcess(test.target), "incorrect desiredTargetExcess")
			}
		})
	}
}
```
?

darioush (approved):

ARR4N (approved):
LGTM for readability (with a cursory nod to a review of correctness too, but there are many other eyes on this).

PR #786, opened 2025-02-11
simplify: remove blockManager type

darioush:
Seems we can inline the 2 fields of the blockManager to the block itself.

This is 1 less "manager" abstraction to maintain and understand. I didn't like this abstraction as I'm not sure what it represents. This link has a good writeup which I agree with https://softwareengineering.stackexchange.com/a/112444/144520

Previously the block already had access to the vm pointer, so it seems fine to add another pointer. It may be possible to simplify this a bit more (perhaps putting the responsibility of processing block hooks back on the VM or entirely wrapping the block type, which I can explore next), but I think this is a good start.

We can subsequently rename block_manager.go or re-distribute the methods of that file.

### DIFF
--- a/plugin/evm/block_manager.go
+++ b/plugin/evm/block_manager.go
@@ -15,7 +15,6 @@ import (
 	"github.com/ava-labs/coreth/constants"
 	"github.com/ava-labs/coreth/core/types"
 	"github.com/ava-labs/coreth/params"
-	"github.com/ava-labs/coreth/plugin/evm/extension"
 	"github.com/ava-labs/coreth/trie"
 )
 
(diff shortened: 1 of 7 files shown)


plugin/evm/block_manager.go
```diff
@@ -24,28 +23,17 @@ var (
...
-func newBlockManager(vm *VM, blockExtension extension.BlockManagerExtension) *blockManager {
-	return &blockManager{
-		blockExtension: blockExtension,
-		vm:             vm,
-	}
-}
-
 // newBlock returns a new Block wrapping the ethBlock type and implementing the snowman.Block interface
-func (bm *blockManager) newBlock(ethBlock *types.Block) (*Block, error) {
+func (vm *VM) newBlock(ethBlock *types.Block) (*Block, error) {
```
qdm12, line 27: nit: could we have this as `func newBlock(ethBlock *types.Block, vm *VM) (*Block, error)`? Not too sure why it would be a method on `VM`.

Also second nit, I would rename `ethBlock` to `block` or `b` given it doesn't conflict with a `block` I think, and is not really a eth-block (aka libevm block)

qdm12 (approved):
Nice work! 💯

plugin/evm/block_manager.go
ceyonur: I agree that we can pass vm as a param (in case we want to rather pass it as a interface).

Eth block is the eth block (it's core/types.Block), not sure what you meant there.

ceyonur (commented):
mostly lgtm, a small suggestion for newBlock (same with @qdm12's comment)

plugin/evm/block_manager.go
darioush: hum. I'm still not done here. appreciate the feedback though and will consider it.

PR #788, opened 2025-02-11
Refactor header extra verification

StephenButtolph:
## Why this should be merged

Removes duplicate code and uniformly enforces the extra data header lengths.

## How this works

Adds a helper and tests it.

## How this was tested

Unit test.

## Need to be documented?

No.

No.

### DIFF
--- a/consensus/dummy/consensus.go
+++ b/consensus/dummy/consensus.go
@@ -18,6 +18,8 @@ import (
 	"github.com/ava-labs/coreth/params"
 	"github.com/ava-labs/coreth/trie"
 	"github.com/ethereum/go-ethereum/common"
+
+	customheader "github.com/ava-labs/coreth/plugin/evm/header"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


plugin/evm/header/extra.go
```diff
@@ -0,0 +1,57 @@
...
+package header
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/ava-labs/coreth/params"
+)
+
+var ErrInvalidExtraLength = errors.New("invalid header.Extra length")
```
darioush, line 13: nit: not sure if this needs to be exported.

darioush (approved):
LGTM, assuming CI passes

consensus/dummy/consensus.go
```diff
@@ -212,25 +214,18 @@ func (eng *DummyEngine) verifyHeaderGasFields(config *params.ChainConfig, header
...
-	switch {
-	case config.IsDurango(header.Time):
-		if len(header.Extra) < params.DynamicFeeExtraDataSize {
-			return fmt.Errorf("expected extra-data field length >= %d, found %d", params.DynamicFeeExtraDataSize, len(header.Extra))
-		}
-	case config.IsApricotPhase3(header.Time):
-		if len(header.Extra) != params.DynamicFeeExtraDataSize {
-			return fmt.Errorf("expected extra-data field to be: %d, but found %d", params.DynamicFeeExtraDataSize, len(header.Extra))
-		}
-	default:
```
ceyonur, line 229: was the apricot phase 1 check missing here?

ceyonur (approved):
changes lgtm just a question with previously missing check

darioush (approved):

consensus/dummy/consensus.go
StephenButtolph: afaict - yes

PR #789, opened 2025-02-11
Implement `eth_suggestPriceOptions`

StephenButtolph:
## Why this should be merged

This API is intended to be used to suggest fees for user facing wallets.

## How this works

Slow = 95% of the recommended baseFee + tipCap (clamped to minimum + maximum values)
Normal = 100% of the recommended baseFee + tipCap (clamped to maximum values)
Fast = 105% of the recommended baseFee + tipCap

## How this was tested

This is currently live on `api.avax-test.network` and `api.avax.network` and is in the QA process for core to adopt.

## Need to be documented?

Perhaps?

Perhaps?


internal/ethapi/api.coreth.go
```diff
@@ -0,0 +1,118 @@
...
+
+	minBaseFee       = params.EtnaMinBaseFee // 1 nAVAX
+	maxNormalBaseFee = 100 * nAVAX
+
+	minGasTip       = 1 // 1 wei
+	maxNormalGasTip = 20 * nAVAX
+
+	slowFeeNum = 19 // 19/20 = 0.95
+	fastFeeNum = 21 // 21/20 = 1.05
+	feeDen     = 20
```
darioush, line 26: nit: I think it makes sense to use full Numerator / Denominator for these names.
slowFeeNum can bring "slowFeeNumber" to mind, even though it's kind of obvious.

internal/ethapi/api.coreth_test.go
```diff
@@ -0,0 +1,159 @@
...
+			estimateBaseFee:  common.Big1,
+			suggestGasTipCap: nil,
+			expected:         nil,
+		},
+		{
+			name:             "minimum_values",
+			estimateBaseFee:  bigMinBaseFee,
+			suggestGasTipCap: bigMinGasTip,
+			expected: &PriceOptions{
+				Slow: &Price{
```
darioush, line 55: nit: might be useful to make a helper function to create Price like `newPrice(gasTip, gasFee)` to avoid repeating `(*hexutil.Big)(big.NewInt(...))` in this test.
or just one for `(*hexutil.Big)(big.NewInt(...))` but I'm not sure what to call that

darioush (approved):

ceyonur (dismissed):

darioush (dismissed):

internal/ethapi/api.coreth.go
```diff
@@ -0,0 +1,118 @@
...
+
+	bigMinGasTip       = big.NewInt(minGasTip)
+	bigMaxNormalGasTip = big.NewInt(maxNormalGasTip)
+
+	bigSlowFeeNumerator = big.NewInt(slowFeeNumerator)
+	bigFastFeeNumerator = big.NewInt(fastFeeNumerator)
+	bigFeeDenominator   = big.NewInt(feeDenominator)
+)
+
+type Price struct {
```
qdm12, line 41: nit I'm not too sure what's meant to be exported or not given this is PR in a series of PRs I suppose 😉 
Maybe to avoid having exported and unused types (especially in the internal directory which isn't exposed to other repositories), have everything unexported and export-as-needed in subsequent PRs?

internal/ethapi/api.coreth.go
```diff
@@ -0,0 +1,118 @@
...
+	Normal *Price `json:"normal"`
+	Fast   *Price `json:"fast"`
+}
+
+// SuggestPriceOptions returns suggestions for what to display to a user for
+// current transaction fees.
+func (s *EthereumAPI) SuggestPriceOptions(ctx context.Context) (*PriceOptions, error) {
+	baseFee, err := s.b.EstimateBaseFee(ctx)
+	if err != nil || baseFee == nil {
+		return nil, err
```
qdm12, line 57: Is it ok to return `nil, nil` in case `baseFee` is `nil` and there is no error (if this can happen)?
Same question for the gasTip below

qdm12, line 57: I think it's safe to use error wrapping especially for new code:

```suggestion
		return nil,fmt.Errorf("failed to estimate base fee: %w", err)
```

internal/ethapi/api.coreth.go
```diff
@@ -0,0 +1,118 @@
...
+	cappedFee := math.BigMin(estimate, maxFee)
+
+	slowFee := new(big.Int).Set(cappedFee)
+	slowFee.Mul(slowFee, bigSlowFeeNumerator)
+	slowFee.Div(slowFee, bigFeeDenominator)
+	slowFee = math.BigMax(slowFee, minFee)
+
+	normalFee := cappedFee
+
+	fastFee := new(big.Int).Set(estimate)
```
qdm12, line 114: name the returns, that would help for quick documentation lookups:

```suggestion
) (slowFee, normalFee, fastFee *big.Int) {
	// Cap the fee to keep slow and normal options reasonable during fee spikes.
	cappedFee := math.BigMin(estimate, maxFee)

	slowFee = new(big.Int).Set(cappedFee)
	slowFee.Mul(slowFee, bigSlowFeeNumerator)
	slowFee.Div(slowFee, bigFeeDenominator)
	slowFee = math.BigMax(slowFee, minFee)

	normalFee = cappedFee

	fastFee = new(big.Int).Set(estimate)
```

internal/ethapi/api.coreth_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"math/big"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/stretchr/testify/require"
+)
+
+type testSuggestPriceOptionsBackend struct {
+	Backend // embeds the interface to avoid implementing unused methods
```
qdm12, line 17: supernit (aka feel free to discard this conversation) otherwise it sounds like `Backend` embeds, but it's just "we" or `testSuggestPriceOptionsBackend`

```suggestion
	Backend // embed the interface to avoid implementing unused methods
```

internal/ethapi/api.coreth_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"context"
+	"math/big"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/stretchr/testify/require"
+)
+
+type testSuggestPriceOptionsBackend struct {
```
qdm12, line 16: supernit perhaps rename it as a stub, i.e. `stubSuggestPriceOptionsBackend`

qdm12 (dismissed):
Nice work, just a few minor comments

internal/ethapi/api.coreth.go
StephenButtolph: I modified the code to return a struct.

internal/ethapi/api.coreth_test.go
```diff
@@ -0,0 +1,130 @@
...
+func (b *testSuggestPriceOptionsBackend) SuggestGasTipCap(context.Context) (*big.Int, error) {
+	return b.suggestGasTipCap, nil
+}
+
+func TestSuggestPriceOptions(t *testing.T) {
+	tests := []struct {
+		name             string
+		estimateBaseFee  *big.Int
+		suggestGasTipCap *big.Int
+		expected         *PriceOptions
```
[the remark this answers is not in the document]
StephenButtolph: ok

internal/ethapi/api.coreth_test.go
StephenButtolph: I think that `testSuggestPriceOptionsBackend` aligns more closely with the existing `testBackend`. (I know in avalanchego we would definitely call this a `testXXX` but not sure what the pattern is in coreth.

internal/ethapi/api.coreth.go
qdm12: Ah interesting; I think we should split this in two then. That if block did confuse me as well [previously](https://github.com/ava-labs/coreth/pull/789/files#r1952861528), so splitting it would be good for readability as well.

```go
if err != nil {
	return nil, fmt.Errorf("failed to estimate base fee: %w", err)
} else if baseFee == nil {
	return nil, nil
}
```

And let's error wrap as much as we can (within reasons, i.e. never in core/vm 😄) - nobody likes an error with no idea where it's coming from and what code path it took.

internal/ethapi/api.coreth.go
```diff
@@ -0,0 +1,128 @@
...
+// current transaction fees.
+func (s *EthereumAPI) SuggestPriceOptions(ctx context.Context) (*PriceOptions, error) {
+	baseFee, err := s.b.EstimateBaseFee(ctx)
+	if err != nil || baseFee == nil {
+		return nil, err
+	}
+	gasTip, err := s.b.SuggestGasTipCap(ctx)
+	if err != nil || gasTip == nil {
+		return nil, err
+	}
```
qdm12, line 62:
```suggestion
	if err != nil {
		return nil, fmt.Errorf("failed to suggest gas tip cap: %w", err)
	} else if  gasTip == nil {
		return nil, nil
	}		
```

darioush (approved):

ceyonur (approved):

PR #790, opened 2025-02-12
chore(core/rawdb): add `ExampleInspectDatabase` to check custom behavior (7)

qdm12:
This test is needed for later PR changing core/rawdb's `InspectDatabase` to use libevm's `InspectDatabase` with custom options.

This is shamelessly taken from https://github.com/ava-labs/libevm/blob/b2c38ce397a2da0881850cd925c34429b6519a97/core/rawdb/database.libevm_test.go#L47 but without using options.


core/rawdb/database_test.go
```diff
@@ -0,0 +1,136 @@
...
+	// | Key-Value store | Storage snapshot        | 0.00 B   |     0 |
+	// | Key-Value store | Clique snapshots        | 0.00 B   |     0 |
+	// | Key-Value store | Singleton metadata      | 28.00 B  |     2 |
+	// | Light client    | CHT trie nodes          | 0.00 B   |     0 |
+	// | Light client    | Bloom trie nodes        | 0.00 B   |     0 |
+	// | State sync      | Trie segments           | 78.00 B  |     1 |
+	// | State sync      | Storage tries to fetch  | 77.00 B  |     1 |
+	// | State sync      | Code to fetch           | 35.00 B  |     1 |
+	// | State sync      | Block numbers synced to | 23.00 B  |     1 |
+	// +-----------------+-------------------------+----------+-------+
```
ceyonur, line 71: should we convert this to a unit test and test these?

core/rawdb/database_test.go
```diff
@@ -0,0 +1,136 @@
...
+				{key: snapshotBlockHashKey, value: []byte("a")},
+				{key: syncRootKey, value: []byte("b")},
+				// Trie segments: 77 + 1 = 78 bytes
+				{key: extendRight(syncSegmentsPrefix, syncSegmentsKeyLength), value: []byte("c")},
+				// Storage tries to fetch: 76 + 1 = 77 bytes
+				{key: extendRight(syncStorageTriesPrefix, syncStorageTriesKeyLength), value: []byte("d")},
+				// Code to fetch: 34 + 1 = 35 bytes
+				{key: extendRight(CodeToFetchPrefix, codeToFetchKeyLength), value: []byte("e")},
+				// Block numbers synced to: 22 + 1 = 23 bytes
+				{key: extendRight(syncPerformedPrefix, syncPerformedKeyLength), value: []byte("f")},
```
ceyonur, line 32: hmm I wonder if we should directly use `rawdb` methods like `rawdb.AddCodeToFetch`, `rawdb.WriteSyncSegment` etc?

core/rawdb/database_test.go
qdm12: No given we don't want to modify the geth code directly, which only outputs to stdout with prints (not very testable).

I have however a geth branch I'll open to change InspectDatabase to accept a io.Writer + unit test (not example test), but that will need to be accepted upstream first.

core/rawdb/database_test.go
qdm12: Indeed, done 👍 !

core/rawdb/accessors_state_sync.go
```diff
@@ -104,11 +104,11 @@ func UnpackSyncSegmentKey(keyBytes []byte) (common.Hash, []byte) {
 }
 
 // packSyncSegmentKey packs root and account into a key for storage in db.
-func packSyncSegmentKey(root common.Hash, start []byte) []byte {
-	bytes := make([]byte, len(syncSegmentsPrefix)+common.HashLength+len(start))
+func packSyncSegmentKey(root common.Hash, start common.Hash) []byte {
```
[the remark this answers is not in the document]
ceyonur: hmm it seems L111 also assumes it must be 32 bytes:
```
copy(bytes[len(syncSegmentsPrefix)+common.HashLength:], start)
``` 
so this should be good.
nice catch!

ceyonur (approved):

core/rawdb/accessors_state_sync.go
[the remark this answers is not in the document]
qdm12: From checking the code it is in fact always 32 bytes. But better safe for future code than future-sorry 😄

darioush (dismissed):

qdm12:
Do not merge for now, I'll merge it right before the InspectDatabase libevm PR coming soon ™️

darioush (approved):

ceyonur (approved):

PR #791, opened 2025-02-12
chore(core/rawdb): `InspectDatabase` uses libevm (8)

qdm12:
- Part of the core/rawdb migration to use libevm
- Using libevm `InspectDatabase`
- Define options passed to libevm `InspectDatabase`

### DIFF
--- a/core/rawdb/database.go
+++ b/core/rawdb/database.go
@@ -241,36 +241,6 @@ func Open(o OpenOptions) (ethdb.Database, error) {
 	return kvdb, nil
 }
 
-type counter uint64
-
-func (c counter) String() string {
-	return fmt.Sprintf("%d", c)
-}
-
-func (c counter) Percentage(current uint64) string {
-	return fmt.Sprintf("%d", current*100/uint64(c))
-}
-
-// stat stores sizes and count for a parameter
-type stat struct {
-	size  common.StorageSize
-	count counter
-}
-
-// Add size to the stat and increase the counter by 1
-func (s *stat) Add(size common.StorageSize) {
-	s.size += size
-	s.count++
-}
-
-func (s *stat) Size() string {
-	return s.size.String()
-}
-
-func (s *stat) Count() string {
-	return s.count.String()
-}
-
 // InspectDatabase traverses the entire database and checks the size
 // of all different categories of data.
 func InspectDatabase(db ethdb.Database, keyPrefix, keyStart []byte) error {
--- a/core/rawdb/database.go
+++ b/core/rawdb/database.go
@@ -31,15 +31,14 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"time"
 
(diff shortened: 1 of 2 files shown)


darioush (approved):

core/rawdb/database.go
```diff
@@ -242,192 +241,64 @@ func Open(o OpenOptions) (ethdb.Database, error) {
...
+			newRows := make([][]string, 0, len(rows))
+			for _, row := range rows {
+				database := row[0]
+				category := row[1]
+				switch {
+				case database == "Key-Value store" && category == "Difficulties",
+					database == "Key-Value store" && category == "Beacon sync headers",
+					database == "Ancient store (Chain)":
+					continue
+				}
```
ceyonur, line 287: can we put some comments what we're doing here?

Also if we have luxury to change `WithDatabaseStatsTransformer` can we avoid checking literal string?

core/rawdb/database.go
```diff
@@ -242,192 +241,64 @@ func Open(o OpenOptions) (ethdb.Database, error) {
...
-				unaccounted.Add(size)
+		}),
+		ethrawdb.WithDatabaseStatsTransformer(func(rows [][]string) [][]string {
+			newRows := make([][]string, 0, len(rows))
+			for _, row := range rows {
+				database := row[0]
+				category := row[1]
+				switch {
+				case database == "Key-Value store" && category == "Difficulties",
+					database == "Key-Value store" && category == "Beacon sync headers",
```
ceyonur, line 284: IMHO we can keep these as 0, not a big deal and certainly not worth to have literal strings being checked for these.

qdm12: Done, although this changes the expected test a bit then.

core/rawdb/database.go
qdm12: Comment added.
And no luxury here, this is original geth code using a 2d slice of strings, so we have to compare this way.

qdm12: Switch case removed

core/rawdb/database.go
qdm12: Reverted (reset) the commit for this, because otherwise our test would fail occasionally because the ancient database rows are received in random order (due to a map in libevm code). The solution would then had been to sort rows, so we might as well remove rows we don't want (same code "quantity" almost)

darioush (approved):

ARR4N (approved):
You can really simplify this by converting it to a [table-driven approach](https://github.com/ava-labs/coreth/compare/qdm12/core/rawdb/inspect-database...arr4n/core/rawdb/inspect-database) as you're applying identical logic to 4 sets of parameters. Feel free to use that commit or merge as is if you prefer.

> [!TIP]
> If you want to use it you can easily fast-forward with `git merge --ff-only origin/arr4n/core/rawdb/inspect-database` from your branch.

ceyonur (approved):

PR #796, opened 2025-02-13
fix(internal/ethapi): fix nil block in debug apis

darioush:
Cherry-picks https://github.com/ethereum/go-ethereum/commit/da71839a270a353bac92e3108e4b74fb0eefec29 from upstream.

Also added a precautionary check at the callsite in api_extra in coreth code.


alarso16 (approved):

qdm12 (approved):

internal/ethapi/api_extra.go
```diff
@@ -115,7 +115,7 @@ func (s *BlockChainAPI) stateQueryBlockNumberAllowed(blockNumOrHash rpc.BlockNum
 		number = uint64(blockNumOrHash.BlockNumber.Int64())
 	} else {
 		block, err := s.b.BlockByNumberOrHash(context.Background(), blockNumOrHash)
```
ceyonur, line 117: nit: I wonder if we should just use `BlockByHash` instead of having an invariant that this will always be called with a hash. (or really directly use `BlockByNumberOrHash` instead of checking `.BlockNumber !=nil)

internal/ethapi/api_extra.go
```diff
@@ -113,12 +114,14 @@ func (s *BlockChainAPI) stateQueryBlockNumberAllowed(blockNumOrHash rpc.BlockNum
 	var number uint64
 	if blockNumOrHash.BlockNumber != nil {
 		number = uint64(blockNumOrHash.BlockNumber.Int64())
-	} else {
-		block, err := s.b.BlockByNumberOrHash(context.Background(), blockNumOrHash)
-		if err != nil {
+	} else if blockHash, ok := blockNumOrHash.Hash(); ok {
+		block, err := s.b.BlockByHash(context.Background(), blockHash)
+		if block == nil || err != nil {
```
qdm12, line 119: if the block is nil, you cannot wrap the error below really (as in it won't panic, but it will wrap oddly with the nil error)

```suggestion
		if block == nil {
			return fmt.Errorf("block from hash doesn't exist")
		} else if err != nil {
```

qdm12 (approved):

ceyonur (approved):

PR #798, opened 2025-02-13
cleanup: chain config changes for similarty with subnet-evm 

darioush:
## Why this should be merged
Separated this from https://github.com/ava-labs/coreth/pull/768, this will make it more similar to subnet-evm.

## How this works
Avoids setting the whole chain config, only setting the NetworkUpgrades from the chain context.

## How this was tested
CI

## Need to be documented?
No

No


params/config.go
```diff
@@ -52,8 +52,8 @@ var (
 		AvalancheContext:    AvalancheContext{utils.TestSnowContext()},
 		ChainID:             big.NewInt(1),
 		HomesteadBlock:      big.NewInt(0),
-		DAOForkBlock:        nil,
-		DAOForkSupport:      false,
+		DAOForkBlock:        big.NewInt(0),
+		DAOForkSupport:      true,
```
qdm12, line 56: Any consequence changing these for coreth in particular?

qdm12 (approved):

ceyonur (approved):

PR #801, opened 2025-02-14
eth, ethapi, plugin/evm: configurable suggest price options

ceyonur:
## Why this should be merged

Currently suggest price option params are constants and requires a release to be configured.

## How this works

It moves those constants to chain configs

## How this was tested

UTs + locally

## Need to be documented?

Yes

Yes


eth/ethconfig/config.go
```diff
@@ -178,4 +179,7 @@ type Config struct {
 	// This is useful for validators that don't need to index transactions.
 	// TransactionHistory can be still used to control unindexing old transactions.
 	SkipTxIndexing bool
+
+	// TODO: remove once we move SuggestPriceOptions API
```
darioush, line 183:
```suggestion
	// TODO: remove once we move SuggestPriceOptions to AVAX/custom API
```

eth/ethconfig/config.go
```diff
@@ -178,4 +179,7 @@ type Config struct {
...
+	ethapi.PriceOptionConfig `toml:",omitempty"`
```
darioush, line 184: is it necessary to embed this field?

internal/ethapi/backend.go
```diff
@@ -60,6 +60,7 @@ type Backend interface {
 	RPCGasCap() uint64            // global gas cap for eth_call over rpc: DoS protection
 	RPCEVMTimeout() time.Duration // global timeout for eth_call over rpc: DoS protection
 	RPCTxFeeCap() float64         // global tx fee cap for all transaction related APIs
+	PriceOptionsConfig() PriceOptionConfig
```
darioush, line 63: nit: let's move this below the wspace to separate our functions from geth's

plugin/evm/config/config.go
```diff
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/ava-labs/avalanchego/utils/constants"
+	"github.com/ava-labs/coreth/params"
```
darioush, line 12: I believe this will cause problems with libevm. Let's try to work around it instead

internal/ethapi/api.coreth.go
```diff
@@ -67,15 +64,23 @@ func (s *EthereumAPI) SuggestPriceOptions(ctx context.Context) (*PriceOptions, e
 		return nil, nil
 	}
 
+	cfg := s.b.PriceOptionsConfig()
+	bigSlowFeePerc := new(big.Int).SetUint64(cfg.SlowFeePercentage)
```
darioush, line 68: nit: seems we can rename to `bigSlowFeePercent` and `SlowFeePercent` to strike a balance between brevity and avoiding abbreviations.
perhaps even `SlowFeeMultiplierPercent`?

internal/ethapi/api.coreth_test.go
```diff
@@ -51,56 +65,84 @@ func TestSuggestPriceOptions(t *testing.T) {
...
 				),
 				Normal: newPrice(
 					minGasTip,
-					minBaseFee+minGasTip,
+					uint64(minBaseFee+minGasTip),
 				),
 				Fast: newPrice(
 					minGasTip,
-					(fastFeeNumerator*minBaseFee)/feeDenominator+(fastFeeNumerator*minGasTip)/feeDenominator,
+					(testCfg.FastFeePercentage*uint64(minBaseFee)/feeDenominator)+(testCfg.FastFeePercentage*uint64(minGasTip)/feeDenominator),
```
darioush, line 80: is it possible to set `fastFeeNumerator := testCfg.FastFeePercentage` and other local variables in this test to make the diff smaller?

RELEASES.md
```diff
@@ -2,6 +2,9 @@
 
 ## Pending Release
 
+- Added `eth_suggestPriceOptions` API to suggest gas prices (slow, normal, fast) based on the current network conditions
+- Added `"po-slow-fee-percentage"`, `"po-fast-fee-percentage"`, `"po-max-base-fee"`, and `"po-max-tip"` fields to chainConfig to configure the new `eth_suggestPriceOptions` API
```
darioush, line 6: nit: let's avoid the term chainConfig, initially it confused me. I think usually we refer to these as configuration options or flags.

eth/ethconfig/config.go
ceyonur: certainly not -.-

plugin/evm/config/config.go
ceyonur: right :/

internal/ethapi/api.coreth_test.go
```diff
@@ -51,56 +70,84 @@ func TestSuggestPriceOptions(t *testing.T) {
...
+					((slowFeeNumerator*maxNormalBaseFee)/feeDenominator)+((slowFeeNumerator*maxNormalGasTip)/feeDenominator),
 				),
 				Normal: newPrice(
 					maxNormalGasTip,
 					maxNormalBaseFee+maxNormalGasTip,
 				),
 				Fast: newPrice(
-					(fastFeeNumerator*maxNormalGasTip)/feeDenominator,
-					(fastFeeNumerator*maxNormalBaseFee)/feeDenominator+(fastFeeNumerator*maxNormalGasTip)/feeDenominator,
+					((fastFeeNumerator * maxNormalGasTip) / feeDenominator),
```
darioush, line 129: seems a few of these diffs can be removed

darioush (approved):

internal/ethapi/api.coreth.go
```diff
@@ -105,26 +109,28 @@ type feeSpeeds struct {
 // calculateFeeSpeeds returns the slow, normal, and fast price options for a
 // given min, estimate, and max,
 //
-// slow   = max(0.95 * min(estimate, maxFee), minFee)
+// slow   = max(slowPerc/100 * min(estimate, maxFee), minFee)
```
qdm12, line 112:
```suggestion
// slow   = max(slowFeePerc/100 * min(estimate, maxFee), minFee)
```

internal/ethapi/api.coreth.go
```diff
@@ -105,26 +109,28 @@ type feeSpeeds struct {
...
 // normal = min(estimate, maxFee)
-// fast   = 1.05 * estimate
+// fast   = fastPerc/100 * estimate
```
qdm12, line 114:
```suggestion
// fast   = fastFeePerc/100 * estimate
```

RELEASES.md
```diff
@@ -4,6 +4,9 @@
 - Bump golang version to v1.23.6
 - Bump golangci-lint to v1.63 and add linters
 
+- Added `eth_suggestPriceOptions` API to suggest gas prices (slow, normal, fast) based on the current network conditions
+- Added `"po-slow-fee-percentage"`, `"po-fast-fee-percentage"`, `"po-max-base-fee"`, and `"po-max-tip"` config flags to configure the new `eth_suggestPriceOptions` API
```
qdm12, line 8: is `po` standard to mean price options? Perhaps i.e. `price-option-slow-fee-percentage` would be clearer?

ceyonur: Yes I think it needs a prefix to specify this is applied to `priceOption` API. I was skeptical having such a long name, but surely can do.

PR #802, opened 2025-02-14
Split `CalcExtraData` out of `CalcBaseFee`

StephenButtolph:
## Why this should be merged

As part of the ACP-176 work, the `header.Extra` field will need to be calculated after the block has has been built but the `header.BaseFee` field will need to be calculated before the block has been built.

In this PR, the location that `header.Extra` is calculated is _not_ modified.

## How this works

- Converts the signature of `CalcBaseFee` from returning both the `extra` and `baseFee` to just return the `baseFee`
- Introduces a `CalcExtraPrefix` that returns the `extra`

## How this was tested

Existing tests (we could separate these) if that makes sense.

## Need to be documented?

No.

No.


consensus/dummy/dynamic_fees.go
```diff
@@ -143,34 +105,119 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header, timestamp uin
...
+		var (
+			timeElapsed    = timestamp - parent.Time
+			windowsElapsed = timeElapsed / params.RollupWindow
+		)
+		if windowsElapsed > 1 {
+			bigWindowsElapsed := new(big.Int).SetUint64(windowsElapsed)
+			// Because baseFeeDelta could actually be [common.Big1], we must not
+			// modify the existing value of `baseFeeDelta` but instead allocate
+			// a new one.
+			baseFeeDelta = new(big.Int).Mul(baseFeeDelta, bigWindowsElapsed)
```
StephenButtolph, line 134: This is technically a real change, but I think the code is much more clear (and it should result in the same behavior)

darioush: I don't see any behavior differences.

consensus/dummy/dynamic_fees.go
```diff
@@ -143,34 +105,119 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header, timestamp uin
 		num.Div(num, parentGasTargetBig)
 		num.Div(num, baseFeeChangeDenominator)
 		baseFeeDelta := math.BigMax(num, common.Big1)
+
+		if timestamp < parent.Time {
```
darioush, line 109: I noticed this check used to apply to both branches of `if totalGas > parentGasTarget`

consensus/dummy/dynamic_fees.go
```diff
@@ -143,34 +105,119 @@ func CalcBaseFee(config *params.ChainConfig, parent *types.Header, timestamp uin
...
+		return DynamicFeeWindow{}, nil
+	}
+
+	dynamicFeeWindow, err := ParseDynamicFeeWindow(parent.Extra)
+	if err != nil {
+		return DynamicFeeWindow{}, err
+	}
+
+	if timestamp < parent.Time {
+		return DynamicFeeWindow{}, fmt.Errorf("cannot calculate base fee for timestamp %d prior to parent timestamp %d", timestamp, parent.Time)
```
darioush, line 175:
```suggestion
		return DynamicFeeWindow{}, fmt.Errorf("cannot calculate fee window for timestamp %d prior to parent timestamp %d", timestamp, parent.Time)
```

darioush (approved):

consensus/dummy/dynamic_fees.go
StephenButtolph: Technically the prior code would have multiplied by `1` where the new code just skips.

The prior code was:
```golang
if timeElapsed > params.RollupWindow {
```

The new code is essentially:
```golang
if timeElapsed >= 2*params.RollupWindow {
```

consensus/dummy/dynamic_fees.go
StephenButtolph: The function will still error due to the prior `calcFeeWindow` call.

alarso16 (approved):
Besides unnecessary timestamp check, everything makes sense to me.

consensus/dummy/dynamic_fees.go
alarso16: It's a little weird to have, but doesn't hurt anything I guess.

PR #804, opened 2025-02-14
Support all upgrades in CalcBaseFee and CalcExtraPrefix

StephenButtolph:
## Why this should be merged

I personally find it more clear to have moved the the activation logic into the `CalcBaseFee` and `CalcExtraPrefix` functions. But if other people feel strongly that is isn't a good direction to move then I can close this PR.

Admittedly I dislike returning a `nil` `*big.Int` without an error, but it does seem to make sense if the expected `BaseFee` for a block is itself `nil`... Open to feedback.

## How this works

Moves `if`s inside a common function.

## How this was tested

It isn't yet.

## Need to be documented?

No.

No.


consensus/dummy/dynamic_fees.go
```diff
@@ -37,22 +37,35 @@ var (
...
-	window, err := calcFeeWindow(config, parent, timestamp)
-	return window.Bytes(), err
+	switch {
+	case config.IsApricotPhase3(timestamp):
+		window, err := calcFeeWindow(config, parent, timestamp)
+		return window.Bytes(), err
+	default:
+		// Prior to AP3 there was no expected extra prefix.
+		return nil, nil
+	}
```
StephenButtolph, line 52: This switch on `timestamp` will need to exist with ACP-176, the question is just if the default case should error (this function shouldn't be called before AP3) or if the function handles it as it does here.

Same comment on `CalcBaseFee` below.

consensus/dummy/dynamic_fees.go
```diff
@@ -37,22 +37,35 @@ var (
...
 }
 
 // CalcBaseFee takes the previous header and the timestamp of its child block
-// and calculates the expected base fee for the child block.
-//
-// CalcBaseFee should only be called if [timestamp] >= [config.ApricotPhase3Timestamp]
+// and calculates the expected base fee for the child block which may be nil.
 func CalcBaseFee(config *params.ChainConfig, parent *types.Header, timestamp uint64) (*big.Int, error) {
+	switch {
```
darioush, line 58: nit: can simplify like:
```
if !config.IsApricotPhase3(timestamp) {
	return nil, nil
}
return calcBaseFeeWithWindow(config, parent, timestamp)
```

consensus/dummy/dynamic_fees.go
```diff
@@ -37,22 +37,35 @@ var (
...
+		// Prior to AP3 there was no expected extra prefix.
+		return nil, nil
+	}
 }
 
 // CalcBaseFee takes the previous header and the timestamp of its child block
-// and calculates the expected base fee for the child block.
-//
-// CalcBaseFee should only be called if [timestamp] >= [config.ApricotPhase3Timestamp]
+// and calculates the expected base fee for the child block which may be nil.
```
darioush, line 56:
```suggestion
// and calculates the expected base fee for the child block which may be nil prior to AP3.
```

darioush (approved):

consensus/dummy/dynamic_fees.go
```diff
@@ -37,22 +37,37 @@ var (
 
 // CalcExtraPrefix takes the previous header and the timestamp of its child
 // block and calculates the expected extra prefix for the child block.
```
ceyonur, line 39: should we also add `which maybe nil` comment here as well?

StephenButtolph: We could, I felt like a `nil` `[]byte` was more normal (because a `nil` slice is essentially equivalent to an empty slice.)

ceyonur (approved):

consensus/dummy/dynamic_fees.go
```diff
@@ -37,22 +37,37 @@ var (
...
+		// Prior to AP3 there was no expected extra prefix.
+		return nil, nil
+	}
 }
 
 // CalcBaseFee takes the previous header and the timestamp of its child block
 // and calculates the expected base fee for the child block.
 //
-// CalcBaseFee should only be called if [timestamp] >= [config.ApricotPhase3Timestamp]
+// Prior to AP3, the returned base fee will be nil.
```
qdm12, line 58: I think this approach is better since it enforces this without having the consumer (dev) of this function pay attention to this comment.

consensus/dummy/dynamic_fees.go
```diff
@@ -37,22 +37,37 @@ var (
...
+	switch {
+	case config.IsApricotPhase3(timestamp):
+		return calcBaseFeeWithWindow(config, parent, timestamp)
+	default:
+		// Prior to AP3 the expected base fee is nil.
+		return nil, nil
+	}
+}
+
+// calcBaseFeeWithWindow should only be called if [timestamp] >= [config.ApricotPhase3Timestamp]
```
qdm12, line 69: I believe using brackets won't work on argument variables. You could use `[params.ChainConfig.ApricotPhase3Timestamp]` otherwise, but that's a bit confusing. I would recommend sticking on backticks, such that the dev doesn't try to ctrl+click the non-working link

```suggestion
// calcBaseFeeWithWindow should only be called if `timestamp` >= `config.ApricotPhase3Timestamp`
```

consensus/dummy/dynamic_fees.go
```diff
@@ -37,22 +37,37 @@ var (
...
 	config *params.ChainConfig,
 	parent *types.Header,
 	timestamp uint64,
 ) ([]byte, error) {
-	window, err := calcFeeWindow(config, parent, timestamp)
-	return window.Bytes(), err
+	switch {
+	case config.IsApricotPhase3(timestamp):
+		window, err := calcFeeWindow(config, parent, timestamp)
+		return window.Bytes(), err
```
qdm12, line 48: nit just to be safe, in case i.e. window is nil now or in the future (+wrap error perhaps - up to you)

```suggestion
		if err != nil {
			return nil, fmt.Errorf("failed to calculate fee window: %w", err)
		}
		return window.Bytes(), nil
```

consensus/dummy/consensus.go
```diff
@@ -139,29 +139,22 @@ func (eng *DummyEngine) verifyHeaderGasFields(config *params.ChainConfig, header
...
-		if !utils.BigEqual(header.BaseFee, expectedBaseFee) {
-			return fmt.Errorf("expected base fee (%d), found (%d)", expectedBaseFee, header.BaseFee)
-		}
+	// Verify header.BaseFee matches the expected value.
+	expectedBaseFee, err := CalcBaseFee(config, parent, header.Time)
+	if err != nil {
+		return fmt.Errorf("failed to calculate base fee: %w", err)
+	}
+	if !utils.BigEqual(header.BaseFee, expectedBaseFee) {
+		return fmt.Errorf("expected base fee (%d), found (%d)", expectedBaseFee, header.BaseFee)
```
qdm12, line 157: nit can we stop using parentheses where it doesn't make much sense to use them 🙏  Unless there is a reason but it itches my eyes 🤩 

```suggestion
		return fmt.Errorf("expected base fee %d, found %d", expectedBaseFee, header.BaseFee)
```

PR #764, opened 2025-01-24
chore(lint): cleanup config and enable default linters for our code

qdm12:
## Why this should be merged

Enable default linters (errcheck, staticcheck, gosimple and staticcheck were missing) + `errorlint`  for **our code only**:

- `./plugin/*.go` files
- `*libevm*.go` files
- `_ext.go` files
- More suggestions are welcome

More [linters](https://golangci-lint.run/usage/linters/) can be added subsequently, but this PR fixes already quite a few errors already coming from enabling the default linters and errorlint in **our** code

## How this works

- [x] Remove unused and unneeded configuration options
- [x] Existing geth code uses the same linters as before
- [x] No golangci-lint upgrade, wait for geth sync to upgrade the linter and get upstream lint error fixes
- [x] Our code is linted with more linters (default ones+errorlint)

## How this was tested

- [x] Linter passing locally
- [x] CI passing

## Need to be documented?

No

## Need to update RELEASES.md?

No

PR #805, opened 2025-02-17
ci: bump golang to 1.23.6

darioush:
## Why this should be merged
Needed to match avalanchego

## How this works
1. Bumps go & avalanchego dependency
2. Cherry-picks geth commit that fixes copyloopvar
3. Fix copyloopvar in our code
4. Align linters w/ upstream with a couple exceptions

## How this was tested
CI

## Need to be documented?
N/A

Y


.golangci.yml
```diff
@@ -5,14 +5,31 @@ run:
...
   disable-all: true
   enable:
     - goimports
+    - govet
+    - ineffassign
+    - misspell
+    - unconvert
+    - typecheck
+    - unused
+    # - staticcheck
```
darioush, line 22: this one has a lot of warnings, we should fix them but probably after taking changes from upstream.

.golangci.yml
```diff
@@ -5,14 +5,31 @@ run:
...
+    - bidichk
+    - durationcheck
+    - copyloopvar
+    - whitespace
+    # - revive # only certain checks enabled
```
darioush, line 27: this one has a separate commit, better take it along with the changes that fixes receiver names

.golangci.yml
```diff
@@ -18,8 +18,19 @@ linters:
     - ineffassign
     - misspell
     - unconvert
+    - typecheck
```
darioush, line 21: took the liberty to align these with upstream with the two exceptions below which would cause lots of warnings

.golangci.yml
```diff
@@ -18,8 +18,19 @@ linters:
...
     - unused
+    # - staticcheck
+    - bidichk
+    - durationcheck
+    - copyloopvar
     - whitespace
+    # - revive # only certain checks enabled
+    - durationcheck
```
darioush, line 29: this one is repeated like this in upstream at this commit, figured I'll leave it for now

.golangci.yml
qdm12: Would it be ok then to merge https://github.com/ava-labs/coreth/pull/764 before? 🤔 
It cleans up linter configuration etc. and only adds new linters for **our** files. Then we can add linters in this PR on top of it? I didn't take it out of draft because I was originally waiting for a libevm x master merge first 😄

.golangci.yml
qdm12: We can early-fix errors in our code by using #764

RELEASES.md
```diff
@@ -1,6 +1,8 @@
 # Release Notes
 
 ## Pending Release
+- Bump golang version to v1.23.6
+- Bump golangci-lint to v1.63
```
qdm12, line 5:
```suggestion
- Bump golangci-lint to v1.63 and add linters
```

qdm12 (approved):

.golangci.yml
```diff
@@ -5,7 +5,7 @@ run:
   tests: true
   # default is true. Enables skipping of directories:
   #   vendor$, third_party$, testdata$, examples$, Godeps$, builtin$
-  skip-dirs-use-default: true
+  issues.exclude-dirs-use-default: true
```
qdm12, line 8: Remove this please, it's not in geth code either

```suggestion
```

PR #808, opened 2025-02-18
Refactor AP3 `Window` and Params

StephenButtolph:
## Why this should be merged

1. This removes _most_ of the dependencies on the `dummy` package. (Only exception is the usage in `plugin/evm/vm.go`.)
2. This moves the fee window definition and it corresponding params to the new ap3 package.
3. This moves the `ExtraPrefix` and `BaseFee` calculations into the `header` package

## How this works

Just moves code around.

## How this was tested

Existing tests (followup PR will be to expand the testing coverage of the `header` package. I had started to improve the tests for this PR be felt it would be better to keep this PR as just a refactor.

## Need to be documented?

No.

No.


plugin/evm/header/dynamic_fee_windower.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func feeWindowBytes(w ap3.Window) []byte {
+	bytes := make([]byte, FeeWindowSize)
+	for i, v := range w {
+		offset := i * wrappers.LongLen
+		binary.BigEndian.PutUint64(bytes[offset:], v)
+	}
+	return bytes
+}
```
StephenButtolph, line 261: The binary representation of the window is in this package rather than the `ap3` package because the binary representation of the acp176 format will essentially need to be in this package to make any sense... And it felt weird for them not to be co-located.

core/txpool/blobpool/blobpool_test.go
```diff
@@ -79,9 +80,9 @@ func init() {
 
 // overrideMinFee sets the minimum base fee to 1 wei for the duration of the test.
 func overrideMinFee(t *testing.T) {
-	orig := dummy.EtnaMinBaseFee
-	dummy.EtnaMinBaseFee = big.NewInt(1)
-	t.Cleanup(func() { dummy.EtnaMinBaseFee = orig })
+	orig := header.EtnaMinBaseFee
+	header.EtnaMinBaseFee = big.NewInt(1)
+	t.Cleanup(func() { header.EtnaMinBaseFee = orig })
```
StephenButtolph, line 85: This is super jank - we should try to get rid of this when possible.

eth/gasprice/gasprice.go
```diff
@@ -64,7 +65,7 @@ const (
 var (
 	DefaultMaxPrice           = big.NewInt(150 * params.GWei)
 	DefaultMinPrice           = big.NewInt(0 * params.GWei)
-	DefaultMinBaseFee         = big.NewInt(params.ApricotPhase3InitialBaseFee)
+	DefaultMinBaseFee         = big.NewInt(ap3.InitialBaseFee)
 	DefaultMinGasUsed         = big.NewInt(6_000_000) // block gas limit is 8,000,000
 	DefaultMaxLookbackSeconds = uint64(80)
 )
```
StephenButtolph, line 71: I'm pretty confused by all of these params fwiw. But I don't think we need to change them in this PR.

plugin/evm/header/base_fee_test.go
StephenButtolph, line 1: I'm going to re-write the tests in this file in a follow-up

plugin/evm/header/dynamic_fee_windower.go
```diff
@@ -0,0 +1,261 @@
...
+	EtnaMinBaseFee = big.NewInt(params.EtnaMinBaseFee)
+
+	ap3MaxBaseFee = big.NewInt(ap3.MaxBaseFee)
+	ap4MaxBaseFee = big.NewInt(ap4.MaxBaseFee)
+
+	ap3BaseFeeChangeDenominator = big.NewInt(ap3.BaseFeeChangeDenominator)
+	ap5BaseFeeChangeDenominator = big.NewInt(ap5.BaseFeeChangeDenominator)
+
+	errDynamicFeeWindowInsufficientLength = errors.New("insufficient length for dynamic fee window")
+)
```
StephenButtolph, line 44: Was pretty happy about unexporting these.

plugin/evm/header/dynamic_fee_windower.go
```diff
@@ -0,0 +1,261 @@
...
+	var (
+		baseFee  = new(big.Int).Set(parent.BaseFee)
+		totalGas = dynamicFeeWindow.Sum()
+	)
+	if totalGas == parentGasTarget {
+		// If the parent block used exactly its target gas, the baseFee stays
+		// the same.
+		//
+		// For legacy reasons, this is true even if the baseFee would have
+		// otherwise been clamped to a different range.
```
StephenButtolph, line 81: Added this comment here as well. I think we can actually technically remove this "bug" because we only ever increased the fee range.

Didn't want to forget about it - but also didn't want to rewrite this function

eth/gasprice/gasprice.go
alarso16: It looks like they're set `eth/ethconfig/config.go` as the default anyway... I feel like it makes more sense to define them there. But this is mostly from upstream (the mins were added by us).

miner/worker.go
```diff
@@ -156,24 +156,25 @@ func (w *worker) commitNewWork(predicateContext *precompileconfig.PredicateConte
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to calculate new extra prefix: %w", err)
 	}
-	header.BaseFee, err = dummy.CalcBaseFee(w.chainConfig, parent, timestamp)
+	baseFee, err := header.BaseFee(w.chainConfig, parent, timestamp)
 	if err != nil {
 		return nil, fmt.Errorf("failed to calculate new base fee: %w", err)
 	}
 
+	header := &types.Header{
```
alarso16, line 169: Possibly dumb question - does this name conflict with the import, and if it doesn't, why not?

plugin/evm/header/dynamic_fee_windower.go
alarso16: I think the behavior would be *slightly* different if you removed this, just because the base fee would be decremented by 1 in this case. Not sure it matters really

alarso16 (approved):

miner/worker.go
StephenButtolph: Golang allows redefining symbols in different (more specific) contexts.

plugin/evm/header/dynamic_fee_windower.go
StephenButtolph: Yeah we wouldn't be able to just remove this if. But we could clamp the parent base fee based on the new rules.

darioush (approved):

PR #810, opened 2025-02-18
Move AP5 params

StephenButtolph:

### DIFF
--- a/consensus/dummy/dynamic_fees.go
+++ b/consensus/dummy/dynamic_fees.go
@@ -11,6 +11,7 @@ import (
 	"github.com/ava-labs/coreth/core/types"
 	"github.com/ava-labs/coreth/params"
 	"github.com/ava-labs/coreth/plugin/evm/ap4"
+	"github.com/ava-labs/coreth/plugin/evm/ap5"
 	"github.com/ava-labs/coreth/plugin/evm/header"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/math"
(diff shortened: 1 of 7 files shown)


plugin/evm/ap5/params.go
```diff
@@ -0,0 +1,39 @@
...
+// See the file LICENSE for licensing terms.
+
+// AP5 defines constants used after the Apricot Phase 5 upgrade.
+package ap5
+
+const (
+	// BlockGasCostStep is the rate at which the block gas cost changes per
+	// second.
+	//
+	// This value modifies the previously used `ap4.BlockGasCostStep`.
```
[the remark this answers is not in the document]
StephenButtolph: The `[]` syntax only works for other packages if those packages are imported by this file - which isn't done, so I don't think these would work correctly.

consensus/dummy/dynamic_fees.go
```diff
@@ -30,7 +31,7 @@ var (
 	EtnaMinBaseFee              = big.NewInt(params.EtnaMinBaseFee)
 
 	ApricotPhase3BaseFeeChangeDenominator = new(big.Int).SetUint64(params.ApricotPhase3BaseFeeChangeDenominator)
```
alarso16, line 33: Just curious, are the ap3 constants coming later?

alarso16 (approved):

consensus/dummy/dynamic_fees.go
StephenButtolph: Yes - the ap3 refactor is going to be much larger, so I wanted to do that last.

PR #811, opened 2025-02-18
chore(all): minimal changes for libevm v1.13.14-0.2.0.rc.3 (1)

darioush:
- Bump libevm dependency to [v1.13.14-0.2.0.rc.3](https://github.com/ava-labs/libevm/releases/tag/v1.13.14-0.2.0.rc.3)
- Add `HeaderExtra` not-implemented implementation
- Minimal changes copied over from #746
- Cherry picked #805 to support libevm release using Go 1.23
- "Gas remaining" slight code changes

This should be the first pull request to get merged into libevm in the series of libevm-phase-2.5 PRs.

## PRs plan

(each of these branches are based off the previous one)

1. This PR: minimal changes for libevm v1.13.14-0.2.0.rc.3
1. Merge master into libevm #812 
1. Re-structure code a bit (no PR yet) using branch [`qdm12/libevm/structure`](https://github.com/ava-labs/coreth/tree/qdm12/libevm/structure)
1. Header extra RLP+JSON implementation PR: #746 
1. Header extra PostCopy hook PR: #759 
1. Body + Block hooks PR: #760
1. core/rawdb InspectDatabase test #790 
1. core/rawdb InspectDatabase uses libevm #791 
1. core/rawdb full replacement #772

### DIFF
--- a/accounts/abi/unpack_test.go
+++ b/accounts/abi/unpack_test.go
@@ -399,7 +399,6 @@ func TestMethodMultiReturn(t *testing.T) {
 		"Can not unpack into a slice with wrong types",
 	}}
 	for _, tc := range testCases {
-		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			require := require.New(t)
 			err := abi.UnpackIntoInterface(tc.dest, "multi", data)
@@ -957,7 +956,7 @@ func TestOOMMaliciousInput(t *testing.T) {
 		}
 		encb, err := hex.DecodeString(test.enc)
 		if err != nil {
-			t.Fatalf("invalid hex: %s" + test.enc)
+			t.Fatalf("invalid hex: %s", test.enc)
 		}
 		_, err = abi.Methods["method"].Outputs.UnpackValues(encb)
 		if err == nil {
--- a/core/rawdb/accessors_chain_test.go
+++ b/core/rawdb/accessors_chain_test.go
@@ -291,10 +291,10 @@ func TestBlockReceiptStorage(t *testing.T) {
 	// Insert the receipt slice into the database and check presence
 	WriteReceipts(db, hash, 0, receipts)
 	if rs := ReadReceipts(db, hash, 0, 0, params.TestChainConfig); len(rs) == 0 {
-		t.Fatalf("no receipts returned")
+		t.Fatal("no receipts returned")
 	} else {
 		if err := checkReceiptsRLP(rs, receipts); err != nil {
-			t.Fatalf(err.Error())
+			t.Fatal(err)
 		}
 	}
 	// Delete the body and ensure that the receipts are no longer returned (metadata can't be recomputed)
@@ -308,7 +308,7 @@ func TestBlockReceiptStorage(t *testing.T) {
 	}
 	// Ensure that receipts without metadata can be returned without the block body too
 	if err := checkReceiptsRLP(ReadRawReceipts(db, hash, 0), receipts); err != nil {
-		t.Fatalf(err.Error())
+		t.Fatal(err)
 	}
 	// Sanity check that body and header alone without the receipt is a full purge
 	WriteHeader(db, header)
--- a/core/types/state_account.go
+++ b/core/types/state_account.go
@@ -46,8 +46,11 @@ var (
 
 type isMultiCoin bool
 
-var IsMultiCoinPayloads = ethtypes.RegisterExtras[isMultiCoin]()
+var (
+	extras              = ethtypes.RegisterExtras[ethtypes.NOOPHeaderHooks, *ethtypes.NOOPHeaderHooks, isMultiCoin]()
+	IsMultiCoinPayloads = extras.StateAccount
+)
 
-func IsMultiCoin(a ethtypes.ExtraPayloadCarrier) bool {
-	return bool(IsMultiCoinPayloads.FromPayloadCarrier(a))
+func IsMultiCoin(s ethtypes.StateOrSlimAccount) bool {
+	return bool(extras.StateAccount.Get(s))
 }
--- a/core/types/state_account.go
+++ b/core/types/state_account.go
@@ -47,7 +47,11 @@ var (
 type isMultiCoin bool
 
 var (
-	extras              = ethtypes.RegisterExtras[ethtypes.NOOPHeaderHooks, *ethtypes.NOOPHeaderHooks, isMultiCoin]()
+	extras = ethtypes.RegisterExtras[
+		ethtypes.NOOPHeaderHooks, *ethtypes.NOOPHeaderHooks,
+		ethtypes.NOOPBodyHooks, *ethtypes.NOOPBodyHooks,
+		isMultiCoin,
+	]()
 	IsMultiCoinPayloads = extras.StateAccount
 )
 
--- a/core/types/state_account.go
+++ b/core/types/state_account.go
@@ -49,7 +49,7 @@ type isMultiCoin bool
 var (
 	extras = ethtypes.RegisterExtras[
 		ethtypes.NOOPHeaderHooks, *ethtypes.NOOPHeaderHooks,
-		ethtypes.NOOPBodyHooks, *ethtypes.NOOPBodyHooks,
+		ethtypes.NOOPBlockBodyHooks, *ethtypes.NOOPBlockBodyHooks,
 		isMultiCoin,
 	]()
 	IsMultiCoinPayloads = extras.StateAccount
--- a/core/types/state_account.go
+++ b/core/types/state_account.go
@@ -56,5 +56,5 @@ var (
 )
 
 func IsMultiCoin(s ethtypes.StateOrSlimAccount) bool {
-	return bool(extras.StateAccount.Get(s))
+	return bool(IsMultiCoinPayloads.Get(s))
 }
--- a/nativeasset/contract.go
+++ b/nativeasset/contract.go
@@ -141,7 +141,7 @@ func (c *NativeAssetCall) Run(accessibleState contract.AccessibleState, caller c
 	stateDB.SubBalanceMultiCoin(caller, assetID, assetAmount)
 	stateDB.AddBalanceMultiCoin(to, assetID, assetAmount)
 
-	ret, remainingGas, err = accessibleState.Call(to, callData, remainingGas, new(uint256.Int), vm.WithUNSAFECallerAddressProxying())
+	ret, err = accessibleState.Call(to, callData, remainingGas, new(uint256.Int), vm.WithUNSAFECallerAddressProxying())
 
 	// When an error was returned by the EVM or when setting the creation code
 	// above we revert to the snapshot and consume any gas remaining. Additionally
--- a/nativeasset/contract.go
+++ b/nativeasset/contract.go
@@ -141,7 +141,8 @@ func (c *NativeAssetCall) Run(accessibleState contract.AccessibleState, caller c
 	stateDB.SubBalanceMultiCoin(caller, assetID, assetAmount)
 	stateDB.AddBalanceMultiCoin(to, assetID, assetAmount)
 
-	ret, remainingGas, err = accessibleState.Call(to, callData, remainingGas, new(uint256.Int), vm.WithUNSAFECallerAddressProxying())
+	// Drop the returned gas, resulting in a free outgoing call to mirror historical behaviour.
+	ret, _, err = accessibleState.Call(to, callData, remainingGas, new(uint256.Int), vm.WithUNSAFECallerAddressProxying())
 
 	// When an error was returned by the EVM or when setting the creation code
 	// above we revert to the snapshot and consume any gas remaining. Additionally
--- a/nativeasset/contract.go
+++ b/nativeasset/contract.go
@@ -141,7 +141,7 @@ func (c *NativeAssetCall) Run(accessibleState contract.AccessibleState, caller c
 	stateDB.SubBalanceMultiCoin(caller, assetID, assetAmount)
 	stateDB.AddBalanceMultiCoin(to, assetID, assetAmount)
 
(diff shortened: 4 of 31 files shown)


qdm12 (approved):
Niceee 👍 !

accounts/abi/unpack_test.go
```diff
@@ -957,7 +956,7 @@ func TestOOMMaliciousInput(t *testing.T) {
 		}
 		encb, err := hex.DecodeString(test.enc)
 		if err != nil {
-			t.Fatalf("invalid hex: %s" + test.enc)
+			t.Fatalf("invalid hex: %s", test.enc)
```
ARR4N, line 959:
```suggestion
			t.Fatalf("decoding hex, error %v; input = %q, err, test.enc)
```

All errors originate from [this function](https://cs.opensource.google/go/go/+/refs/tags/go1.23.6:src/encoding/hex/hex.go;l=87) and explain what exactly was invalid. Using `%q` makes it easier to see empty strings.

core/rawdb/accessors_chain_test.go
```diff
@@ -291,10 +291,10 @@ func TestBlockReceiptStorage(t *testing.T) {
 	// Insert the receipt slice into the database and check presence
 	WriteReceipts(db, hash, 0, receipts)
 	if rs := ReadReceipts(db, hash, 0, 0, params.TestChainConfig); len(rs) == 0 {
-		t.Fatalf("no receipts returned")
+		t.Fatal("no receipts returned")
```
ARR4N, line 294: The `else` is unnecessary  because of this `t.Fatal`, which stops the test immediately. Reducing the nesting on the next check improves clarity.

I'd add a suggestion but GitHub sucks for code review and won't allow them on deleted lines.

core/types/state_account.go
```diff
@@ -46,8 +46,15 @@ var (
...
 type isMultiCoin bool
 
-var IsMultiCoinPayloads = ethtypes.RegisterExtras[isMultiCoin]()
+var (
+	extras = ethtypes.RegisterExtras[
+		ethtypes.NOOPHeaderHooks, *ethtypes.NOOPHeaderHooks,
+		ethtypes.NOOPBlockBodyHooks, *ethtypes.NOOPBlockBodyHooks,
+		isMultiCoin,
+	]()
+	IsMultiCoinPayloads = extras.StateAccount
```
ARR4N, line 55:
```suggestion
```

You've modified `IsMultiCoin()` to use `extras` directly so I suspect this is now unnecessary.

params/hooks_libevm.go
```diff
@@ -171,7 +172,7 @@ func (a accessableState) GetSnowContext() *snow.Context {
 	return GetExtra(a.env.ChainConfig()).SnowCtx
 }
 
-func (a accessableState) Call(addr common.Address, input []byte, gas uint64, value *uint256.Int, opts ...vm.CallOption) (ret []byte, gasRemaining uint64, _ error) {
+func (a accessableState) Call(addr common.Address, input []byte, gas uint64, value *uint256.Int, opts ...vm.CallOption) (ret []byte, err error) {
```
ARR4N, line 175: Removing the return argument from here risks your implementations not consuming gas as expected. My first inclination is to return `a.env.Gas()` along with the `([]byte, error)` but I think this needs a test to demonstrate that it's correct (is there already one that you're confident with?).

nativeasset/contract.go
```diff
@@ -141,7 +141,7 @@ func (c *NativeAssetCall) Run(accessibleState contract.AccessibleState, caller c
 	stateDB.SubBalanceMultiCoin(caller, assetID, assetAmount)
 	stateDB.AddBalanceMultiCoin(to, assetID, assetAmount)
 
-	ret, remainingGas, err = accessibleState.Call(to, callData, remainingGas, new(uint256.Int), vm.WithUNSAFECallerAddressProxying())
+	ret, err = accessibleState.Call(to, callData, remainingGas, new(uint256.Int), vm.WithUNSAFECallerAddressProxying())
```
ARR4N, line 144: See my comment on the change to the return args of `Call()`. This will now never consume any gas; is that correct?

params/hooks_libevm.go
qdm12: If removing the returned gas had the CI pass, I doubt there is an existing test checking this. I reverted this to return `a.env.Gas()`. If you don't mind, given there isn't another way to return back remaining gas from the accessible state anyway, let's just skip adding-a-test for now (added to #814) since that test should had been there before (master branch or libevm branch). Just trying to get libevm out ASAP, and then I can work on that test (ideally, whilst final libevm reviews are ongoing)

nativeasset/contract.go
qdm12: Probably incorrect, changed this to return `a.env.Gas()` as suggested, good catch!

params/hooks_libevm.go
darioush: :) Good thing I added that UT. We shouldn't modify the core VM path based on what looks good subjectively.

ARR4N: Sorry, I looked more closely and the correct implementation is as follows:

```go
func (a accessableState) Call(addr common.Address, input []byte, gas uint64, value *uint256.Int, opts ...vm.CallOption) (ret []byte, gasRemaining uint64, err error) {
	start := a.env.Gas()
	ret, err = a.env.Call(addr, input, gas, value, opts...)
	used := start - a.env.Gas()
	if used > gas {
		return ret, 0, vmerrs.ErrOutOfGas
	}
	return ret, gas - used, err
}
```

While experimenting I changed UTs to add an arbitrary amount on top of the minimum cost, expecting that as the returned gas instead of zero. This gives me a higher degree of confidence in the correctness of the gas test. 👉 Recommend you do the same 👈

darioush: if used > gas {
		return ret, 0, vmerrs.ErrOutOfGas
	}
seems like unreachable code right? env.Gas() can only go down and cannot go below 0 because of the checks in UseGas()
So this is equivalent to what it was before

ARR4N: > So this is equivalent to what it was before

The introduction of `start` and `used` is what differs and fixes the potential problem of not consuming any gas at the call site of this function.

> seems like unreachable code right?

Hehe I was going to comment that it's probably overly defensive! It can actually only ever go down by at most `gas` ([use](https://github.com/ava-labs/libevm/blob/d32c7e0090fc380d8b9c36cccd82fdc38e7ffe95/core/vm/environment.libevm.go#L137) all and then [refund](https://github.com/ava-labs/libevm/blob/d32c7e0090fc380d8b9c36cccd82fdc38e7ffe95/core/vm/environment.libevm.go#L157)) so, yes, the check can be removed.

darioush: :( Unfortunately your suggested implementation does not pass `TestPrestateWithDiffModeANTTracer`

I think this is because in the new convention, .Call() itself should consume the gas

qdm12: I was going schizophrenic into call stacks, thanks @darioush for the save!

nit: do we even need the if check? If not, we could just have

```go
	ret, err = a.env.Call(addr, input, gas, value, opts...)
	return ret, gas, err
```

And probably add a comment precising why we return `gas` and not `env.Gas()` or another variation.

ARR4N: If subtracting non-zero `used` from the returned gas breaks the tracer test but fixes the NAC test then I think there's something deeper at play here. It suggests to me that there is gas being consumed by the contract at `addr`, which the precompile using `a.Call()` (not `a.env.Call()`) isn't* charging in the tracer test—this could explain why the miner's balance is now higher than expected in the tracer test:

```
prestate_tracer_test.go:101: Miner: 0xD049bfd667cB46Aa3Ef5Df0dA3e57DB3Be39E511
        prestate_tracer_test.go:102: Tx from: 0xe296F389F90f3cF5EdeFb63690eb7a6257B69203
        prestate_tracer_test.go:103: trace mismatch
             have: {"post":{"0xd049bfd667cb46aa3ef5df0da3e57db3be39e511":{"balance":"0x24c9c5ac6abc00"},"0xe296f389f90f3cf5edefb63690eb7a6257b69203":{"balance":"0xcf3bbc9bfcd854400","nonce":1}},"pre":{"0xd049bfd667cb46aa3ef5df0da3e57db3be39e511":{"balance":"0x0"},"0xe296f389f90f3cf5edefb63690eb7a6257b69203":{"balance":"0xcf3e0938579f00000"}}}
             want: {"post":{"0xd049bfd667cb46aa3ef5df0da3e57db3be39e511":{"balance":"0x231b6c08a01600"},"0xe296f389f90f3cf5edefb63690eb7a6257b69203":{"balance":"0xcf3bd7819714fea00","nonce":1}},"pre":{"0xd049bfd667cb46aa3ef5df0da3e57db3be39e511":{"balance":"0x0"},"0xe296f389f90f3cf5edefb63690eb7a6257b69203":{"balance":"0xcf3e0938579f00000"}}}
```

Miner has 473,175 gwei (nAVAX) too much and sender has the same amount too little. I'm looking into this but sharing in case someone else wants to too.

*Possibly _wasn't_ charging if this is now deprecated functionality.

ARR4N: `used` is `0` for the NAC test and `2103` for the tracer one. `473,175 / 2103 == 225`, which is the gas price of the tracer test. This confirms my hypothesis re the change in miner and tx-sender balances in the tracer test.

To me this suggests that NAC (or something else in the plumbing) had a bug whereby the gas of the outgoing call wasn't being charged. Where we recreate this for backwards compatibility depends on where the bug was in the first place—I'll keep exploring and will add another comment as I find out more.

ARR4N: Assuming the problem was in the NAC implementation (would need to check chain history to confirm), I would recommend addressing it like in [this commit](https://github.com/ava-labs/coreth/compare/darioush/qdm12/libevm-phase-2.5...arr4n/libevm-phase-2.5), which (a) passes both NAC and tracer tests; and (b) returns the actual gas remaining from `accessableState.Call()`.

Also, please can we fix the typo in the type name `s/accessable/accessible/`?

darioush: > whereby the gas of the outgoing call wasn't being charged

Yes, this is the case. the NAC precompile existing behavior does not charge for the Call, where a.env.Call does

```
	snapshot := evm.StateDB.Snapshot()

	if !evm.StateDB.Exist(to) {
		if remainingGas < params.CallNewAccountGas {
			return nil, 0, vmerrs.ErrOutOfGas
		}
		remainingGas -= params.CallNewAccountGas
		evm.StateDB.CreateAccount(to)
	}

	// Increment the call depth which is restricted to 1024
	evm.depth++
	defer func() { evm.depth-- }()

	// Send [assetAmount] of [assetID] to [to] address
	evm.Context.TransferMultiCoin(evm.StateDB, caller, to, assetID, assetAmount)
	
	// typically EVM would charge gas for this call
	// but NAC doesn't do that

	ret, remainingGas, err = evm.Call(AccountRef(caller), to, callData, remainingGas, new(uint256.Int))
```

ARR4N:
```go
// typically EVM would charge gas for this call here
// but NAC doesn't do that

ret, remainingGas, err = evm.Call(AccountRef(caller), to, callData, remainingGas, new(uint256.Int))
```

The overwriting of `remainingGas` suggests that it's only the 100 for the `CALL` opcode that's not being charged. `evm.Call(...)` will do its own gas accounting and `remainingGas` will almost certainly reduce.

> There is also the chance that NAC was correct w.r.t. gas but that there is now double-counting of gas consumption. I'll explore this avenue and report back.

Looking at `hooks_libevm.go` as a whole, I'm leaning more heavily towards this hypothesis but am yet to formulate an experiment that can test it. It looks like `accessableState.Call()` is only used inside a `legacy` wrapper, which is good evidence for double counting. Sorry if this whole thing has been a wild-goose chase, but it's important that we know for sure.

ARR4N: Pushed ecb5e48f6eba48e8d044ff6ff2f450c155cd135c to resolve this.

ARR4N (approved):
🚀

nativeasset/contract.go
```diff
@@ -105,57 +105,67 @@ func UnpackNativeAssetCallInput(input []byte) (common.Address, common.Hash, *big
 
 // Run implements StatefulPrecompiledContract
 func (c *NativeAssetCall) Run(accessibleState contract.AccessibleState, caller common.Address, addr common.Address, input []byte, suppliedGas uint64, readOnly bool) (ret []byte, remainingGas uint64, err error) {
-	if suppliedGas < c.GasCost {
+	env := accessibleState.GetPrecompileEnv()
```
qdm12, line 108: nit should we rename `GetPrecompileEnv` to `PrecompileEnv`?

nativeasset/contract.go
```diff
@@ -105,57 +105,67 @@ func UnpackNativeAssetCallInput(input []byte) (common.Address, common.Hash, *big
...
+	if !env.UseGas(c.GasCost) {
 		return nil, 0, vmerrs.ErrOutOfGas
 	}
-	remainingGas = suppliedGas - c.GasCost
+	ret, err = c.run(env, accessibleState.GetStateDB(), caller, addr, input, readOnly)
+	// This precompile will be wrapped in a libevm `legacy` wrapper, which
```
qdm12, line 113: nit do we have a link to this legacy wrapper? Otherwise let's just not use backticks for it (was looking for a variable named `legacy`)

```suggestion
	// This precompile will be wrapped in a libevm legacy wrapper, which
```

nativeasset/contract.go
qdm12: Keeping it as GetXYZ to fit with other methods of the interface

nativeasset/contract.go
qdm12: Unimportant

PR #816, opened 2025-02-19
Improve `plugin/evm/header` tests

StephenButtolph:
## Why this should be merged

This PR primarily cleans up the tests in the `plugin/evm/header` package.

Additionally, it slightly cleans up the `MinRequiredTip` function and renames it to `EstimateRequiredTip` to align with the `EstimateNextBaseFee` function.

## How this works

1. Add tests
2. Rename

## How this was tested

CI

## Need to be documented?

No.

No.

### DIFF
--- a/params/denomination.go
+++ b/params/denomination.go
@@ -32,6 +32,6 @@ package params
 //	new(big.Int).Mul(value, big.NewInt(params.GWei))
 const (
 	Wei   = 1
-	GWei  = 1e9
-	Ether = 1e18
+	GWei  = 1_000_000_000 * Wei
+	Ether = 1_000_000_000 * GWei
 )
--- a/params/denomination.go
+++ b/params/denomination.go
@@ -32,6 +32,6 @@ package params
 //	new(big.Int).Mul(value, big.NewInt(params.GWei))
 const (
 	Wei   = 1
-	GWei  = 1_000_000_000 * Wei
-	Ether = 1_000_000_000 * GWei
+	GWei  = 1e9
+	Ether = 1e18
 )
--- a/plugin/evm/ap3/window.go
+++ b/plugin/evm/ap3/window.go
@@ -7,7 +7,7 @@ package ap3
 import (
 	"math"
 
-	"github.com/ava-labs/coreth/params"
+	"github.com/ava-labs/coreth/utils"
 	safemath "github.com/ethereum/go-ethereum/common/math"
 )
 
@@ -19,13 +19,13 @@ const (
 	// upgrade.
 	//
 	// This value was modified in Apricot Phase 4.
-	MinBaseFee = 75 * params.GWei
+	MinBaseFee = 75 * utils.GWei
 
 	// MaxBaseFee is the maximum base fee that is allowed after Apricot Phase 3
 	// upgrade.
 	//
 	// This value was modified in Apricot Phase 4.
-	MaxBaseFee = 225 * params.GWei
+	MaxBaseFee = 225 * utils.GWei
 
 	// InitialBaseFee is the base fee that is used for the first Apricot Phase 3
 	// block.
--- a/plugin/evm/header/base_fee_test.go
+++ b/plugin/evm/header/base_fee_test.go
@@ -142,14 +142,14 @@ func TestBaseFee(t *testing.T) {
 			timestamp: 1,
 			want: func() *big.Int {
 				const (
-					gasTarget                  = ap3.TargetGas
-					gasUsed                    = 2*ap3.TargetGas + ap3.IntrinsicBlockGas
-					amountOverTarget           = gasUsed - gasTarget
-					parentBaseFee              = ap3.MinBaseFee
-					smoothingFactor            = ap3.BaseFeeChangeDenominator
-					baseFeeFractionUnderTarget = amountOverTarget * parentBaseFee / gasTarget
-					delta                      = baseFeeFractionUnderTarget / smoothingFactor
-					baseFee                    = parentBaseFee + delta
+					gasTarget                 = ap3.TargetGas
+					gasUsed                   = 2*ap3.TargetGas + ap3.IntrinsicBlockGas
+					amountOverTarget          = gasUsed - gasTarget
+					parentBaseFee             = ap3.MinBaseFee
+					smoothingFactor           = ap3.BaseFeeChangeDenominator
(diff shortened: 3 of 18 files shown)


params/denomination.go
```diff
@@ -32,6 +32,6 @@ package params
 //	new(big.Int).Mul(value, big.NewInt(params.GWei))
 const (
 	Wei   = 1
-	GWei  = 1e9
-	Ether = 1e18
+	GWei  = 1_000_000_000 * Wei
+	Ether = 1_000_000_000 * GWei
```
StephenButtolph, line 36: Using scientific notation results in treating these as untyped _float_s rather than untyped _int_s. Golang will happily truncate integer constants, but it will not convert a float into an int unless it is lossless.

darioush: Can we make our own constants instead of modifying the shared code with geth?

plugin/evm/header/block_gas_cost.go
```diff
@@ -69,37 +69,40 @@ func BlockGasCostWithStep(
...
-// proportional to its gas usage). In reality, there is no minimum tip that
-// is enforced by the consensus engine and high tip paying transactions can
-// subsidize the inclusion of low tip paying transactions. The only
-// correctness check performed is that the sum of all tips is >= the
-// required block fee.
+// EstimateRequiredTip is the estimated tip a transaction would have needed to
+// pay to be included in a given block (assuming it paid a tip proportional to
+// its gas usage).
+//
+// In reality, there is no minimum tip that is enforced by the consensus engine
```
darioush, line 76: nit: active language :)
```suggestion
// In reality, the consensus engine does not enforce a minimum tip
```

plugin/evm/header/block_gas_cost.go
```diff
@@ -69,37 +69,40 @@ func BlockGasCostWithStep(
...
+	// totalGasUsed = GasUsed + ExtDataGasUsed
+	totalGasUsed := new(big.Int).SetUint64(header.GasUsed)
+	totalGasUsed.Add(totalGasUsed, header.ExtDataGasUsed)
+
+	// totalRequiredTips = blockGasCost * baseFee
+	totalRequiredTips := new(big.Int)
+	totalRequiredTips.Mul(header.BlockGasCost, header.BaseFee)
+
+	// requiredTip = totalRequiredTips / totalGasUsed
+	requiredTip := totalRequiredTips.Div(totalRequiredTips, totalGasUsed)
```
darioush, line 106: nit: maybe estimatedTipPerGas? or estimatedTip? since it's not really required

plugin/evm/header/block_gas_cost_test.go
```diff
@@ -160,3 +161,74 @@ func TestBlockGasCostWithStep(t *testing.T) {
...
+			},
+			wantErr: errExtDataGasUsedNil,
+		},
+		{
+			name:         "success",
+			ap4Timestamp: utils.NewUint64(0),
+			header: &types.Header{
+				GasUsed:        123,
+				BaseFee:        big.NewInt(456),
+				ExtDataGasUsed: big.NewInt(789),
```
darioush, line 211: nit: maybe put this next to GasUsed?

plugin/evm/header/base_fee_test.go
```diff
@@ -11,328 +11,344 @@ import (
...
 			},
+			timestamp: 1,
+			want: func() *big.Int {
+				const (
+					gasTarget                  = ap3.TargetGas
+					gasUsed                    = 2*ap3.TargetGas + ap3.IntrinsicBlockGas
+					amountOverTarget           = gasUsed - gasTarget
+					parentBaseFee              = ap3.MinBaseFee
+					smoothingFactor            = ap3.BaseFeeChangeDenominator
+					baseFeeFractionUnderTarget = amountOverTarget * parentBaseFee / gasTarget
```
darioush, line 150: should this be Over?
I understand it's for clarity but I think the calculation from this point can be done in a (local) function like baseFee(amount, target, parentBaseFee, smoothingFactor)
on the other hand, it's nice that one can mouseover and see the value.. not sure.

plugin/evm/header/base_fee_test.go
```diff
@@ -11,328 +11,344 @@ import (
...
-	parentTimestamp := uint64(1)
-	timestamp := parentTimestamp + ap3.WindowLen + 1000
+			config := &params.ChainConfig{
+				NetworkUpgrades: test.upgrades,
+			}
+			got, err := BaseFee(config, test.parent, test.timestamp)
+			require.ErrorIs(err, test.wantErr)
+			require.Equal(test.want, got)
 
-	parentHeader := &types.Header{
```
darioush, line 298: should we keep this regression test? (can be a case in the new test)
I see we kept the assertion but not sure that we have the conditions that triggered this.

params/denomination.go
StephenButtolph: moved to the `utils` package.

plugin/evm/header/block_gas_cost.go
StephenButtolph: went with `estimatedTip`

plugin/evm/header/block_gas_cost_test.go
StephenButtolph: sgtm

plugin/evm/header/base_fee_test.go
StephenButtolph: 🦅 👁️

StephenButtolph: I did it like this so that we could see the value of each constant in the IDE... but 🤷

plugin/evm/header/base_fee_test.go
StephenButtolph: This test was migrated to the new test as `ap3_big_1_not_modified`

darioush (approved):

plugin/evm/ap3/window.go
```diff
@@ -19,13 +19,13 @@ const (
 	// upgrade.
 	//
 	// This value was modified in Apricot Phase 4.
-	MinBaseFee = 75 * params.GWei
+	MinBaseFee = 75 * utils.GWei
```
ceyonur, line 22: I seriously wonder why this is in params package -,-

ceyonur (approved):

PR #822, opened 2025-02-20
Move `GasLimit` handling into the `header` package

StephenButtolph:
## Why this should be merged

1. Currently `header.GasLimit` is verified in multiple places (with duplicate code). This centralizes the verification in a single location.
2. Currently `header.GasLimit` is calculated in multiple places (with duplicate code). This centralizes the generation in a single location.
3. Adds tests for `GasLimit` calculations and verifications

## How this works

This PR utilizes the observation that `core.CalcGasLimit` is an optional block building rule, and is not required to be utilized. This PR changes the block building logic for pre-AP1 code to build blocks with the same gas limit as its parent (rather than moving toward the `ap1.GasLimit`.

## How this was tested

Added unit tests + CI

## Need to be documented?

No.

No.

### DIFF
--- a/plugin/evm/block_verification.go
+++ b/plugin/evm/block_verification.go
@@ -19,7 +19,6 @@ import (
 	"github.com/ava-labs/coreth/plugin/evm/upgrade/ap0"
 	"github.com/ava-labs/coreth/plugin/evm/upgrade/ap1"
 	"github.com/ava-labs/coreth/plugin/evm/upgrade/ap5"
-	"github.com/ava-labs/coreth/plugin/evm/upgrade/cortina"
 	"github.com/ava-labs/coreth/trie"
 	"github.com/ava-labs/coreth/utils"
 )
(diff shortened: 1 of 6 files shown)


plugin/evm/block_verification.go
```diff
@@ -109,23 +108,6 @@ func (v blockValidator) SyntacticVerify(b *Block, rules params.Rules) error {
...
-		}
-	} else if rules.IsApricotPhase1 {
-		if ethHeader.GasLimit != ap1.GasLimit {
-			return fmt.Errorf(
-				"expected gas limit to be %d after apricot phase 1 but got %d",
-				ap1.GasLimit, ethHeader.GasLimit,
-			)
-		}
-	}
-
```
StephenButtolph, line 128: With ACP-176, validating the gasLimit will require the parent header. This verification is already being performed in `consensus.go`, so removing this seemed to be the best approach.

alarso16 (approved):

ceyonur (approved):

PR #824, opened 2025-02-20
Move predicate header parsing

StephenButtolph:
## Why this should be merged

ACP-176 changes the parsing of the `header.Extra` field, so any parsing of the field now requires access to the chain rules.

## How this works

Moves the parsing of the `header.Extra` field into the `NewEVM` function so that the chain rules are available. 

## How this was tested

CI

## Need to be documented?

No.

No.

### DIFF
--- a/core/vm/evm.go
+++ b/core/vm/evm.go
@@ -34,6 +34,7 @@ import (
 	"github.com/ava-labs/coreth/constants"
 	"github.com/ava-labs/coreth/core/types"
 	"github.com/ava-labs/coreth/params"
+	"github.com/ava-labs/coreth/plugin/evm/header"
 	"github.com/ava-labs/coreth/precompile/contract"
 	"github.com/ava-labs/coreth/precompile/modules"
 	"github.com/ava-labs/coreth/precompile/precompileconfig"
@@ -41,6 +42,7 @@ import (
 	"github.com/ava-labs/coreth/vmerrs"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/log"
 	"github.com/holiman/uint256"
 )
 
@@ -122,9 +124,14 @@ type BlockContext struct {
 	TransferMultiCoin TransferMCFunc
 	// GetHash returns the hash corresponding to n
 	GetHash GetHashFunc
-	// PredicateResults are the results of predicate verification available throughout the EVM's execution.
-	// PredicateResults may be nil if it is not encoded in the block's header.
+	// PredicateResults are the results of predicate verification available
+	// throughout the EVM's execution.
+	//
+	// PredicateResults may be nil if it is not encoded in the extra field of
+	// the block's header or if the extra field has not been parsed yet.
 	PredicateResults *predicate.Results
+	// Extra is the extra field from the block header.
+	Extra []byte
 
 	// Block information
 	Coinbase    common.Address // Provides information for COINBASE
@@ -220,6 +227,30 @@ func NewEVM(blockCtx BlockContext, txCtx TxContext, statedb StateDB, chainConfig
 		chainRules:  chainConfig.Rules(blockCtx.BlockNumber, blockCtx.Time),
 	}
 	evm.interpreter = NewEVMInterpreter(evm)
+
+	// If the predicate results are already available, use them.
+	if blockCtx.PredicateResults != nil {
+		return evm
+	}
+
+	// Parse the predicate results from the extra field and store them in the
+	// block context.
+	predicateBytes := header.PredicateBytesFromExtra(blockCtx.Extra)
+	if len(predicateBytes) == 0 {
+		return evm
+	}
+
+	// The VM has already verified the correctness of the results during header
+	// validation.
+	results, err := predicate.ParseResults(predicateBytes)
+	if err != nil {
+		log.Error("Unexpected error parsing predicate results",
+			"err", err,
+		)
+		return evm
+	}
+
+	evm.Context.PredicateResults = results
 	return evm
 }
 
--- a/core/vm/evm.go
+++ b/core/vm/evm.go
@@ -228,7 +228,7 @@ func NewEVM(blockCtx BlockContext, txCtx TxContext, statedb StateDB, chainConfig
 	}
 	evm.interpreter = NewEVMInterpreter(evm)
 
-	// If the predicate results are already available, use them.
+	// If the predicate results were set by the miner, use them.
 	if blockCtx.PredicateResults != nil {
 		return evm
 	}
--- a/core/vm/evm.go
+++ b/core/vm/evm.go
@@ -250,6 +250,8 @@ func NewEVM(blockCtx BlockContext, txCtx TxContext, statedb StateDB, chainConfig
 		return evm
 	}
 
+	// Because the BlockContext is pass-by-value, this does not cache the
+	// results for future calls to NewEVM.
 	evm.Context.PredicateResults = results
 	return evm
 }
--- a/plugin/evm/block.go
+++ b/plugin/evm/block.go
@@ -19,6 +19,7 @@ import (
 	"github.com/ava-labs/coreth/core/types"
 	"github.com/ava-labs/coreth/params"
 	"github.com/ava-labs/coreth/plugin/evm/atomic"
+	"github.com/ava-labs/coreth/plugin/evm/header"
 	"github.com/ava-labs/coreth/precompile/precompileconfig"
 	"github.com/ava-labs/coreth/predicate"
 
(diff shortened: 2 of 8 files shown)


predicate/predicate_bytes.go
```diff
@@ -50,15 +49,3 @@ func UnpackPredicate(paddedPredicate []byte) ([]byte, error) {
...
-// true iff the predicate results bytes have non-zero length.
-func GetPredicateResultBytes(extraData []byte) ([]byte, bool) {
-	// Prior to Durango, the VM enforces the extra data is smaller than or equal to this size.
-	// After Durango, the VM pre-verifies the extra data past the dynamic fee rollup window is
-	// valid.
-	if len(extraData) <= header.FeeWindowSize {
-		return nil, false
-	}
-	return extraData[header.FeeWindowSize:], true
-}
```
StephenButtolph, line 64: > true iff the predicate results bytes have non-zero length

Since this is true, the bool value being returned is not needed.

plugin/evm/vm_warp_test.go
```diff
@@ -652,8 +653,7 @@ func testReceiveWarpMessage(
 
 	// Require the block was built with a successful predicate result
 	ethBlock := block2.(*chain.BlockWrapper).Block.(*Block).ethBlock
-	headerPredicateResultsBytes, ok := predicate.GetPredicateResultBytes(ethBlock.Extra())
-	require.True(ok)
+	headerPredicateResultsBytes := customheader.PredicateBytesFromExtra(ethBlock.Extra())
 	results, err := predicate.ParseResults(headerPredicateResultsBytes)
```
StephenButtolph, line 657: `ParseResults` is guaranteed to return an `err` if `headerPredicateResultsBytes` is empty.

plugin/evm/block.go
```diff
@@ -381,10 +382,7 @@ func (b *Block) verifyPredicates(predicateContext *precompileconfig.PredicateCon
 		return fmt.Errorf("failed to marshal predicate results: %w", err)
 	}
 	extraData := b.ethBlock.Extra()
-	headerPredicateResultsBytes, ok := predicate.GetPredicateResultBytes(extraData)
-	if !ok {
-		return fmt.Errorf("failed to find predicate results in extra data: %x", extraData)
-	}
```
StephenButtolph, line 387: `predicateResultsBytes` is guaranteed not to be empty here, so this check is not required.

core/vm/evm.go
```diff
@@ -220,6 +227,30 @@ func NewEVM(blockCtx BlockContext, txCtx TxContext, statedb StateDB, chainConfig
 		chainRules:  chainConfig.Rules(blockCtx.BlockNumber, blockCtx.Time),
 	}
 	evm.interpreter = NewEVMInterpreter(evm)
+
+	// If the predicate results are already available, use them.
+	if blockCtx.PredicateResults != nil {
+		return evm
+	}
```
StephenButtolph, line 234: This check supports the `NewEVMBlockContextWithPredicateResults` usage during block building.

core/vm/evm.go
```diff
@@ -220,6 +227,30 @@ func NewEVM(blockCtx BlockContext, txCtx TxContext, statedb StateDB, chainConfig
...
+
+	// Parse the predicate results from the extra field and store them in the
+	// block context.
+	predicateBytes := header.PredicateBytesFromExtra(blockCtx.Extra)
+	if len(predicateBytes) == 0 {
+		return evm
+	}
```
StephenButtolph, line 241: This check supports the pre-durango headers.

core/vm/evm.go
```diff
@@ -220,6 +227,30 @@ func NewEVM(blockCtx BlockContext, txCtx TxContext, statedb StateDB, chainConfig
...
+	// validation.
+	results, err := predicate.ParseResults(predicateBytes)
+	if err != nil {
+		log.Error("Unexpected error parsing predicate results",
+			"err", err,
+		)
+		return evm
+	}
+
+	evm.Context.PredicateResults = results
```
StephenButtolph, line 253: Note: because `BlockContext` is pass-by-value, this does not cache `results` for future calls to `NewEVM`.

ceyonur: can we comment that?

ceyonur (approved):

predicate/predicate_bytes.go
```diff
@@ -50,15 +49,3 @@ func UnpackPredicate(paddedPredicate []byte) ([]byte, error) {
 
 	return trimmedPredicateBytes[:len(trimmedPredicateBytes)-1], nil
 }
-
-// GetPredicateResultBytes returns the predicate result bytes from the extra data and
-// true iff the predicate results bytes have non-zero length.
-func GetPredicateResultBytes(extraData []byte) ([]byte, bool) {
```
ceyonur, line 56: cross ref here https://github.com/ava-labs/coreth/pull/823#discussion_r1965388402

core/vm/evm.go
```diff
@@ -122,9 +124,14 @@ type BlockContext struct {
...
-	// PredicateResults are the results of predicate verification available throughout the EVM's execution.
-	// PredicateResults may be nil if it is not encoded in the block's header.
+	// PredicateResults are the results of predicate verification available
+	// throughout the EVM's execution.
+	//
+	// PredicateResults may be nil if it is not encoded in the extra field of
+	// the block's header or if the extra field has not been parsed yet.
 	PredicateResults *predicate.Results
+	// Extra is the extra field from the block header.
+	Extra []byte
```
ARR4N, line 134: (info) `libevm` makes the entire [`Header` available in the `BlockContext`](https://github.com/ava-labs/libevm/blob/d32c7e0090fc380d8b9c36cccd82fdc38e7ffe95/core/vm/evm.go#L83) in lieu of adding the `PredicateResults`. Given the comment on `Extra` here, this should therefore be compatible.

core/vm/evm.go
```diff
@@ -220,6 +227,32 @@ func NewEVM(blockCtx BlockContext, txCtx TxContext, statedb StateDB, chainConfig
 		chainRules:  chainConfig.Rules(blockCtx.BlockNumber, blockCtx.Time),
 	}
 	evm.interpreter = NewEVMInterpreter(evm)
+
+	// If the predicate results are already available, use them.
+	if blockCtx.PredicateResults != nil {
```
ARR4N, line 232: (info) As above, this should be compatible with `libevm`. The `PredicateResults` will be in the payload registered to be carried with every `Header`. If the pattern in #746 of `types.GetHeaderExtra(*Header)` is what's eventually used then that will expose it. That said, this entire method will probably be defined on the registered type, not `BlockContext` as that will be replaced with the `libevm` equivalent.

core/vm/evm.go
```diff
@@ -220,6 +227,32 @@ func NewEVM(blockCtx BlockContext, txCtx TxContext, statedb StateDB, chainConfig
...
+		return evm
+	}
+
+	// Parse the predicate results from the extra field and store them in the
+	// block context.
+	predicateBytes := header.PredicateBytesFromExtra(blockCtx.Extra)
```
ARR4N, line 238: Why is the `Extra []byte` being copied from `Header` to `BlockContext` if it's eventually being interpreted by a method on the former? This suggests that either the copy is unnecessary (if the slices are the same) or that having the method defined on the `Header` is inappropriate (if the slices are different) because the two objects have become decoupled. I've not done much work in this predicate stuff so I may not have context, but this feels like a code smell that could point to something more serious.

core/vm/evm.go
```diff
@@ -220,6 +227,32 @@ func NewEVM(blockCtx BlockContext, txCtx TxContext, statedb StateDB, chainConfig
...
+	if err != nil {
+		log.Error("Unexpected error parsing predicate results",
+			"err", err,
+		)
+		return evm
+	}
+
+	// Because the BlockContext is pass-by-value, this does not cache the
+	// results for future calls to NewEVM.
+	evm.Context.PredicateResults = results
```
ARR4N, line 255: (info) As above, this will probably remain in the registered `libevm` payload for `Header`, which can be accessed via the [`vm.PrecompileEnvironment`](https://pkg.go.dev/github.com/ava-labs/libevm/core/vm#PrecompileEnvironment).

ARR4N (approved):
LGTM re `libevm` compatibility. Although some refactoring will be necessary, it already has to be done and all fields required by this change are accessible.

> [!CAUTION]
> I don't know enough about handling of predicates to give a complete sign off so please consider whether an extra approval is necessary.

core/vm/evm.go
StephenButtolph: header is a package here

ceyonur (approved):

darioush (approved):
this LGTM:
- like the cleanup
- doesn't seem incompatible with libevm

plugin/evm/block.go
StephenButtolph: Yes, exactly

PR #825, opened 2025-02-20
Move `header.ExtraPrefix` calculation

StephenButtolph:
## Why this should be merged

ACP-176 changes the `ExtraPrefix` to be dependent on the `GasUsed` and `ExtDataGasUsed`, in the block. Which is only available after building it. This PR moves the generation of the `ExtraPrefix` to live inside of `FinalizeAndAssemble` rather than at the initial header creation.

## How this works

Moves code around.

## How this was tested

Existing CI.

## Need to be documented?

No.

No.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -161,10 +161,6 @@ func (w *worker) commitNewWork(predicateContext *precompileconfig.PredicateConte
 		gasLimit = core.CalcGasLimit(parent.GasUsed, parent.GasLimit, ap1.GasLimit, ap1.GasLimit)
 	}
 
-	extra, err := header.ExtraPrefix(w.chainConfig, parent, timestamp)
-	if err != nil {
-		return nil, fmt.Errorf("failed to calculate new extra prefix: %w", err)
-	}
 	baseFee, err := header.BaseFee(w.chainConfig, parent, timestamp)
 	if err != nil {
 		return nil, fmt.Errorf("failed to calculate new base fee: %w", err)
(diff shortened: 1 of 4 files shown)


RodrigoVillar (approved):
LGTM

miner/worker.go
```diff
@@ -161,10 +161,6 @@ func (w *worker) commitNewWork(predicateContext *precompileconfig.PredicateConte
 		gasLimit = core.CalcGasLimit(parent.GasUsed, parent.GasLimit, ap1.GasLimit, ap1.GasLimit)
 	}
 
-	extra, err := header.ExtraPrefix(w.chainConfig, parent, timestamp)
```
ceyonur, line 164: Should we also move this since it appends to the extra in header? Not sure why we need to keep it in two different places. Moving this will probably require to pass the env or predicateResults though
https://github.com/ava-labs/coreth/blob/move-predicate-header-parsing/miner/worker.go#L470-L476

StephenButtolph: If we are okay with changing the signature of `FinalizeAndAssemble` then we can do that

ceyonur: I guess we don't have many options. Otherwise it seems we need to create the extra here with gap to be later filled by `FinalizeAndAssemble`

ceyonur: uh I misread that,. we actually prepend in the `FinalizeAndAssemble` it should be fine.

ceyonur (approved):

PR #827, opened 2025-02-21
Update ACP-176 math

StephenButtolph:
## Why this should be merged

- `K` was changed from `43` to `87`.
- Moves an implementation detail out of a godoc comment into the function body
- Fixes a bug where the `Capacity` could exceed `MaxCapacity` when the `MaxCapacity` is decreased.

## How this works

- Updates constant
- Adds bounds check
- Moves comment

## How this was tested

- Updated tests
- Added regression tests

## Need to be documented?

No.

No.


plugin/evm/upgrade/acp176/acp176.go
```diff
@@ -134,15 +134,21 @@ func (s *State) UpdateTargetExcess(desiredTargetExcess gas.Gas) {
 		newTargetPerSecond,
 		previousTargetPerSecond,
 	)
+
+	// Ensure the gas capacity does not exceed the maximum capacity.
+	newMaxCapacity, err := safemath.Mul(targetToMaxCapacity, newTargetPerSecond)
+	if err != nil {
+		newMaxCapacity = math.MaxUint64
+	}
+	s.Gas.Capacity = min(s.Gas.Capacity, newMaxCapacity)
```
alarso16, line 143: Good catch

alarso16 (approved):

plugin/evm/upgrade/acp176/acp176.go
```diff
@@ -180,3 +168,11 @@ func scaleExcess(
 	bigExcess.Div(&bigExcess, &bigTarget)
 	return gas.Gas(bigExcess.Uint64())
 }
+
+func mul[T constraints.Unsigned](a, b T) T {
```
ARR4N, line 172: Perhaps `mulWithUpperBound` or `clippedMultiply`? I dunno, naming is hard, but I think it's worth indicating what it will do on error.

ceyonur: I agree that we can rename this + put a comment

plugin/evm/upgrade/acp176/acp176.go
```diff
@@ -180,3 +168,11 @@ func scaleExcess(
...
+	product, err := math.Mul(a, b)
+	if err != nil {
+		return math.MaxUint[T]()
```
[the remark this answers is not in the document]
StephenButtolph: We talked about this previously, the signature of `Mul` should be changed to return a bool.

If you wanted to make that PR to avalanchego I think it would be a good cleanup.

PR #828, opened 2025-02-24
Merge master branch into libevm branch

qdm12:
Ongoing branch keeping libevm up to date with master.

This branch should be updated regularly with both (and only both):

- `git merge master` to merge commit master into this branch
- `git merge libevm` to merge commit libevm into this branch

We should then `git checkout libevm && git merge --ff-only qdm12/libevm-master-merge && git push` once this PR is approved (Github allows it thank god, and thank ARR4N for pointing this out). We should NOT merge this through the Github UI. This branch should be merged either:

- when the libevm branch is ready to be reviewed for merging it into the master branch
- when changes to the master branch are big enough it's worth updating the libevm branch and recreate another branch+PR like this one

Note to self: last commit hash before last merge: ~210fbee70~ ~7a1f5df8c~ ~0f330d3~ ~9a17233~ 7f104c8


qdm12:
A bug was introduced with [this line](https://github.com/ava-labs/coreth/compare/0cdf88eeb82cf815e1bc492acbfab9a701d63c47...master#diff-689874b30f6f905ce6c47cdefeddcf052b467a7936a981f4b6cf38939e10d924L178) and the libevm changes reverted in https://github.com/ava-labs/coreth/pull/828/commits/1cfe453eb0771242eaf1277facb55a21c7030813

darioush (approved):

core/evm.go
```diff
@@ -168,11 +137,26 @@ func newEVMBlockContext(header *types.Header, chain ChainContext, author *common
...
 }
 
+// NewEVMBlockContextWithPredicateResults creates a new context for use in the
+// EVM with an override for the predicate results. The miner uses this to pass
+// predicate results to the EVM when header.Extra is not fully formed yet.
+func NewEVMBlockContextWithPredicateResults(rules extras.AvalancheRules, header *types.Header, chain ChainContext, author *common.Address, predicateResults []byte) vm.BlockContext {
+	blockCtx := NewEVMBlockContext(header, chain, author)
+	// Note this only sets the block context, which is the hand-off point for
+	// the EVM. The actual header is not modified.
+	blockCtx.Header.Extra = customheader.SetPredicateBytesInExtra(
```
alarso16, line 152: nit: diff from coreth would be lower if you named this `blockContext` instead, but I'm sure it's fine.

alarso16 (approved):
LGTM, I just left one extremely nitpicky comment that you can ignore

plugin/evm/header/block_gas_cost.go
```diff
@@ -11,21 +11,27 @@ import (
...
 	errExtDataGasUsedNil = errors.New("extDataGasUsed is nil")
+	errNoGasUsed         = errors.New("no gas used")
 )
 
 // BlockGasCost calculates the required block gas cost based on the parent
 // header and the timestamp of the new block.
+// Prior to AP4, the returned block gas cost will be nil.
 func BlockGasCost(
 	config *extras.ChainConfig,
 	parent *types.Header,
```
ceyonur, line 29: what's our typical process of using header vs headerExtra? This is pretty much our own code, I think we can require headerExtra here. but certainly not a big issue.

plugin/evm/header/extra.go
```diff
@@ -4,56 +4,147 @@
...
 	"github.com/ava-labs/coreth/params/extras"
+	"github.com/ava-labs/coreth/plugin/evm/upgrade/acp176"
+	"github.com/ava-labs/coreth/plugin/evm/upgrade/ap3"
 )
 
-var errInvalidExtraLength = errors.New("invalid header.Extra length")
+const (
+	// Note: MaximumExtraDataSize has been reduced to 32 in Geth, but is kept the same in Coreth for
+	// backwards compatibility.
+	MaximumExtraDataSize uint64 = 64 // Maximum size extra data may be after Genesis.
```
ceyonur, line 21: how do we plan to cope with those changes in params? Would it make sense to keep these in params package so we wont copy anything there by mistake?

plugin/evm/header/extra.go
```diff
@@ -66,14 +157,49 @@ func VerifyExtra(rules extras.AvalancheRules, extra []byte) error {
...
+	if len(extra) <= offset {
+		return nil
+	}
+	return extra[offset:]
+}
+
+// SetPredicateBytesInExtra sets the predicate result bytes in the header's extra
+// data. If the extra data is not long enough (i.e., an incomplete header.Extra
+// as built in the miner), it is padded with zeros.
+func SetPredicateBytesInExtra(rules extras.AvalancheRules, extra []byte, predicateBytes []byte) []byte {
```
ceyonur, line 192: can we add a UT for this?

plugin/evm/header/extra.go
```diff
@@ -66,14 +157,49 @@ func VerifyExtra(rules extras.AvalancheRules, extra []byte) error {
...
+		offset = acp176.StateSize
+	}
+
+	if len(extra) < offset {
+		// pad extra with zeros
+		extra = append(extra, make([]byte, offset-len(extra))...)
+	}
+	extra = extra[:offset] // truncate extra to the offset
+	extra = append(extra, predicateBytes...)
+	return extra
```
ceyonur, line 204: can we just return `append` here?

plugin/evm/header/extra.go
```diff
@@ -66,14 +157,49 @@ func VerifyExtra(rules extras.AvalancheRules, extra []byte) error {
...
+	offset := ap3.WindowSize
+	if rules.IsFortuna {
+		offset = acp176.StateSize
+	}
+
+	if len(extra) < offset {
+		// pad extra with zeros
+		extra = append(extra, make([]byte, offset-len(extra))...)
+	}
+	extra = extra[:offset] // truncate extra to the offset
```
ceyonur, line 202: I wonder if we should just truncate if len(extra) > offset here (in else above)

params/config.go
```diff
@@ -485,23 +305,8 @@ var (
...
-				BanffBlockTimestamp:             utils.NewUint64(0),
-				CortinaBlockTimestamp:           utils.NewUint64(0),
-				DurangoBlockTimestamp:           utils.NewUint64(0),
-				EtnaTimestamp:                   nil,
-			},
-		})
+		extras.TestDurangoChainConfig,
+	)
 
 	TestEtnaChainConfig = WithExtra(
```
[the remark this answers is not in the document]
darioush: we can't import `params` in `extra`.

params/protocol_params.go
```diff
@@ -33,14 +33,8 @@ import (
...
 
 const (
-	GasLimitBoundDivisor uint64 = 1024               // The bound divisor of the gas limit, used in update calculations.
-	MinGasLimit          uint64 = 5000               // Minimum the gas limit may ever be.
-	MaxGasLimit          uint64 = 0x7fffffffffffffff // Maximum the gas limit (2^63-1).
-	GenesisGasLimit      uint64 = 4712388            // Gas limit of the Genesis block.
-
-	// Note: MaximumExtraDataSize has been reduced to 32 in Geth, but is kept the same in Coreth for
-	// backwards compatibility.
-	MaximumExtraDataSize  uint64 = 64    // Maximum size extra data may be after Genesis.
```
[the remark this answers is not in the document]
darioush: they had to move to avoid importing `params` from the new header packages.

params/hooks_libevm.go
```diff
@@ -109,7 +110,8 @@ func makePrecompile(contract contract.StatefulPrecompiledContract) libevm.Precom
 			panic(err) // Should never happen
 		}
 		var predicateResults *predicate.Results
-		if predicateResultsBytes := predicate.GetPredicateResultBytes(header.Extra); len(predicateResultsBytes) > 0 {
+		rules := GetRulesExtra(env.Rules()).AvalancheRules
+		if predicateResultsBytes := customheader.PredicateBytesFromExtra(rules, header.Extra); len(predicateResultsBytes) > 0 {
 			predicateResults, err = predicate.ParseResults(predicateResultsBytes)
 			if err != nil {
 				panic(err) // Should never happen, as results are already validated in block validation
```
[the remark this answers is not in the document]
darioush: I think it's more correct to panic on a programming error.
but if you think it's better to handle it differently for caution we can do it.

consensus/dummy/consensus.go
```diff
@@ -158,28 +167,26 @@ func (eng *DummyEngine) verifyHeaderGasFields(config *params.ChainConfig, header
 		return fmt.Errorf("expected base fee %d, found %d", expectedBaseFee, header.BaseFee)
 	}
 
-	// Verify BlockGasCost, ExtDataGasUsed not present before AP4
 	headerExtra := types.GetHeaderExtra(header)
```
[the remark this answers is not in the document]
qdm12: `utils.BigEqual(headerExtra.BlockGasCost, expectedBlockGasCost)` uses headerExtra as first agument then expectedBlockGasCost, I would rather keep it as it is:

1. Define argument 1
2. Define argument 2
3. Call function with argument 1, argument 2

plugin/evm/header/extra.go
darioush: we can add a UT to confirm they match the upstream code if we desire. but I don't think there's a way to leave them there based on the import structure we have.

params/extras/config.go
```diff
@@ -14,6 +14,89 @@ import (
...
+
+	TestCortinaChainConfig = copyAndSet(TestBanffChainConfig, func(c *ChainConfig) {
+		c.NetworkUpgrades.CortinaBlockTimestamp = utils.NewUint64(0)
+	})
+
+	TestDurangoChainConfig = copyAndSet(TestCortinaChainConfig, func(c *ChainConfig) {
+		c.NetworkUpgrades.DurangoBlockTimestamp = utils.NewUint64(0)
+	})
+
+	TestEtnaChainConfig = copyAndSet(TestDurangoChainConfig, func(c *ChainConfig) {
```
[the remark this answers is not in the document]
qdm12: That's Darioush who refactored this - we have to keep the params and extras configs separated to avoid a cycle dependency. The copyAndSet format is indeed different than upstream, but I think it's a good idea and shows clearly the difference between each config.

plugin/evm/header/extra.go
qdm12: Actually this constant can be removed and replaced with ap0.MaximumExtraDataSize (working on integrating this in the previous merge by rebasing this)

plugin/evm/header/extra.go
qdm12: Agreed with darioush; on the other hand, given this function is meant to return the predicate result bytes, it's confusing to update extra, hinting we are returning a new extra value. I'll change it to use `predicateResult := extra` from the beginning of the function

params/protocol_params.go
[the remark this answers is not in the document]
qdm12: These were actually moved to packages such as `ap0` in merged commits

plugin/evm/header/extra.go
qdm12: It's the same thing really. I like the else approach you mention a bit better for clarity, but, given how rebasing this branch is mega troublesome (and we want to merge --ff-only it in libevm), I'll skip this unfortunately!

plugin/evm/header/extra.go
qdm12: Done in 21163f1808c07ee81e7eb07e17538971ccef75de

plugin/evm/header/extra.go
qdm12: I have it in a local stash, I'll try applying a bunch of small changes to the libevm branch later!

consensus/dummy/consensus.go
```diff
@@ -416,26 +424,26 @@ func (eng *DummyEngine) FinalizeAndAssemble(chain consensus.ChainHeaderReader, h
...
+	headerExtra.BlockGasCost = customheader.BlockGasCost(
+		configExtra,
+		parent,
+		header.Time,
+	)
 	if configExtra.IsApricotPhase4(header.Time) {
-		headerExtra.ExtDataGasUsed = extDataGasUsed
-		if headerExtra.ExtDataGasUsed == nil {
-			headerExtra.ExtDataGasUsed = new(big.Int).Set(common.Big0)
+		ext := types.GetHeaderExtra(header)
```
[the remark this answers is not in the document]
qdm12: Indeed we can, bundled in my "local stash" given rebasing this branch is near-impossible and don't want to add too-little-refactor-commits to this branch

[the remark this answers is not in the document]
ceyonur: Not sure I understand what you mean. Why do you need to rebase anything?

params/extras/config.go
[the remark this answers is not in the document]
ceyonur: I'm just ultra confused having two `TestFortunaChainConfig` and etc

plugin/evm/header/extra.go
ceyonur: I'm again confused with what you mean by rebasing. I'd like to keep those changes in the same PR if it's changed in the same PR. Alternatively we can have review PRs but IMHO they should be opened before we close base PRs.

PR #829, opened 2025-02-24
remove vmerrs

ceyonur:
## Why this should be merged

The pkg created to prevent a cyclic dependency with core/vm pkg, now we don't have such problem since we can use libevm.

## How this works

Removed vmerrs pkg and imports libevm core/vm 

## How this was tested

CI

## Need to be documented?

No

No


darioush (approved):

ARR4N (approved):

qdm12 (approved):
🔴 ✅

PR #830, opened 2025-02-24
WIP: Libevm upstream sync with pivot change

alarso16:
There's a few things that I should mention that are mostly related to structure.
- The `downloader.go` logic is very similar to that of the `downloader.go` in geth. The `SnapSync()` method essentially `processSnapSyncContent`. I removed a lot of the (seemingly) excessive logic here, but it seems to successfully move the pivot with rather rare errors.
- Prior to this PR, the state trie sync and atomic trie sync could be done parallelly or in either order. Now, it is changed to  completely finish the atomic trie sync before beginning the snap sync, because we will perform atomic operations synchronously throughout the state sync. However, we could likely still do these parallelly. A channel was added to notify when atomic sync is completely, and we can thus start executing atomic operations as normal.
- We cannot postpone verify/accept operations on atomic operations until the pivot, because this can cause a deadlock with the P-Chain in avalancego. Because of this, we perform atomic operations synchronously, and then after the sync is complete, we perform the non-atomic operations in order as they are buffered to catch up the state trie to the atomic trie. 
- Passing the lock from `stateSyncerclient` to `Downloader` is necessary because we need to block from queueing blocks once we finish the state sync, and then once the queue is cleared, we release the lock, allowing blocks to be verified as normal.
- The weird channels in `Downloader` weren't my idea - copied from geth with minimal revision. I'm not married to it, if you have a more readable suggestion.

My next steps:
1. There are likely some issues around exactly how atomic txs are handled during the final parts of the sync, and testing on mainnet (or testing multiple times on fuji)
2. WIP resuming a sync after canceled by SIGINT or something else.
3. The server must run in non-pruning mode for this to work, since the pivot can be too far back to respond to requests. A more permanent fix would be increasing the `TipToBuffer` and adjusting the pivot interval to find a "sweet spot".
4. None of the FUpgrade is here, so I need to rebase (probably very carefully) and retest to be able to work on fuji again.


plugin/evm/statesync/queue.go
```diff
@@ -0,0 +1,92 @@
+// (c) 2019-2020, Ava Labs, Inc.
```
darioush, line 1: nit: please use the 2 line copyright header as exampled in plugin/evm/vm.go

plugin/evm/block.go
```diff
@@ -223,7 +236,20 @@ func (b *Block) handlePrecompileAccept(rules extras.Rules) error {
...
+func (b *Block) Reject(ctx context.Context) error {
+
+	if b.vm.StateSyncClient.AsyncReceive() {
+		log.Debug("Queueing block for later reject", "block", b.ID(), "height", b.Height())
+		return b.vm.StateSyncClient.QueueRejectBlock(b)
+	}
+	return b.reject(ctx)
+}
+
+func (b *Block) RejectDuringSync(ctx context.Context) error {
```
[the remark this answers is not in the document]
alarso16: It does get called occasionally - another one is `SetPreference` which I didn't really expect. I added it only because it happened

plugin/evm/statesync/queue.go
```diff
@@ -0,0 +1,92 @@
...
+
+type Queue[K Executable] struct {
+	buffer  []*K
+	l       sync.RWMutex
+	nextPos int
+	compare func(*K, *K) int
+	closed  bool
+}
+
+func NewQueue[K Executable](size int, compare func(*K, *K) int) *Queue[K] {
```
darioush, line 29: not sure Executable needs to be a type parameter here? we can have buffer []Executable

plugin/evm/syncervm_client.go
```diff
@@ -441,3 +555,202 @@ func (client *stateSyncerClient) updateVMMarkers() error {
...
+		started: make(chan struct{}),
+	}
+}
+
+// run starts the task assignment and response processing loop, blocking until
+// it finishes, and finally notifying any goroutines waiting for the loop to
+// finish.
+func (s *stateSync) run() {
+	close(s.started)
+	log.Info("Starting new sync")
```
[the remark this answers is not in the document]
alarso16: C&P from `eth/downloader/statesync.go`, but yeah definitely would be more helpful

plugin/evm/statesync/queue.go
```diff
@@ -0,0 +1,92 @@
...
+	q.nextPos++
+
+	return nil
+}
+
+func (q *Queue[K]) Flush(max *K, close bool) error {
+	q.l.Lock()
+	defer q.l.Unlock()
+
+	newBuffer := make([]*K, len(q.buffer))
```
darioush, line 54: seems this can be done without allocating a new queue.
1. find the position to evict,
2. evict all stuff after that,
3. set nextPos to the found position

plugin/evm/syncervm_client.go
```diff
@@ -97,6 +106,12 @@ type StateSyncClient interface {
 	ClearOngoingSummary() error
 	Shutdown() error
 	Error() error
+
+	// Methods to try to enable dynamic state sync
```
darioush, line 110: nit: can avoid the "try" (tentative) language so the draft work is closer to the complete work

plugin/evm/syncervm_client.go
```diff
@@ -108,6 +123,71 @@ type Syncer interface {
...
+	return client.useUpstream && client.syncing.Get() && client.dl != nil && !client.dl.execQueue.Closed()
+}
+
+func (client *stateSyncerClient) QueueVerifyBlock(b *Block) error {
+	if !client.AsyncReceive() {
+		return fmt.Errorf("cannot queue block when not using upstream syncing")
+	}
+
+	verify := func(a *Block) error {
+		return a.VerifyDuringSync(context.TODO()) // is this right?
```
darioush, line 138: this is not gonna work since during sync the state is incomplete. let's chat about it.

plugin/evm/syncervm_client.go
```diff
@@ -441,3 +555,202 @@ func (client *stateSyncerClient) updateVMMarkers() error {
...
+	for {
+		select {
+		// If stateSync is ended, clear queue and return
+		// If err, just return so we can see it
+		case <-sync.done:
+			if sync.err != nil {
+				return sync.err
+			}
+			return d.execQueue.Flush(nil, true) // might need to provide a cancle channel
+		case np := <-d.newPivot:
```
darioush, line 663: nit: let's use better names like "newPivot" or "pivot"

plugin/evm/syncervm_client.go
```diff
@@ -441,3 +555,202 @@ func (client *stateSyncerClient) updateVMMarkers() error {
...
+		}
+	}
+}
+
+// processSnapSyncContent takes fetch results from the queue and writes them to the
+// database. It also controls the synchronisation of state nodes of the pivot block.
+func (d *downloader) SnapSync() error {
+	// Start syncing state of the reported head block. This should get us most of
+	// the state of the pivot block.
+	d.pivotLock.RLock()
```
darioush, line 643: not sure this lock is needed, we can always send the pivots over the channel and avoid this lock.
it's better to only use locks when strictly needed and prefer channels

plugin/evm/syncervm_client.go
```diff
@@ -441,3 +555,202 @@ func (client *stateSyncerClient) updateVMMarkers() error {
...
+	// Create the state sync
+	s := newStateSync(d, root)
+	select {
+	case d.stateSyncStart <- s:
+		// If we tell the statesync to restart with a new root, we also need
+		// to wait for it to actually also start -- when old requests have timed
+		// out or been delivered
+		<-s.started
+	case <-d.quitCh:
+		s.err = fmt.Errorf("errCancelStateFetch") //errCancelStateFetch
```
darioush, line 684: nit: use errors.New if there is no "%" directive in the err

plugin/evm/statesync/queue.go
alarso16: I'll wait to make changes to this until I know whether I can just delete it or not.

plugin/evm/syncervm_client.go
```diff
@@ -441,3 +555,202 @@ func (client *stateSyncerClient) updateVMMarkers() error {
...
+
+	for {
+		select {
+		// If stateSync is ended, clear queue and return
+		// If err, just return so we can see it
+		case <-sync.done:
+			if sync.err != nil {
+				return sync.err
+			}
+			return d.execQueue.Flush(nil, true) // might need to provide a cancle channel
```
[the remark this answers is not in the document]
alarso16: Deleted this line anyway

plugin/evm/syncervm_client.go
alarso16: I don't think it is. I was worried about changes bootstrapping sending blocks and triggering several pivots right at the start, but I addressed this differently.

PR #832, opened 2025-02-25
fix(core/evm): set difficulty post shanghai to 0

darioush:

### DIFF
--- a/core/evm.go
+++ b/core/evm.go
@@ -54,7 +54,7 @@ type hooks struct{}
 // It allows for the modification of the EVM arguments before the EVM is created.
 // Specifically, we set Random to be the same as Difficulty since Shanghai.
 // This allows using the same jump table as upstream.
-// Then we set Difficulty to nil as it is post Merge in upstream.
+// Then we set Difficulty to 0 as it is post Merge in upstream.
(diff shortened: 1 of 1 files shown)


core/evm.go
```diff
@@ -65,7 +65,7 @@ func (hooks) OverrideNewEVMArgs(args *vm.NewEVMArgs) *vm.NewEVMArgs {
 	if rules.IsShanghai {
 		args.BlockContext.Random = new(common.Hash)
 		args.BlockContext.Random.SetBytes(args.BlockContext.Difficulty.Bytes())
-		args.BlockContext.Difficulty = nil
+		args.BlockContext.Difficulty = new(big.Int)
```
ceyonur, line 68: aren't we expecting this to be 1 [here](https://github.com/ava-labs/coreth/blob/libevm/plugin/evm/block_verification.go#L100)?

darioush: in the header yes, but not in the block context, which is the handoff point to the EVM interpreter.
The interpreter will start using the "Random" opcode, which will read the value of "Random"
The value of Difficulty will actually be unused (Which is how this was previously working), but I still figure it's better to set it the same way as upstream.

ARR4N (approved):

PR #839, opened 2025-02-26
Cleanup `SyntacticVerify`

StephenButtolph:
## Why this should be merged

This removes code that is not needed (and would complicate the introduction of ACP-176 if it were kept.

## How this works

Replaces equality checks with existing helpers and removes unnecessary checks.

## How this was tested

existing CI.

## Need to be documented?

No.

No.

### DIFF
--- a/plugin/evm/block_verification.go
+++ b/plugin/evm/block_verification.go
@@ -180,16 +180,8 @@ func (v blockValidator) SyntacticVerify(b *Block, rules params.Rules) error {
 
 	// If we are in ApricotPhase4, ensure that ExtDataGasUsed is populated correctly.
 	if rules.IsApricotPhase4 {
-		// Make sure ExtDataGasUsed is not nil and correct
-		if ethHeader.ExtDataGasUsed == nil {
-			return errNilExtDataGasUsedApricotPhase4
-		}
 		if rules.IsApricotPhase5 {
 			if !utils.BigLessOrEqualUint64(ethHeader.ExtDataGasUsed, ap5.AtomicGasLimit) {
-				return fmt.Errorf("too large extDataGasUsed: %d", ethHeader.ExtDataGasUsed)
-			}
-		} else {
-			if !ethHeader.ExtDataGasUsed.IsUint64() {
 				return fmt.Errorf("too large extDataGasUsed: %d", ethHeader.ExtDataGasUsed)
 			}
 		}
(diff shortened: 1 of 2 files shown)


plugin/evm/block_verification.go
```diff
@@ -180,18 +180,10 @@ func (v blockValidator) SyntacticVerify(b *Block, rules params.Rules) error {
 
 	// If we are in ApricotPhase4, ensure that ExtDataGasUsed is populated correctly.
 	if rules.IsApricotPhase4 {
-		// Make sure ExtDataGasUsed is not nil and correct
-		if ethHeader.ExtDataGasUsed == nil {
```
StephenButtolph, line 184: `utils.BigLessOrEqualUint64` and `utils.BigEqualUint64` both handle `nil` values internally.

plugin/evm/block_verification.go
```diff
@@ -180,18 +180,10 @@ func (v blockValidator) SyntacticVerify(b *Block, rules params.Rules) error {
...
-			return errNilExtDataGasUsedApricotPhase4
-		}
 		if rules.IsApricotPhase5 {
 			if !utils.BigLessOrEqualUint64(ethHeader.ExtDataGasUsed, ap5.AtomicGasLimit) {
 				return fmt.Errorf("too large extDataGasUsed: %d", ethHeader.ExtDataGasUsed)
 			}
-		} else {
-			if !ethHeader.ExtDataGasUsed.IsUint64() {
-				return fmt.Errorf("too large extDataGasUsed: %d", ethHeader.ExtDataGasUsed)
-			}
```
StephenButtolph, line 194: This is checked in `utils.BigEqualUint64` (and was checked previously by the existing equality check)

plugin/evm/block_verification.go
```diff
@@ -180,18 +180,10 @@ func (v blockValidator) SyntacticVerify(b *Block, rules params.Rules) error {
 
 	// If we are in ApricotPhase4, ensure that ExtDataGasUsed is populated correctly.
 	if rules.IsApricotPhase4 {
```
ceyonur, line 182: I think we're removing checks from previous upgrades (ap4). Since ap4 is far behind now we should not see any blocks prior ap5, but just wanted to point out.

StephenButtolph: What checks do you think this is removing?

ceyonur: ah ok my bad it seems we handle all these cases in switch case below.

ceyonur (approved):

PR #843, opened 2025-02-26
Filter by total gas rather than GasUsed

StephenButtolph:
## Why this should be merged

Currently the gas price oracle doesn't account for extraDataGasUsed. This PR includes that gas and adds a unit test for it.

## How this works

Sums the gas.

## How this was tested

unit test

## Need to be documented?

no.

no.

### DIFF
--- a/eth/gasprice/gasprice_test.go
+++ b/eth/gasprice/gasprice_test.go
@@ -395,3 +395,13 @@ func TestSuggestTipCapMaxBlocksSecondsLookback(t *testing.T) {
 		expectedTip:     big.NewInt(92_212_529_424),
 	}, timeCrunchOracleConfig())
 }
+
+func TestSuggestTipCapIncludesExtraDataGas(t *testing.T) {
+	applyGasPriceTest(t, suggestTipCapTest{
+		chainConfig:     params.TestChainConfig,
+		numBlocks:       20,
+		extDataGasUsage: DefaultMinGasUsed,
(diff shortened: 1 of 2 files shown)


darioush (approved):

eth/gasprice/gasprice_test.go
```diff
@@ -395,3 +395,13 @@ func TestSuggestTipCapMaxBlocksSecondsLookback(t *testing.T) {
...
 	}, timeCrunchOracleConfig())
 }
+
+func TestSuggestTipCapIncludesExtraDataGas(t *testing.T) {
+	applyGasPriceTest(t, suggestTipCapTest{
+		chainConfig:     params.TestChainConfig,
+		numBlocks:       20,
+		extDataGasUsage: DefaultMinGasUsed,
+		genBlock:        testGenBlock(t, 100_000, 1),
+		expectedTip:     big.NewInt(83_603_561_239),
```
ARR4N, line 405: Where does this number come from?

image

StephenButtolph: So, the tip is equal to: `blockGasCost` * `baseFee` / `totalGasUsed`... But I doubt anyone has actually calculated the expected values for these tests.

The number ends up coming from:
1. The base fee changing as blocks are produced
2. The block gas cost increasing as blocks are produced

(the `totalGasUsed` is constant throughout the test)

I added a couple comments saying what the test is actually doing.

ARR4N (approved):

PR #844, opened 2025-02-27
patch up master merge (takes 94dca86) + refactor customheaders

darioush:
may want to review & merge 1st commit separately


core/evm.go
```diff
@@ -163,11 +135,25 @@ func newEVMBlockContext(header *types.Header, chain ChainContext, author *common
...
 
+// NewEVMBlockContextWithPredicateResults creates a new context for use in the
+// EVM with an override for the predicate results. The miner uses this to pass
+// predicate results to the EVM when header.Extra is not fully formed yet.
+func NewEVMBlockContextWithPredicateResults(header *types.Header, chain ChainContext, author *common.Address, predicateResults []byte) vm.BlockContext {
+	blockCtx := NewEVMBlockContext(header, chain, author)
+	// Note this only sets the block context, which is the hand-off point for
+	// the EVM. The actual header is not modified.
+	blockCtx.Header.Extra = customheader.SetPredicateBytesInExtra(
+		blockCtx.Header.Extra,
```
darioush, line 151: I think this might need to be slices.Clone(header.Extra) or something instead.. just to avoid altering header.Extra out of caution.

plugin/evm/header/extra.go
```diff
@@ -66,14 +71,40 @@ func VerifyExtra(rules extras.AvalancheRules, extra []byte) error {
...
+	if len(extra) <= FeeWindowSize {
+		return nil
+	}
+	return extra[FeeWindowSize:]
+}
+
+// SetPredicateBytesInExtra sets the predicate result bytes in the header's extra
+// data. If the extra data is not long enough (i.e., an incomplete header.Extra
+// as built in the miner), it is padded with zeros.
+func SetPredicateBytesInExtra(extra []byte, predicateBytes []byte) []byte {
```
darioush, line 102: this is a little jank but it's the cost of the way we're passing the predicate bytes into libevm. unless you can come up with a better idea I think it's fine for now.

params/extras/config.go
```diff
@@ -14,6 +14,88 @@ import (
...
+			DurangoBlockTimestamp:           utils.NewUint64(0),
+			EtnaTimestamp:                   utils.NewUint64(0),
+		},
+	}
+
+	copyAndSet = func(c *ChainConfig, set func(*ChainConfig)) *ChainConfig {
+		newConfig := *c
+		set(&newConfig)
+		return &newConfig
+	}
```
darioush, line 40: I think this style reads nicer but feel free to put it back

plugin/evm/header/extra.go
```diff
@@ -8,10 +8,15 @@ import (
...
 
 	"github.com/ava-labs/coreth/core/types"
-	"github.com/ava-labs/coreth/params"
 	"github.com/ava-labs/coreth/params/extras"
 )
 
+const (
+	// Note: MaximumExtraDataSize has been reduced to 32 in Geth, but is kept the same in Coreth for
+	// backwards compatibility.
+	MaximumExtraDataSize uint64 = 64 // Maximum size extra data may be after Genesis.
```
StephenButtolph, line 17:
```suggestion
	MaximumExtraDataSize = 64 // Maximum size extra data may be after Genesis.
```

I think generally it is better if we let the compiler infer the types. This would allow us to remove the explicit cast lower in the file.

StephenButtolph, line 17: Does this actually make sense to be defined in the `header` package? `header.MaximumExtraDataSize` seems to imply that the `len(ExtraData) <= header.MaximumExtraDataSize` - but that is only true for legacy AP0/launch blocks. Should this be defined in the `ap0` package?

plugin/evm/header/gas_limit.go
```diff
@@ -9,12 +9,17 @@ import (
...
 	"github.com/ava-labs/coreth/params/extras"
 	"github.com/ava-labs/coreth/plugin/evm/upgrade/ap1"
 	"github.com/ava-labs/coreth/plugin/evm/upgrade/cortina"
 )
 
+const (
+	MinGasLimit          uint64 = 5000               // Minimum the gas limit may ever be.
+	MaxGasLimit          uint64 = 0x7fffffffffffffff // Maximum the gas limit (2^63-1).
+	GasLimitBoundDivisor uint64 = 1024               // The bound divisor of the gas limit, used in update calculations.
+)
```
StephenButtolph, line 21: Similar comment as the other file, these aren't constants that apply generally to the header, they are specific to `ap0`

Also:
```suggestion
const (
	MinGasLimit          = 5000               // Minimum the gas limit may ever be.
	MaxGasLimit          = 0x7fffffffffffffff // Maximum the gas limit (2^63-1).
	GasLimitBoundDivisor = 1024               // The bound divisor of the gas limit, used in update calculations.
)
```

darioush: moving all them to ap0

plugin/evm/header/extra.go
qdm12: Fine with me!

params/extras/config.go
qdm12: Fine with me, I just moved it to a "normal" function instead of being a global var which is a bit strange-ish.

qdm12:
I incorporated the changes in the merge commits of https://github.com/ava-labs/coreth/pull/828 given we need to merge commit this branch into libevm, to keep a cleaner history. I managed to split it in minimal requirements per merge commit to make CI pass, but overall it's all integrated in there.

qdm12:
Closing this in favor of #828

PR #849, opened 2025-03-01
move ap4 block gas cost check to header

ceyonur:
## Why this should be merged

moves AP4 gas limit calc to header (return nil)

## How this works

BlockGasCost return type changed to big.Int (from uint64), and returns nil for !AP4 

## How this was tested

added UT

## Need to be documented?

no

no

### DIFF
--- a/consensus/dummy/consensus.go
+++ b/consensus/dummy/consensus.go
@@ -178,9 +178,6 @@ func verifyHeaderGasFields(config *params.ChainConfig, header *types.Header, par
 
 	// Verify BlockGasCost, ExtDataGasUsed not present before AP4
 	if !config.IsApricotPhase4(header.Time) {
-		if header.BlockGasCost != nil {
-			return fmt.Errorf("invalid blockGasCost before fork: have %d, want <nil>", header.BlockGasCost)
-		}
 		if header.ExtDataGasUsed != nil {
 			return fmt.Errorf("invalid extDataGasUsed before fork: have %d, want <nil>", header.ExtDataGasUsed)
 		}
--- a/consensus/dummy/consensus.go
+++ b/consensus/dummy/consensus.go
@@ -176,7 +176,7 @@ func verifyHeaderGasFields(config *params.ChainConfig, header *types.Header, par
 		return fmt.Errorf("invalid block gas cost: have %d, want %d", header.BlockGasCost, expectedBlockGasCost)
 	}
 
-	// Verify BlockGasCost, ExtDataGasUsed not present before AP4
+	// Verify ExtDataGasUsed not present before AP4
 	if !config.IsApricotPhase4(header.Time) {
 		if header.ExtDataGasUsed != nil {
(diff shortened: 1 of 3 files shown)


plugin/evm/header/block_gas_cost_test.go
```diff
@@ -19,40 +19,54 @@ import (
 func TestBlockGasCost(t *testing.T) {
 	tests := []struct {
 		name         string
+		beforeAP4    bool
```
darioush, line 22: nit: could we do ap4 or ap4Timestamp instead?

darioush (approved):

consensus/dummy/consensus.go
```diff
@@ -166,6 +166,16 @@ func verifyHeaderGasFields(config *params.ChainConfig, header *types.Header, par
...
+		parent,
+		header.Time,
+	)
+	if !utils.BigEqual(header.BlockGasCost, expectedBlockGasCost) {
+		return fmt.Errorf("invalid block gas cost: have %d, want %d", header.BlockGasCost, expectedBlockGasCost)
+	}
+
 	// Verify BlockGasCost, ExtDataGasUsed not present before AP4
 	if !config.IsApricotPhase4(header.Time) {
 		if header.BlockGasCost != nil {
```
StephenButtolph, line 181: I think this can be removed right?

plugin/evm/header/block_gas_cost.go
```diff
@@ -27,7 +27,10 @@ func BlockGasCost(
 	config *params.ChainConfig,
 	parent *types.Header,
 	timestamp uint64,
-) uint64 {
+) *big.Int {
```
StephenButtolph, line 30: Can we document this function like `BaseFee`?

```
// Prior to AP4, the returned block gas cost will be nil.
```

plugin/evm/header/block_gas_cost_test.go
ceyonur: I did not want to increase the diff for this one, but sure.

consensus/dummy/consensus.go
ceyonur: yes

plugin/evm/header/block_gas_cost.go
ceyonur: done

consensus/dummy/consensus.go
```diff
@@ -166,27 +166,24 @@ func verifyHeaderGasFields(config *params.ChainConfig, header *types.Header, par
...
 		config,
 		parent,
 		header.Time,
 	)
-	if !utils.BigEqualUint64(header.BlockGasCost, expectedBlockGasCost) {
+	if !utils.BigEqual(header.BlockGasCost, expectedBlockGasCost) {
 		return fmt.Errorf("invalid block gas cost: have %d, want %d", header.BlockGasCost, expectedBlockGasCost)
 	}
 
+	// Verify BlockGasCost, ExtDataGasUsed not present before AP4
```
StephenButtolph, line 179: Nit: this doesn't verify the BlockGasCost anymore

StephenButtolph (approved):

PR #852, opened 2025-03-03
chore(accounts/keystore): return nil when error has already been checked

rustco:
## Why this should be merged

The logic of err not being nil has been processed and returned before, so err must be nil here.


qdm12 (approved):

qdm12:
Good point @darioush it is indeed upstream code.
@rustco please patch this upstream at:

- https://github.com/ethereum/go-ethereum/blob/d219e9b005358f96d85f4eb66e1418c605b2bb0f/accounts/keystore/presale.go#L113
- https://github.com/ethereum/go-ethereum/blob/d219e9b005358f96d85f4eb66e1418c605b2bb0f/accounts/keystore/presale.go#L128

if you want this to be bundled this here, thanks! Closing this.

PR #855, opened 2025-03-05
refactor(core/types): add `WithHeaderExtra` helper function

qdm12:
## Why this should be merged

Add `WithHeaderExtra` in order to simplify tests.

I did not add `WithBlockExtra` since `SetBlockExtra` is only used in two places and it would make it more confusing to read than how it is now.

## How this was tested

Existing CI passing

## Need to be documented?

no

no


darioush (approved):

core/types/block_ext_test.go
```diff
@@ -370,24 +373,19 @@ func TestNewBlockWithExtData(t *testing.T) {
 
 	tests := []struct {
 		name      string
-		header    func() *Header
```
ARR4N, line 373: I always hate when I have to introduce a constructor into a table test so it's nice to see this disappearing :)

core/types/block_ext_test.go
```diff
@@ -370,24 +373,19 @@ func TestNewBlockWithExtData(t *testing.T) {
...
 		wantBlock func() *Block
 	}{
 		{
-			name: "empty",
-			header: func() *Header {
-				header := &Header{}
-				SetHeaderExtra(header, &HeaderExtra{})
-				return header
-			},
+			name:   "empty",
```
ARR4N, line 385: (optional) You could add another case where the header is a raw `&Header{}` to demonstrate that the getter will return a non-nil `&HeaderExtra{}`. Not sure of the utility other than someone reviewing the test learning that this is a valid use case (although there's definitely value in that).

ARR4N (approved):

core/types/block_ext_test.go
qdm12: Done!

Issue #833, opened 2025-02-25
core/types: pin code generators versions when alias types are supported


qdm12:
We use two code generators to generate codec functions for JSON and RLP codecs:

1. `github.com/fjl/gencodec` for JSON
2. `github.com/ava-labs/libevm/rlp/rlpgen` for RLP (code forked from geth)

The support for type aliases is ongoing as follows:

1. https://github.com/fjl/gencodec/pull/25 (need to add tests)
2. https://github.com/ava-labs/libevm/pull/154 which requires go 1.22 so this is blocked by a Geth sync

Once these two above are done, we should:

1. Upgrade libevm to the version containing this new feature
2. Pin fjl/gencodec version to the version containing this new feature, using `tools.go` + `go mod tidy`
3. Add a CI check to ensure these generated files are kept in sync (similar to how it is done with mockgen)

ARR4N:
https://github.com/ava-labs/libevm/pull/154 isn't blocked by a Geth sync; it's unblocked by https://github.com/ava-labs/libevm/pull/155 and https://github.com/ava-labs/libevm/pull/156.

qdm12:
Blocked by https://github.com/fjl/gencodec/pull/26

qdm12:
> [ava-labs/libevm#154](https://github.com/ava-labs/libevm/pull/154) isn't blocked by a Geth sync; it's unblocked by [ava-labs/libevm#155](https://github.com/ava-labs/libevm/pull/155) and [ava-labs/libevm#156](https://github.com/ava-labs/libevm/pull/156).

Awesome! 💯 Should we then do another libevm release soon-ish so I can incorporate this in our coreth libevm branch? 🤔

PR #856, opened 2025-03-05
chore(ci): check gencodec generated files are up to date

qdm12:
## Why this should be merged

- fix generation of code using type aliases and Go 1.23
- reproducible generation locally (pin gencodec version)
- keep generated Go files up to date with other source code
- Partly addresses #833 

## How this works

- Pin fjl/codec version using `tools.go`
- Generate missing gencodec files
- Add CI step to remove generated files and re-generate them, and check there is no git difference

## How this was tested

CI / test job passing with new step included

## Need to be documented?

No

No


.github/workflows/ci.yml
ARR4N, line 1: Why not just have a single `go generate ./...` and git check instead of splitting them by the tool being used?

ARR4N: In libevm we do [something very basic](https://github.com/ava-labs/libevm/blob/main/.github%2Fworkflows%2Fgo.yml#L56-L71). Are we missing something there, are we overcomplicating it here, or somewhere in between?

qdm12: Some go generate commands don't work for now (libevm's rlp code generator)

ARR4N (approved):

darioush (approved):

PR #859, opened 2025-03-05
chore(scripts): automated avalanchego version detection if unset

maru-ava:
## Why this should be merged

Previously it was necessary to manually maintain the avalanchego version in versions.sh. Now the version is automatically detected via go modules.

## How this works

 - updates versions.sh to automatically determine the version/SHA from go module

## How this was tested

- CI
- Manually checked the version used matched the go module version when invoking `bash -x scripts/build.sh`

## Need to be documented?

N/A

N/A

### DIFF
--- a/scripts/versions.sh
+++ b/scripts/versions.sh
@@ -5,5 +5,22 @@
 
 set -euo pipefail
 
-# Don't export them as they're used in the context of other calls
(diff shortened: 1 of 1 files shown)


qdm12 (approved):
Awesome!

PR #860, opened 2025-03-06
fix(metrics): avoid conflicts between `coreth` and `libevm`

ARR4N:
## Why this should be merged

`geth` (and hence `libevm`) global metrics tend to be created with `NewRegistered*()`, which assumes that there are no name conflicts on the same `Registry`. If there are then the second call fails silently to register the new metric.

## How this works

Except for in `warp` (where there won't be any conflicts) metric registration is changed globally to use `GetOrRegister*()` instead of `NewRegistered*()`. With GNU tools (not the Mac budge replacements):

```shell
find -iname '*.go' -not -iwholename '*/warp/*' | xargs sed -i 's|metrics.NewRegistered|metrics.GetOrRegister|g'
```

All modified files that don't already import their `libevm` counterparts now blank (`_`) import them to force the initialisation order. The `libevm` construction with `NewRegistered*()` now always comes before the respective `GetOrRegister*()`.

This doesn't account for mismatched metric types between `coreth` and `libevm`, which is beyond the scope of this PR. The override mechanism of ava-labs/libevm#159 is likely the best approach, but it belongs in the `coreth` code IMO.

## How this was tested

Not tested.

## Need to be documented?

No

No


core/blockchain_ext.go
```diff
@@ -0,0 +1,23 @@
+// (c) 2025, Ava Labs, Inc.
```
darioush, line 1: nit: let's use the 2 line header format for consistency

core/main_test.go
```diff
@@ -15,6 +15,7 @@ func TestMain(m *testing.M) {
 	opts := []goleak.Option{
 		// No good way to shut down these goroutines:
 		goleak.IgnoreTopFunction("github.com/ava-labs/coreth/core/state/snapshot.(*diskLayer).generate"),
+		goleak.IgnoreTopFunction("github.com/ava-labs/libevm/core.(*txSenderCacher).cache"),
```
darioush, line 18: importing this actually starts a bunch of goroutines. kind of silly but not sure what we can do about it.

triedb/hashdb/database.go
```diff
@@ -46,42 +46,45 @@ import (
...
-	memcacheGCTimeTimer  = metrics.NewRegisteredResettingTimer("hashdb/memcache/gc/time", nil)
-	memcacheGCNodesMeter = metrics.NewRegisteredMeter("hashdb/memcache/gc/nodes", nil)
-	memcacheGCBytesMeter = metrics.NewRegisteredMeter("hashdb/memcache/gc/bytes", nil)
-
-	memcacheCommitMeter         = metrics.NewRegisteredMeter("hashdb/memcache/commit/count", nil)
-	memcacheCommitTimeTimer     = metrics.NewRegisteredResettingTimer("hashdb/memcache/commit/time", nil)
-	memcacheCommitLockTimeTimer = metrics.NewRegisteredResettingTimer("hashdb/memcache/commit/locktime", nil)
-	memcacheCommitNodesMeter    = metrics.NewRegisteredMeter("hashdb/memcache/commit/nodes", nil)
-	memcacheCommitBytesMeter    = metrics.NewRegisteredMeter("hashdb/memcache/commit/bytes", nil)
+	memcacheCleanHitMeter   = metrics.GetOrRegisterMeter("hashdb/memcache/clean/hit", nil)
```
darioush, line 59: can we add a comment to the files as a guide for whoever is doing merge conflict resolution?

darioush (approved):

core/blockchain_ext.go
```diff
@@ -0,0 +1,23 @@
...
+import "github.com/ava-labs/libevm/metrics"
+
+// getOrOverrideAsRegisteredCounter searches for a metric already registered
+// with`name`. If a metric is found and it is a Counter, it is returned. If a
+// metric is found and it is not a Counter, it is unregistered and replaced with
+// a new registered Counter. If no metric is found, a new Counter is constructed
+// and registered.
+//
+// This is necessary for a metric defined in libevm with the same name but a
+// different type to what we expect.
```
qdm12, line 13: nit

```suggestion
// getOrOverrideAsRegisteredCounter searches for a metric already registered
// with`name`. If a metric is found and it is a [metrics.Counter], it is returned. If a
// metric is found and it is not a [metrics.Counter], it is unregistered and replaced with
// a new registered [metrics.Counter]. If no metric is found, a new [metrics.Counter] is constructed
// and registered.
//
// This is necessary for a metric defined in libevm with the same name but a
// different type than a [metrics.Counter].
```

core/blockchain_ext.go
```diff
@@ -0,0 +1,23 @@
...
+func getOrOverrideAsRegisteredCounter(name string, r metrics.Registry) metrics.Counter {
+	if r == nil {
+		r = metrics.DefaultRegistry
+	}
+	if c, ok := r.GetOrRegister(name, metrics.NewCounter).(metrics.Counter); ok {
```
qdm12, line 18: nit this is a bit of a long line, we could split it as

```suggestion
	m := r.GetOrRegister(name, metrics.NewCounter)
	if c, ok := m.(metrics.Counter); ok {
```

qdm12 (approved):

core/main_test.go
ARR4N: Yeah this isn't something we need to manage here. Maybe in `libevm` if we think it's an issue, but ideally in `geth` (as a low priority).

core/blockchain_ext.go
ARR4N: I've split it but with a type switch. The call to `Unregister()` always made me a little uncomfortable so now the `default` branch gives a natural place for an explanatory comment.

PR #861, opened 2025-03-06
Fix warp tests

qdm12:


qdm12:
Not needed because `arr4n/metrics-avoid-conflicts` does not change metric registration in the warp/ packages

PR #862, opened 2025-03-07
chore(all): final small changes to the libevm branch

qdm12:
## Why this should be merged

Final changes that will be squashed into the libevm branch, suggested from the PR review of #820 

This also notably contains:

- suggested changes from the feedback on #828 
- add test to check PredicateBytesFromExtra is the inverse of SetPredicateBytesInExtra
- Cherry pick 2773bbb7d6ee2616600085d473ff81de0330a825

Let me know if you want me to split this PR, that's no problem as well.

## How this works

- Each line changed has a corresponding resolved conversation on #820 which you can check by finding the same file/line number in the files changed
- Commits with `revert` means the code is reverted to how it is on the master branch because the change is actually not needed.

You can compare this branch to master [here](https://github.com/ava-labs/coreth/compare/master...qdm12/libevm-last-changes) as well, for example to check on reversion changes.


consensus/dummy/consensus.go
```diff
@@ -439,10 +439,9 @@ func (eng *DummyEngine) FinalizeAndAssemble(chain consensus.ChainHeaderReader, h
 		header.Time,
 	)
 	if configExtra.IsApricotPhase4(header.Time) {
-		ext := types.GetHeaderExtra(header)
-		ext.ExtDataGasUsed = extDataGasUsed
-		if ext.ExtDataGasUsed == nil {
-			ext.ExtDataGasUsed = new(big.Int).Set(common.Big0)
+		headerExtra.ExtDataGasUsed = extDataGasUsed
+		if headerExtra.ExtDataGasUsed == nil {
+			headerExtra.ExtDataGasUsed = new(big.Int).Set(common.Big0)
```
ARR4N, line 444:
```suggestion
			headerExtra.ExtDataGasUsed = new(big.Int)
```

Or `= big.NewInt(0)` but the other way is unnecessarily verbose.

alarso16 (approved):
lgtm

darioush (approved):

accounts/abi/bind/util_test.go
```diff
@@ -125,10 +125,7 @@ func TestWaitDeployedCornerCases(t *testing.T) {
 	// Create a transaction to an account.
 	code := "6060604052600a8060106000396000f360606040526008565b00"
 	tx := types.NewTransaction(0, common.HexToAddress("0x01"), big.NewInt(0), 3000000, gasPrice, common.FromHex(code))
-	tx, err := types.SignTx(tx, types.LatestSignerForChainID(big.NewInt(1337)), testKey)
-	if err != nil {
-		t.Fatalf("Failed to sign transaction: %s", err)
-	}
+	tx, _ = types.SignTx(tx, types.LatestSignerForChainID(big.NewInt(1337)), testKey)
```
ARR4N, line 128: Why remove an error check in a test?

qdm12: Because it's not checked in the master branch 🤷  It feels like it's out of scope of libevm

ARR4N (changes requested):

accounts/abi/bind/util_test.go
qdm12: Why does it say

> Error Handling Improvements (already added in master)

though

core/blockchain.go
```diff
@@ -194,7 +196,7 @@ func (c *CacheConfig) triedbConfig() *triedb.Config {
 		config.DBOverride = hashdb.Config{
 			CleanCacheSize:                  c.TrieCleanLimit * 1024 * 1024,
 			StatsPrefix:                     trieCleanCacheStatsNamespace,
-			ReferenceRootAtomicallyOnUpdate: true, // Automatically reference root nodes when an update is made
+			ReferenceRootAtomicallyOnUpdate: true,
```
darioush, line 199: nit: why remove the comment?

triedb/pathdb/metrics.go
```diff
@@ -35,11 +35,12 @@ import (
...
-// metrics.GetOrRegister*() and this package's corresponding libevm package
-// imported above. Together these ensure that the metric here is the same as the
-// one with the same name in libevm.
-// nolint: unused
+// All calls to metrics.NewRegistered*() for metrics also defined in libevm/triedb/pathdb
+// have been replaced with metrics.GetOrRegister*() to get metrics already registered in
+// libevm/triedb/pathdb or register them here otherwise. These replacements ensure the same
+// metrics are shared between the two packages.
+//
+//nolint:unused
```
qdm12, line 43: related to that: why do we have the last 3 metrics unused? Shouldn't we remove them??

core/blockchain.go
qdm12: Because it's not part of the master branch https://github.com/ava-labs/coreth/pull/820#discussion_r1984982667

plugin/evm/rawdb/accessors_state_sync_test.go
```diff
@@ -7,13 +7,13 @@ import (
...
 	"github.com/ava-labs/libevm/common"
-	"github.com/ava-labs/libevm/core/rawdb"
+	ethrawdb "github.com/ava-labs/libevm/core/rawdb"
 	"github.com/stretchr/testify/require"
 )
 
 func TestClearPrefix(t *testing.T) {
 	require := require.New(t)
-	db := rawdb.NewMemoryDatabase()
+	db := ethrawdb.NewMemoryDatabase()
```
darioush, line 16: not sure why the rename to ethrawdb is needed here?

triedb/pathdb/metrics.go
darioush: I left them in for easier merges.

plugin/evm/rawdb/accessors_state_sync_test.go
qdm12: For consistency, given it's used as ethrawdb in other places in plugin/evm/rawdb. Also to remove the confusion given the package is also named `rawdb`.

core/blockchain.go
qdm12: It's a revert commit, this was added in the libevm branch but not present in master branch, and I don't think it's related to libevm really right?

alarso16 (approved):

core/state/statedb.go
```diff
@@ -162,15 +162,15 @@ func (s *StateDB) Copy() *StateDB {
 }
 
 // NormalizeCoinID ORs the 0th bit of the first byte in
-// [coinID], which ensures this bit will be 1 and all other
+// `coinID`, which ensures this bit will be 1 and all other
```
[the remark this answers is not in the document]
qdm12: Bracket style is for godoc links (which works on the go.dev documentation pages and in the IDE as well, which is pretty useful). However, it doesn't always work. It works notably with:

- local-package **global** symbols (types, functions, variables, methods in the form `[type.method]`)
- external exported symbols with an imported package in the same file, i.e. `package.Type.Method` if `github.com/x/package` is imported in the file where the comment is

It however notably does NOT work for:

- function arguments (which we over-use, and it's frustrating to CMD+click and it doesn't resolve)
- type 'alias' methods (sometimes a notable problem with libevm)

So the rule is backticks (or nothing) when brackets don't work, otherwise prefer using brackets.

Also we (at least I) really really really need a linter for this 😄

core/state_processor_ext.go
```diff
@@ -15,16 +17,17 @@ import (
 )
 
 // ApplyPrecompileActivations checks if any of the precompiles specified by the chain config are enabled or disabled by the block
-// transition from [parentTimestamp] to the timestamp set in [blockContext]. If this is the case, it calls [Configure]
+// transition from `parentTimestamp“ to the timestamp set in `blockContext`. If this is the case, it calls [modules.Module]'s Configure
```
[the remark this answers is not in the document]
qdm12: Good catch!
1. I was wondering how the hell I typed that `“` character, I don't even know how.
2. Proceeds to change this weirdo character
3. Enter the backtick key on my keyboard
4. VSCode inputs two backticks when inputting this at end of a word (`parentTimestamp`), trying to act clever or something. Might be due to a recent vscode update, I never noticed that before.
5. Save and the go formatter `gofmt` formats two backticks into `“`

TIL......

- two backticks get formatted in `“` 🤯  Useless
- VScode has silly choices with its updates sometimes... 🤷

plugin/evm/rawdb/accessors_state_sync_test.go
qdm12: This also makes me think, it might be interesting to rename plugin/evm/rawdb package name from `rawdb` to `customrawdb`, and then use `rawdb` instead of `ethrawdb`? So that it fits with other source files? Thoughts?

triedb/hashdb/database.go
```diff
@@ -114,15 +114,8 @@ type Config struct {
 	ReferenceRootAtomicallyOnUpdate bool   // Whether to reference the root node on update
 }
 
-func (c Config) BackendConstructor(diskdb ethdb.Database, config *triedb.Config) triedb.DBOverride {
-	var resolver ChildResolver
-	if config.IsVerkle {
-		// TODO define verkle resolver
-		log.Crit("Verkle node resolver is not defined")
```
darioush, line 121: why did we remove this? we'll want to pass the correct option when verkle has a trie resolver type we can use here.

predicate/predicate_slots.go
```diff
@@ -9,15 +9,15 @@ import (
 	"github.com/ava-labs/libevm/core/types"
 )
 
-type predicaters interface {
+type PredicaterExistChecker interface {
```
darioush, line 12: doesn't seem this type needs to be exported

plugin/evm/rawdb/accessors_state_sync_test.go
darioush: that works for me, overall I think the "extra"/"custom" packages should follow a single naming convention (custom seems fine) and live under `plugin/evm`

plugin/evm/vm_warp_test.go
```diff
@@ -656,7 +656,6 @@ func testReceiveWarpMessage(
 	ethBlock := block2.(*chain.BlockWrapper).Block.(*Block).ethBlock
 	rules := params.GetExtra(vm.chainConfig).GetAvalancheRules(ethBlock.Time())
 	headerPredicateResultsBytes := customheader.PredicateBytesFromExtra(rules, ethBlock.Extra())
-	require.NotEmpty(headerPredicateResultsBytes)
```
darioush, line 659: not sure why this is removed?

precompile/contracts/warp/contract_test.go
```diff
@@ -11,7 +11,6 @@ import (
 	"github.com/ava-labs/avalanchego/ids"
 	agoUtils "github.com/ava-labs/avalanchego/utils"
 	"github.com/ava-labs/avalanchego/utils/set"
-	"github.com/ava-labs/avalanchego/vms/platformvm/warp"
 	avalancheWarp "github.com/ava-labs/avalanchego/vms/platformvm/warp"
```
darioush, line 14: not sure why we don't just import this as warp

triedb/hashdb/database.go
qdm12: Removed following your comment [here](https://github.com/ava-labs/coreth/pull/820#discussion_r2010989796) 😉 
IMO we can always re-add this later, once it's supported.

predicate/predicate_slots.go
qdm12: Correct; I don't know why I kept exporting argument types of exported functions, I must had read somewhere a reason for it but I cannot find it again 🤷  Unexporting ⛵  !

plugin/evm/vm_warp_test.go
qdm12: This was added in libevm (not present in master), seems a bit off to have that diff, maybe it was added to debug something back then? 🤔

core/extstate/statedb.go
```diff
@@ -27,17 +26,26 @@ type StateDB struct {
 	VmStateDB
```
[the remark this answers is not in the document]
qdm12: We could, but let's keep it exported given `New` returns it and is used by other packages.

predicate/predicate_slots.go
qdm12: Actually, re-exporting!

> As a general rule I'd say yes because it's kind to the users of your package.

I did read this as well somewhere, and, although it goes against encapsulation, it does help the consumer of the exported function. For example godoc will be generated on that type as well, whereas it would not if unexported.

plugin/evm/rawdb/accessors_state_sync_test.go
qdm12: Done in #888 so that file movements are tracked properly and not mixed up with changes done here.

qdm12: Also, regarding this, should we rename `_ext.go` files to `_custom.go` files? Or it's fine as is? 🤔

plugin/evm/header/extra_test.go
```diff
@@ -610,146 +610,68 @@ func TestVerifyExtra(t *testing.T) {
...
-		{
-			name:      "extra_too_short",
-			extra:     []byte{1},
 			predicate: []byte{2},
-			want: []byte{
+			wantExtraWithPredicate: []byte{
 				0:              1,
 				ap3.WindowSize: 2,
 			},
+			wantPredicateBytes: []byte{2},
```
ceyonur, line 650: seems like these are always equal to `test.predicate`, right?

plugin/evm/header/extra_test.go
```diff
@@ -610,146 +610,68 @@ func TestVerifyExtra(t *testing.T) {
...
 			},
-			extra:    make([]byte, acp176.StateSize),
-			expected: nil,
+			extra:                  nil,
+			predicate:              nil,
+			wantExtraWithPredicate: make([]byte, acp176.StateSize),
+			wantPredicateBytes:     nil,
 		},
 		{
-			name: "fortuna_non_empty_predicate",
```
ceyonur, line 667: should we also move this to new test?

plugin/evm/header/extra_test.go
```diff
@@ -610,146 +610,68 @@ func TestVerifyExtra(t *testing.T) {
...
+			wantExtraWithPredicate: make([]byte, ap3.WindowSize),
+			wantPredicateBytes:     nil,
 		},
 		{
-			name:     "empty_predicate",
-			extra:    make([]byte, ap3.WindowSize),
-			expected: nil,
-		},
-		{
-			name: "non_empty_predicate",
```
ceyonur, line 636: should we move this as well?

plugin/evm/vm_warp_test.go
ceyonur: I think `ParseResults` will end up returning an error if it's empty.

core/extstate/statedb.go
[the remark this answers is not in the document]
qdm12: 3a6e390ec unexports the internal field `VmStateDB` (as `vmStateDB` as a type reference), so there is now no point trying to initialize `StateDB{}`, it's obvious it would fail and you would be force to use the constructor.

params/hooks_libevm.go
```diff
@@ -74,7 +75,7 @@ func (r RulesExtra) ActivePrecompiles(existing []common.Address) []common.Addres
 	}
 
 	var addresses []common.Address
-	addresses = append(addresses, maps.Keys(precompiles)...)
+	addresses = slices.AppendSeq(addresses, maps.Keys(precompiles))
```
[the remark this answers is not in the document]
qdm12: What do you mean by

> Also it looks weird we use it here but not below.

plugin/evm/header/extra_test.go
qdm12: Not really, this is covered by test case `extra_too_short_fortuna`

plugin/evm/header/extra_test.go
qdm12: Done in dfb97fe92

params/hooks_libevm.go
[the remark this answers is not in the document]
ceyonur: I was referring to L70, but nevermind I was not aware that `maps.Keys` was an iterator.

core/extstate/statedb.go
```diff
@@ -22,8 +22,10 @@ type VmStateDB interface {
 	SubBalanceMultiCoin(common.Address, common.Hash, *big.Int)
 }
 
+type vmStateDB = VmStateDB
```
darioush, line 25: this is a cool trick

darioush (approved):

ceyonur (approved):

core/state/database.go
```diff
@@ -27,165 +27,16 @@
...
-}
-
-// TrieDB retrieves any intermediate trie-node caching layer.
-func (db *cachingDB) TrieDB() *triedb.Database {
-	return db.triedb
-}
+var (
+	NewDatabase           = ethstate.NewDatabase
+	NewDatabaseWithConfig = ethstate.NewDatabaseWithConfig
+	NewDatabaseWithNodeDB = ethstate.NewDatabaseWithNodeDB
```
ARR4N, line 41: (optional) I think these should be actual functions that propagate values, but the others have all approved of this so up to y'all.

core/state/statedb_multicoin_test.go
```diff
@@ -13,38 +13,37 @@ import (
...
 
 	s.state.AddBalanceMultiCoin(addr, assetID, big.NewInt(10))
 	s.state.SubBalanceMultiCoin(addr, assetID, big.NewInt(5))
 	s.state.AddBalanceMultiCoin(addr, assetID, big.NewInt(3))
 
 	balance = s.state.GetBalanceMultiCoin(addr, assetID)
-	if balance.Cmp(big.NewInt(8)) != 0 {
-		t.Fatal("expected multicoin balance to be 8")
-	}
+	require.Equal(t, big.NewInt(8).String(), balance.String(), "unexpected multicoin balance")
```
ARR4N, line 39:
```suggestion
	require.Equal(t, "8", balance.String(), "big.Int multicoin balance as string")
```

core/state/statedb_multicoin_test.go
```diff
@@ -13,38 +13,37 @@ import (
...
+	s.state, err = New(root, s.state.db, s.state.snaps)
+	require.NoError(t, err, "creating statedb")
 
 	s.state.AddBalance(addr, new(uint256.Int))
 
 	balance := s.state.GetBalanceMultiCoin(addr, assetID)
-	if balance.Cmp(big.NewInt(0)) != 0 {
-		t.Fatal("expected zero multicoin balance")
-	}
+	require.Equal(t, big.NewInt(0).String(), balance.String(), "expected zero multicoin balance")
```
ARR4N, line 32:
```suggestion
	require.Equal(t, "0", balance.String(), "big.Int multicoin balance as string")
```

<rant>I wish I had the time to make a drop-in replacement for `testify`</rant>

params/hooks_libevm.go
ARR4N, line 78: (musing) This suggests that variadic expansion with `...` can be applied to iterators (at least in `append`), which is nice.

plugin/evm/customrawdb/accessors_snapshot_ext.go
```diff
@@ -5,7 +5,7 @@ package customrawdb
 
 import (
 	"github.com/ava-labs/libevm/common"
-	"github.com/ava-labs/libevm/core/rawdb"
+	ethrawdb "github.com/ava-labs/libevm/core/rawdb"
```
ARR4N, line 8: Why does this need to be aliased? We've completely deleted the local `./core/rawdb` so `libevm/core/rawdb` should just be added to the list of allowed imports.

plugin/evm/customrawdb/accessors_state_sync.go
```diff
@@ -8,7 +8,7 @@ import (
 
 	"github.com/ava-labs/avalanchego/utils/wrappers"
 	"github.com/ava-labs/libevm/common"
-	"github.com/ava-labs/libevm/core/rawdb"
+	ethrawdb "github.com/ava-labs/libevm/core/rawdb"
```
ARR4N, line 11: See other comment re `libevm/core/rawdb` not being aliased.

plugin/evm/customrawdb/accessors_state_sync_test.go
```diff
@@ -7,13 +7,13 @@ import (
 	"testing"
 
 	"github.com/ava-labs/libevm/common"
-	"github.com/ava-labs/libevm/core/rawdb"
+	ethrawdb "github.com/ava-labs/libevm/core/rawdb"
```
ARR4N, line 10: See other comment re `libevm/core/rawdb` not being aliased.

plugin/evm/customrawdb/database_ext_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"fmt"
 
 	"github.com/ava-labs/libevm/common"
-	"github.com/ava-labs/libevm/core/rawdb"
+	ethrawdb "github.com/ava-labs/libevm/core/rawdb"
```
ARR4N, line 10: See other comment re `libevm/core/rawdb` not being aliased.

plugin/evm/header/block_gas_cost.go
```diff
@@ -89,20 +89,19 @@ func EstimateRequiredTip(
 	config *extras.ChainConfig,
 	header *types.Header,
 ) (*big.Int, error) {
-	extra := customtypes.GetHeaderExtra(header)
+	headerExtra := customtypes.GetHeaderExtra(header)
```
ARR4N, line 92: (subjective) The `header` prefix IMO is redundant as it can be inferred by both the RHS of the declaration and the fact that we're in a package called `header`; it therefore only adds noise but limited signal wherever the variable is used.

plugin/evm/header/extra_test.go
```diff
@@ -610,102 +610,37 @@ func TestVerifyExtra(t *testing.T) {
 	}
 }
 
-func TestPredicateBytesFromExtra(t *testing.T) {
+func TestPredicateBytesExtra(t *testing.T) {
```
ARR4N, line 613: Why is this test being overhauled in a PR that's supposed to be limited to "small changes to the libevm branch"? The same goes for deletion of `TestSetPredicateBytesInExtra`.

This is way too significant a change to just lump in like this because it risks hiding a correlated break somewhere else that the change in the test no longer picks up.

plugin/evm/vm.go
```diff
@@ -807,9 +807,8 @@ func (vm *VM) preBatchOnFinalizeAndAssemble(header *types.Header, state *state.S
 		// Note: snapshot is taken inside the loop because you cannot revert to the same snapshot more than
 		// once.
 		snapshot := state.Snapshot()
-		rules := vm.chainConfig.Rules(header.Number, params.IsMergeTODO, header.Time)
-		rulesExtra := params.GetRulesExtra(rules)
-		if err := vm.verifyTx(tx, header.ParentHash, header.BaseFee, state, *rulesExtra); err != nil {
+		rules := params.GetRulesExtra(vm.chainConfig.Rules(header.Number, params.IsMergeTODO, header.Time))
```
ARR4N, line 810:
```suggestion
		ethrules := vm.chainConfig.Rules(header.Number, params.IsMergeTODO, header.Time)
		rules := params.GetRulesExtra(ethrules)
```

I agree with keeping the variable actually being used as named `rules` but think the line is a bit unwieldy as it is.

EDIT: I see that this is a common occurrence so, alternatively, you could abstract the pattern as a method like `rules := vm.rulesExtra(header)`.

plugin/evm/vm.go
```diff
@@ -855,8 +854,7 @@ func (vm *VM) postBatchOnFinalizeAndAssemble(
 		batchAtomicUTXOs  set.Set[ids.ID]
 		batchContribution *big.Int = new(big.Int).Set(common.Big0)
 		batchGasUsed      *big.Int = new(big.Int).Set(common.Big0)
-		rules                      = vm.chainConfig.Rules(header.Number, params.IsMergeTODO, header.Time)
-		rulesExtra                 = *params.GetRulesExtra(rules)
+		rules                      = *params.GetRulesExtra(vm.chainConfig.Rules(header.Number, params.IsMergeTODO, header.Time))
```
ARR4N, line 857: As above

plugin/evm/vm.go
```diff
@@ -972,11 +970,10 @@ func (vm *VM) onExtraStateChange(block *types.Block, parent *types.Header, state
 		batchContribution *big.Int = big.NewInt(0)
 		batchGasUsed      *big.Int = big.NewInt(0)
 		header                     = block.Header()
-		rules                      = vm.chainConfig.Rules(header.Number, params.IsMergeTODO, header.Time)
-		rulesExtra                 = *params.GetRulesExtra(rules)
+		rules                      = *params.GetRulesExtra(vm.chainConfig.Rules(header.Number, params.IsMergeTODO, header.Time))
```
ARR4N, line 973: As above

plugin/evm/vm_warp_test.go
ARR4N: Alignment with `master` is not the correct litmus test here; correctness of the `libevm` branch is what we should be aiming for. This was added for a reason; if that reason is incorrect then yes, remove it, but otherwise it should be kept as a requirement.

predicate/predicate_slots.go
ARR4N: It also allows a provider of the type to locally assert that they implement it, in an idiomatic way.

```go
var _ predicate.PredicaterExistChecker = (*myImpl)(nil) // idiomatic

var _ = predicate.PreparePredicateStorageSlots((*myImpl)(nil), ...) // wtf? might also panic
```

eth/tracers/tracers.go
ARR4N, line 1: (note to self) `git diff b0332b516 -- ./eth/tracers/tracers.go` reveals these parts to be identical.

core/state/database.go
ARR4N, line 1: (note to self) Deleted code is the same as `libevm@b0332b516` (a recent commit to `libevm/main` that I happen to have in `coreth`).

core/state/snapshot/iterator.go
ARR4N, line 1: (note to self) Deleted code is the same as libevm@b0332b516 (a recent commit to libevm/main that I happen to have in coreth).

ARR4N (commented):
Please _DO NOT_ force push to this branch. It will wipe GitHub's memory of my last reviews.

plugin/evm/vm_warp_test.go
qdm12: It has no reason to be there, and, yes, I agree, given we won't squash-merge libevm, aligning with master doesn't help much, except maybe to compare the libevm and master branches for the time being.

core/blockchain.go
```diff
@@ -1851,7 +1852,7 @@ func (bc *BlockChain) reprocessState(current *types.Block, reexec uint64) error
 		// Flatten snapshot if initialized, holding a reference to the state root until the next block
 		// is processed.
 		if err := bc.flattenSnapshot(func() error {
-			if previousRoot != (common.Hash{}) {
+			if previousRoot != (common.Hash{}) && previousRoot != root {
```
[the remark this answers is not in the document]
qdm12: it's a reversion, it was there on the master branch, and there isn't a clear reason why it was removed.
Subnet-evm has it on the master branch and the libevm branch, and has its e2e tests fail without it.
Better avoid weird subtle surprises by keeping it in the libevm branch as well.

core/extstate/statedb.go
```diff
@@ -16,28 +16,38 @@ import (
...
 	GetBalanceMultiCoin(common.Address, common.Hash) *big.Int
 	AddBalanceMultiCoin(common.Address, common.Hash, *big.Int)
 	SubBalanceMultiCoin(common.Address, common.Hash, *big.Int)
 }
 
+type vmStateDB = VmStateDB
+
 type StateDB struct {
-	VmStateDB
+	vmStateDB
```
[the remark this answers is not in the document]
qdm12: For https://github.com/ava-labs/coreth/pull/862#discussion_r2013893702

core/state/database.go
qdm12: Done! I just omitted the comments on the actual functions to avoid having to maintain in sync the libevm comments and the coreth comments.

core/state/statedb_multicoin_test.go
qdm12: That's not really the context, the problem is really an `unexpected multicoin balance string`. Otherwise it sounds like the string conversion is the problem 😉 

```suggestion
	require.Equal(t, big.NewInt(8).String(), balance.String(), "unexpected multicoin balance string")
```

core/state/statedb_multicoin_test.go
qdm12: This 🆙 and a linter for square brackets godoc links in comments

qdm12:
```suggestion
	require.Equal(t, "0", balance.String(), "expected zero big.Int multicoin balance as string")
```

params/hooks_libevm.go
qdm12: Well TIL `musing` 🧠 🧠 

Out of the vocabulary realm, I don't think you can right? For example `append(addresses, maps.Keys(precompiles)...)` doesn't work.

plugin/evm/header/extra_test.go
qdm12: That was a suggestion made by @ceyonur:

> On the side note should this [test](https://github.com/ava-labs/coreth/pull/828/commits/21163f1808c07ee81e7eb07e17538971ccef75de) check against PredicateBytesFromExtra  to make sure they're inverse of each other?

and was not included in the (merge master) PR to avoid having to rebase + re-review everything. I can split it in another PR if you feel it should be separate, fine with me as well.

plugin/evm/vm.go
qdm12: Did it with function

```go
func (vm *VM) rulesExtra(number *big.Int, time uint64) extras.Rules {
	ethrules := vm.chainConfig.Rules(number, params.IsMergeTODO, time)
	return *params.GetRulesExtra(ethrules)
}
```

So number+time instead of header argument, such that it can be used with blocks without having to deep copy the header.

plugin/evm/header/block_gas_cost.go
qdm12: Done! (changed below code to use `extra` instead of `headerExtra`

plugin/evm/header/extra_test.go
qdm12: Re-added previous tests, and made this new `TestPredicateBytesExtra` test just an addition on of of the two tests.

ARR4N: Thanks. I'm not disagreeing with Cey's suggestion; just saying it should be standalone from a review-safety perspective. I'm very happy with adding new tests (as you've done it).

plugin/evm/vm.go
```diff
@@ -1745,11 +1745,15 @@ func (vm *VM) chainConfigExtra() *extras.ChainConfig {
 	return params.GetExtra(vm.chainConfig)
 }
 
+func (vm *VM) rulesExtra(number *big.Int, time uint64) extras.Rules {
```
ARR4N, line 1748: (optional) Since you're calling the other one `currentRules()`, consider just calling this `rules()`.

core/blockchain.go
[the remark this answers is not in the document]
ARR4N: Reversion sounds good to me. @darioush / @ceyonur WDYT? I know you've reviewed this PR already but just want to double-check that you saw this and are happy with it. It's very much in your areas of expertise and not in mine.

@qdm12 I'm going to approve but please double-check with Darioush or Cey (unless you already have).

core/extstate/statedb.go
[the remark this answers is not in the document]
ARR4N: To force use of the constructor?

params/hooks_libevm.go
ARR4N: It looked like that was the exact pattern in the old code though, no?

ARR4N (approved):

core/extstate/statedb.go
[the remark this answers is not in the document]
qdm12: Yes sort of 🤷

plugin/evm/vm.go
qdm12: Done!

params/hooks_libevm.go
qdm12: It was but the exp-version maps behaves differently than the std maps package

core/blockchain.go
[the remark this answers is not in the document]
darioush: It's needed because [this](https://github.com/ava-labs/subnet-evm/commit/ecf5cc9cdf2d1de2e4528c7dca8c15161a700596#diff-c3757dc9e9d868f63bc84a0cc67159c1d5c22cc5d8c9468757098f0492e0658cR1346) only happens if `root != origin` so the deref should also only happen then

qdm12:
[No-op force push](https://github.com/ava-labs/coreth/compare/1c6b47ff5e3ecf85f8589d6aac59464302169938..8b5354b51d20c4ad1b45ce2e9bba75eb3dd9b378) due to a missing signature 🤷

ARR4N:
> [No-op force push](https://github.com/ava-labs/coreth/compare/1c6b47ff5e3ecf85f8589d6aac59464302169938..8b5354b51d20c4ad1b45ce2e9bba75eb3dd9b378) due to a missing signature 🤷

Even "no-op" force pushes can break other people's workflows. Just GPG-sign everything.

By definition there can never be a completely no-op force push. While it may not change the file tree, it _will_ change the commit history, and that's what GitHub relies on. If it was a no-op then a force push would be unnecessary.

Issue #858, opened 2025-03-05
Remove BaseFee sampling in the API


StephenButtolph:
## Description

Currently `EstimateBaseFee` returns the minimum of:
- The 60th percentile `BaseFee` of the last `40` blocks
- An upper bound on the next block's `BaseFee`

This means that the `EstimateBaseFee` is systemically under-estimating when the `BaseFee` is increasing.

However, if the `BaseFee` is stable or decreasing then `EstimateBaseFee` will return a reasonable number.

## Suggestion

Just return the upper bound on the next block's `BaseFee`.

PR #865, opened 2025-03-10
fix(eth/gasprice): EstimateBaseFee using Next Fee

alarso16:
## Why this should be merged
See issue #858 

## How this works
Rather than sampling base fees as well as tips, simply returns the next base fee in `SuggestPrice` and `EstimateBaseFee`. Additionally, removes sampling and tracking of gas fees from sampling method.

## How this was tested
Existing unit tests

## Need to be documented?
Yes, but unsure where

No

### DIFF
--- a/eth/gasprice/gasprice.go
+++ b/eth/gasprice/gasprice.go
@@ -37,11 +37,9 @@ import (
 	"github.com/ava-labs/coreth/params"
 	customheader "github.com/ava-labs/coreth/plugin/evm/header"
 	"github.com/ava-labs/coreth/plugin/evm/upgrade/acp176"
-	"github.com/ava-labs/coreth/plugin/evm/upgrade/ap3"
 	"github.com/ava-labs/coreth/rpc"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/lru"
-	"github.com/ethereum/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/event"
 	"github.com/ethereum/go-ethereum/log"
 	"golang.org/x/exp/slices"
@@ -66,7 +64,6 @@ const (
 var (
 	DefaultMaxPrice           = big.NewInt(150 * params.GWei)
 	DefaultMinPrice           = big.NewInt(acp176.MinGasPrice)
-	DefaultMinBaseFee         = big.NewInt(ap3.InitialBaseFee)
 	DefaultMinGasUsed         = big.NewInt(acp176.MinTargetPerSecond)
 	DefaultMaxLookbackSeconds = uint64(80)
 )
@@ -106,10 +103,9 @@ type OracleBackend interface {
 // Oracle recommends gas prices based on the content of recent
 // blocks. Suitable for both light and full clients.
 type Oracle struct {
-	backend     OracleBackend
-	lastHead    common.Hash
-	lastPrice   *big.Int
-	lastBaseFee *big.Int
+	backend   OracleBackend
+	lastHead  common.Hash
+	lastPrice *big.Int
 	// [minPrice] ensures we don't get into a positive feedback loop where tips
 	// sink to 0 during a period of slow block production, such that nobody's
 	// transactions will be included until the full block fee duration has
@@ -196,7 +192,6 @@ func NewOracle(backend OracleBackend, config Config) (*Oracle, error) {
 	return &Oracle{
 		backend:             backend,
 		lastPrice:           minPrice,
-		lastBaseFee:         DefaultMinBaseFee,
 		minPrice:            minPrice,
 		maxPrice:            maxPrice,
 		checkBlocks:         blocks,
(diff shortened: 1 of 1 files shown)


eth/gasprice/gasprice.go
```diff
@@ -213,26 +208,12 @@ func NewOracle(backend OracleBackend, config Config) (*Oracle, error) {
...
-	// We calculate the [nextBaseFee] if a block were to be produced immediately.
-	// If [nextBaseFee] is lower than the estimate from sampling, then we return it
-	// to prevent returning an incorrectly high fee when the network is quiescent.
 	nextBaseFee, err := oracle.estimateNextBaseFee(ctx)
-	if err != nil {
-		log.Warn("failed to estimate next base fee", "err", err)
-		return baseFee, nil
-	}
-	// If base fees have not been enabled, return a nil value.
 	if nextBaseFee == nil {
```
ceyonur, line 212: I think we can have `err != nil` but `nextBase == nil` here and in that case we won't return the err. I wonder if we should just `oracle.estimateNextBaseFee(ctx)` directly.

eth/gasprice/gasprice.go
```diff
@@ -260,25 +241,18 @@ func (oracle *Oracle) estimateNextBaseFee(ctx context.Context) (*big.Int, error)
...
 
-	// We calculate the [nextBaseFee] if a block were to be produced immediately.
-	// If [nextBaseFee] is lower than the estimate from sampling, then we return it
-	// to prevent returning an incorrectly high fee when the network is quiescent.
+	// We calculate the `nextBaseFee` if a block were to be produced immediately.
 	nextBaseFee, err := oracle.estimateNextBaseFee(ctx)
-	if err != nil {
-		log.Warn("failed to estimate next base fee", "err", err)
+	if nextBaseFee == nil || err != nil {
+		log.Warn("Failed to estimate next base fee, returning tip", "err", err)
```
ceyonur, line 252: might be weird to have a nil err in warn log here

alarso16: True, it is expected to be nil in some well-defined behavior.

eth/gasprice/gasprice.go
```diff
@@ -288,42 +262,41 @@ func (oracle *Oracle) SuggestPrice(ctx context.Context) (*big.Int, error) {
 // necessary to add the basefee to the returned number to fall back to the legacy
 // behavior.
 func (oracle *Oracle) SuggestTipCap(ctx context.Context) (*big.Int, error) {
-	tip, _, err := oracle.suggestDynamicFees(ctx)
+	tip, err := oracle.suggestTip(ctx)
 	return tip, err
```
StephenButtolph, line 269:
```suggestion
	return oracle.suggestTip(ctx)
```

eth/gasprice/gasprice.go
StephenButtolph: I think this function can just be `return oracle.estimateNextBaseFee(ctx)`. `estimateNextBaseFee` returns `nil` if the `BaseFee` on the last accepted block is `nil`, which guarantees that dynamic fees has not only been scheduled, but is already activated.

ceyonur: I think this won't ever return it (since we already past AP3). if you want to maintain the old behaviour then we should log.warn the err and then `return nil,nil`, but I think we can return the err as well.

eth/gasprice/gasprice.go
```diff
@@ -260,25 +236,21 @@ func (oracle *Oracle) estimateNextBaseFee(ctx context.Context) (*big.Int, error)
...
 
-	// We calculate the [nextBaseFee] if a block were to be produced immediately.
-	// If [nextBaseFee] is lower than the estimate from sampling, then we return it
-	// to prevent returning an incorrectly high fee when the network is quiescent.
+	// We calculate the `nextBaseFee` if a block were to be produced immediately.
 	nextBaseFee, err := oracle.estimateNextBaseFee(ctx)
-	if err != nil {
+	if nextBaseFee == nil {
+		// This occurs if AP3 has not been activated
+		return nil, nil
```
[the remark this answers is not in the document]
alarso16: I think you're right.

eth/gasprice/gasprice.go
```diff
@@ -260,25 +236,21 @@ func (oracle *Oracle) estimateNextBaseFee(ctx context.Context) (*big.Int, error)
...
-	// We calculate the [nextBaseFee] if a block were to be produced immediately.
-	// If [nextBaseFee] is lower than the estimate from sampling, then we return it
-	// to prevent returning an incorrectly high fee when the network is quiescent.
+	// We calculate the `nextBaseFee` if a block were to be produced immediately.
 	nextBaseFee, err := oracle.estimateNextBaseFee(ctx)
-	if err != nil {
+	if nextBaseFee == nil {
+		// This occurs if AP3 has not been activated
+		return tip, nil
+	} else if err != nil {
```
StephenButtolph, line 249: Sorry for the back&forth, but I think it is unexpected to check `nextBaseFee == nil` _before_ `err != nil`. If `err != nil`, I don't think we should consider `nextBaseFee` a meaningful value here.

alarso16: If `err != nil` for any reason, it isn't used. However, `nextBaseFee == nil` implies `err != nil` , but not vice-versa. I guess the real point here is what's the behavior pre-AP3? Prior to this PR, we would default to `suggestDynamicFees`, but that should not be returned. I don't think we should return an error on expected behavior.

ceyonur: internal implications should not matter for err handling in call places. Errs should (almost) always checked before the value is being used/check. 

I think in this case we can check err first and return the err(to keep it same behaviour with  `EstimateBaseFee`) and then perform addition if nextBaseFee is not nil.

ceyonur: the code got updated before my comment here.

eth/gasprice/gasprice.go
```diff
@@ -260,25 +236,22 @@ func (oracle *Oracle) estimateNextBaseFee(ctx context.Context) (*big.Int, error)
...
 		log.Warn("failed to estimate next base fee", "err", err)
-	}
-	// Separately from checking the error value, check that [nextBaseFee] is non-nil
-	// before attempting to take the minimum.
-	if nextBaseFee != nil {
-		baseFee = math.BigMin(baseFee, nextBaseFee)
+		return nil, err
+	} else if nextBaseFee == nil {
+		// This occurs if AP3 has not been scheduled yet
+		return tip, nil
```
ceyonur, line 251:
```suggestion
		return nil, err
	} 
	if nextBaseFee == nil {
		// This occurs if AP3 has not been scheduled yet
		return tip, nil
```

ceyonur: I'd say if `else if` was deleted in a go update, no one would know lol.

ceyonur (approved):

StephenButtolph (approved):
Behavior here in the `Dynamic Fees` tests lgtm: https://github.com/ava-labs/avalanchego/pull/3782

StephenButtolph (approved):

darioush (approved):

PR #851, opened 2025-03-02
chore(rawdb): migration to plugin/evm/rawdb

qdm12:
## Why this should be merged

Suggested by @darioush to clean up rawdb a bit further. This is non blocking for the libevm branch to be merged into master.

## How this works

- Move our custom code from core/rawdb to plugin/evm/rawdb
- Remove rawdb/imports.go and import from libevm directly in other files

## How this was tested

CI passing

## Need to be documented?

No

## Need to update RELEASES.md?

No

PR #866, opened 2025-03-11
chore(accounts/abi/bind): revert accidental changes to TestGetSenderNativeAssetCall

qdm12:
## Why this should be merged

I've added by mistake [a refactor commit](https://github.com/ava-labs/coreth/pull/851/commits/2773bbb7d6ee2616600085d473ff81de0330a825) (destined for #862) to already-approved #851 (yesterday evening in a rush, normal git push), and squashed merged it (overlooking that erroneous commit) this morning.

This reverses the changes such that these can be reviewed later in #862 where it should be. Kind of important to merge this so the changes for that test won't be associated with the rawdb commit, since it has nothing to do with it.

~I also suggest we require approvals on the last commit before merging, but keep the "no need to update branch" (as it is, at least for libevm), to keep it flexible whilst also preventing this kind of problem from happening again 😉~

## How this works

Essentially reverts 2773bbb7d6ee2616600085d473ff81de0330a825 (squashed with other commits)

### DIFF
--- a/accounts/abi/bind/bind_extra_test.go
+++ b/accounts/abi/bind/bind_extra_test.go
@@ -18,12 +18,11 @@ import (
 	"github.com/ava-labs/coreth/params"
 	"github.com/ava-labs/libevm/common"
 	"github.com/ava-labs/libevm/crypto"
-	"github.com/stretchr/testify/require"
 )
 
 // TestGetSenderNativeAssetCall checks that the NativeAssetCall proxies the
 // caller address This behavior is disabled on the network and is only to test
-// previous behavior. Note the test uses [params.TestApricotPhase2Config].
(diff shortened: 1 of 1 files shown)


ARR4N:
> [!TIP]
> I recommend reviewing the [diff](https://github.com/ava-labs/coreth/compare/cf26ac159479e7b28e6321eb52a19c2449a9c535...3f4ff4d34780db303315f629906f473c213e0ecb) between this single commit of this PR and the state of the `libevm` branch before the corrupt PR. This gives the final state after combining (a) the corrupt PR and (b) the fix.

ARR4N (approved):

darioush (approved):

PR #872, opened 2025-03-13
refactor(core/types): separate coreth and upstream identifiers

ARR4N:
> [!NOTE]
> ~The broken CI will be fixed by fjl/gencodec#28. I'll update to that version once merged, but review of this PR can begin in the interim.~

## Why this should be merged

Reduces overlap between `coreth` and upstream files.

## How this works

1. Types and variables that are identical to `libevm` are moved to `imports.go`.
2. Files that include only `coreth` have the `_ext.go` suffix added.

> [!NOTE]
> This is part of a chain of PRs, followed by #875. The chain is split so that each one is "obviously" correct. I will not consider any review changes that deviate from this simplicity—if this is an issue, please explain why and add a "Do not merge" label.

## How this was tested

This is a no-op with respect to logic already undergoing bootstrap testing. The existing CI and a pre-release canary are therefore sufficient.

## Need to be documented?

No

No


core/types/account.go
ARR4N, line 1: This was identical so `Account` and `GenesisAlloc` were moved to `imports.go`.

core/types/hashes.go
ARR4N, line 1: The only differences here were the addition `EmptyExtDataHash` (now in `hashes_ext.go`) and the removal of `EmptyWithdrawalsHash` (which doesn't need any changes to undo the file deletion).

core/types/log.go
ARR4N, line 1: The existing `type Log = ethtypes.Log` was moved to `imports.go` for consistency and the remaining file contents was renamed as `log_ext.go`.

core/types/state_account.go
ARR4N, line 1: Renamed to `state_account_ext.go`.

core/types/hashing.go
ARR4N, line 1: Removal of `rlpHash()` is blocked by ava-labs/libevm#163.

qdm12: Looks like it's merged, should we cut a new libevm release for that? Or rather later?

core/types/log_ext.go
```diff
@@ -0,0 +1,12 @@
+// (c) 2019-2025, Ava Labs, Inc. All rights reserved.
```
qdm12, line 1: supernit

```suggestion
// (c) 2025, Ava Labs, Inc. All rights reserved.
```

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/crate-crypto/go-ipa v0.0.0-20231025140028-3c0104f4b233
 	github.com/davecgh/go-spew v1.1.1
 	github.com/deckarep/golang-set/v2 v2.1.0
-	github.com/fjl/gencodec v0.1.1-0.20250303202438-8effaadd1b1b
+	github.com/fjl/gencodec v0.1.1-0.20250314100332-a3c3302847ce
```
qdm12, line 12: Awesome 👍 !

qdm12 (approved):

core/types/hashing.go
ARR4N: We're going to have to cut at least one more release anyway so I'm waiting until later as there's no rush with such a tiny thing here.

go.mod
ARR4N: The hardest part was recreating the failing conversion in tests!

ceyonur (approved):

PR #873, opened 2025-03-13
fix(internal/ethapi): fix minBaseFee for fortuna

alarso16:
## Why this should be merged

On Fortuna activation, the `ethapi` uses Etna's min base fee. After ACP-176 is activated, the min base fee is lowered, resulting in an improper min application.

## How this works

If the Fortuna timestamp is reached, will use `acp176.MinGasPrice` instead of `etna.MinBaseFee`.

## How this was tested

Additional unit test checking the `chainCfg` in the eth backend.

## Need to be documented?

No

Unsure

### DIFF
--- a/internal/ethapi/api.coreth.go
+++ b/internal/ethapi/api.coreth.go
@@ -47,11 +47,6 @@ type PriceOptions struct {
 // SuggestPriceOptions returns suggestions for what to display to a user for
 // current transaction fees.
 func (s *EthereumAPI) SuggestPriceOptions(ctx context.Context) (*PriceOptions, error) {
-	var (
-		minBaseFee    int64
-		bigMinBaseFee *big.Int
-	)
-
 	baseFee, err := s.b.EstimateBaseFee(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to estimate base fee: %w", err)
@@ -66,14 +61,17 @@ func (s *EthereumAPI) SuggestPriceOptions(ctx context.Context) (*PriceOptions, e
 		return nil, nil
 	}
 
+	// Find min base fee based on chain config
+	// TODO: This can be removed after Fortuna is activated
 	time := s.b.CurrentHeader().Time
 	chainConfig := s.b.ChainConfig()
+	var minBaseFee int64
 	if chainConfig.IsFortuna(time) {
 		minBaseFee = acp176.MinGasPrice
 	} else {
 		minBaseFee = etna.MinBaseFee
 	}
-	bigMinBaseFee = big.NewInt(minBaseFee)
+	bigMinBaseFee := big.NewInt(minBaseFee)
 
 	cfg := s.b.PriceOptionsConfig()
 	bigSlowFeePercent := new(big.Int).SetUint64(cfg.SlowFeePercentage)
--- a/internal/ethapi/api.coreth.go
+++ b/internal/ethapi/api.coreth.go
@@ -8,20 +8,18 @@ import (
 	"fmt"
 	"math/big"
 
+	"github.com/ava-labs/coreth/plugin/evm/upgrade/acp176"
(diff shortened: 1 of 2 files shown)


darioush (approved):

internal/ethapi/api.coreth.go
```diff
@@ -49,6 +47,11 @@ type PriceOptions struct {
 // SuggestPriceOptions returns suggestions for what to display to a user for
 // current transaction fees.
 func (s *EthereumAPI) SuggestPriceOptions(ctx context.Context) (*PriceOptions, error) {
+	var (
+		minBaseFee    int64
+		bigMinBaseFee *big.Int
+	)
+
```
StephenButtolph, line 54: Can we define these where they are used? Feels odd to define them at the top of the function and then to initialize them later.

internal/ethapi/api.coreth.go
```diff
@@ -63,6 +66,15 @@ func (s *EthereumAPI) SuggestPriceOptions(ctx context.Context) (*PriceOptions, e
...
 	}
 
+	time := s.b.CurrentHeader().Time
+	chainConfig := s.b.ChainConfig()
+	if chainConfig.IsFortuna(time) {
+		minBaseFee = acp176.MinGasPrice
+	} else {
+		minBaseFee = etna.MinBaseFee
+	}
+	bigMinBaseFee = big.NewInt(minBaseFee)
```
StephenButtolph, line 76: Should we add a TODO to remove this once Fortuna activates?

alarso16: Sure, added

StephenButtolph (approved):

internal/ethapi/api.coreth.go
```diff
@@ -63,6 +61,18 @@ func (s *EthereumAPI) SuggestPriceOptions(ctx context.Context) (*PriceOptions, e
...
+	// TODO: This can be removed after Fortuna is activated
+	time := s.b.CurrentHeader().Time
+	chainConfig := s.b.ChainConfig()
+	var minBaseFee int64
+	if chainConfig.IsFortuna(time) {
+		minBaseFee = acp176.MinGasPrice
+	} else {
+		minBaseFee = etna.MinBaseFee
+	}
+	bigMinBaseFee := big.NewInt(minBaseFee)
```
qdm12, line 74: nit could assign directly to a big.Int

```suggestion
	minBaseFee := new(big.Int)
	if chainConfig.IsFortuna(time) {
		minBaseFee.SetUint64(acp176.MinGasPrice)
	} else {
		minBaseFee.SetUint64(etna.MinBaseFee)
	}
```

and remove `bigMinBaseFee`

internal/ethapi/api.coreth_test.go
```diff
@@ -35,13 +38,25 @@ func (b *testSuggestPriceOptionsBackend) PriceOptionsConfig() PriceOptionConfig
...
+
 func TestSuggestPriceOptions(t *testing.T) {
 	testCfg := PriceOptionConfig{
 		SlowFeePercentage: 95,
 		FastFeePercentage: 105,
 		MaxBaseFee:        100 * params.GWei,
 		MaxTip:            20 * params.GWei,
 	}
+	minBaseFee := 1 * params.GWei
+	bigMinBaseFee := big.NewInt(int64(minBaseFee))
```
qdm12, line 57: nit we could just have this as one variable, for example

```suggestion
	minBaseFee := big.NewInt(int64(1 * params.GWei))
```

same below

qdm12 (approved):
Nice 👍

internal/ethapi/api.coreth_test.go
alarso16: I could restructure the test helpers, but seems out of scope for this fix

PR #875, opened 2025-03-14
refactor: use upstream `core/types`

ARR4N:
## Why this should be merged

Complete the transition of `core/types` to upstream code.

## How this works

Note that #872 results in _all_ upstream types / vars / consts being in the `imports.go` file. This results in a strict partition of upstream and local identifiers so it is safe to simply change import paths.

1. Globally change all imports from `coreth/core/types` to `libevm/core/types`;
2. Rename all `coreth/core/types` imports to `customtypes` for ease of review here;
3. Dot import `libevm/core/types` in tests only to avoid hundreds of noisy changes.

> [!NOTE]
> This is part of a chain of PRs ~and will have its base changed once #872 is merged~. The chain is split so that each one is "obviously" correct. I will not consider any review changes that deviate from this simplicity—if this is an issue, please explain why and add a "Do not merge" label.

## How this was tested

CI of existing tests.

## Need to be documented?

No

No


core/types/avalanchego.go
```diff
@@ -0,0 +1,58 @@
...
+	"math/big"
+
+	"github.com/ava-labs/libevm/common"
+	ethtypes "github.com/ava-labs/libevm/core/types"
+)
+
+// TODO(arr4n): this is a temporary workaround because of the circular
+// dependency between the coreth and avalanchego repos. The latter depends on
+// these types via coreth instead of via libevm so there's a chicken-and-egg
+// problem with refactoring both repos.
```
qdm12, line 17: 🐔 🥚 🤯

core/types/block_ext_test.go
```diff
@@ -174,7 +173,7 @@ func TestBlockWithNonZeroFields(t *testing.T) {
 // bodyWithNonZeroFields returns a [Body] and a [BlockBodyExtra],
 // each with all fields set to non-zero values.
 // The [BlockBodyExtra] extra payload is set in the [Body] via `extras.Block.Set`
-// and the extra copying done in [ethtypes.Block.Body].
+// and the extra copying done in [Block.Body].
```
qdm12, line 176: godoc does not support accessing a field from an aliased type (at least my IDE doesn't), maybe just change to

```suggestion
// and the extra copying done in `Block.Body()`.
```

core/types/header_ext_test.go
```diff
@@ -164,7 +163,7 @@ func allFieldsSet[T interface {
 				assertNonZero(t, f)
 			case *Header:
 				assertNonZero(t, f)
-			case []uint8, []*Header, Transactions, []*Transaction, ethtypes.Withdrawals, []*ethtypes.Withdrawal:
+			case []uint8, []*Header, Transactions, []*Transaction, Withdrawals, []*Withdrawal:
```
qdm12, line 166: Should we import from libevm for coreth test files in this package, and keep imports_test.go reserved for avalanchego tests?

rpc/subscription_test.go
```diff
@@ -38,8 +38,11 @@ import (
 	"testing"
 	"time"
 
-	"github.com/ava-labs/coreth/core/types"
 	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/libevm/core/types"
+
+	// Side effect: registration of libevm extras.
+	_ "github.com/ava-labs/coreth/core/types"
```
qdm12, line 45: 👍

qdm12 (approved):

core/types/header_ext_test.go
ARR4N: I'm not sure what version you reviewed that had the `ethtypes.` in there because I'm seeing it without.

Once AvalancheGo is updated to use the `libevm` types instead of `coreth` then we can do a refactor here to remove all aliasing (i.e. delete `avalanchego.go` and `imports_test.go`). This will force all tests to use `libevm/core/types` so we'll have no choice; I'd be open to dot-importing it in these tests to avoid hundreds of changes.

core/types/avalanchego.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"crypto/ecdsa"
+	"math/big"
+
+	"github.com/ava-labs/libevm/common"
+	ethtypes "github.com/ava-labs/libevm/core/types"
+)
+
+// TODO(arr4n): this is a temporary workaround because of the circular
```
darioush, line 14: Can't we make a change to avalanchego first to not import coreth types and import libevm instead, then have this PR without this file?

core/blockchain.go
```diff
@@ -43,7 +43,7 @@ import (
 	"github.com/ava-labs/coreth/consensus/misc/eip4844"
 	"github.com/ava-labs/coreth/core/state"
 	"github.com/ava-labs/coreth/core/state/snapshot"
-	"github.com/ava-labs/coreth/core/types"
+	customtypes "github.com/ava-labs/coreth/core/types"
```
darioush, line 46: I think this is going to be error-prone / confusing.
I'm okay if we do a fast-follow to move the package to at least a different name.

darioush (approved):

michaelkaplan13 (approved):

core/blockchain.go
qdm12: Agreed, some people will forget to use the customtypes alias which will get confusing.
Perhaps we can have a plugin/evm/customtypes or plugin/evm/core/customtypes package?

core/types/avalanchego.go
ARR4N: I considered that but I'm pretty sure that `avalanchego` will only start using `libevm` transitively through `coreth` (currently its `go.mod` points to `ethereum/go-ethereum`). If we were to do it on a branch then the E2E test here would have to check out that branch instead, which would invalidate the test unless we kept said branch up to date (which we would almost certainly forget to do).

core/blockchain.go
ARR4N: I'll start a fast-follow PR to move everything except the `avalanchego` shim. Doing it here would break the KISS requirements of this PR chain.

core/types/avalanchego.go
ARR4N: Summarising our DM chat: I missed that here we depend on such a branch. I've opened ava-labs/avalanchego#3803 and will then remove all aliases here.

core/types/header_ext_test.go
ARR4N: They've [been removed](https://github.com/ava-labs/coreth/pull/875#discussion_r1995903235).

core/types/avalanchego.go
ARR4N: Done.

core/blockchain.go
ARR4N: Moving in #877

PR #745, opened 2025-01-13
chore(metrics): use geth metrics package and delete local metrics

qdm12:
ℹ️  pretty much copy pasta of https://github.com/ava-labs/subnet-evm/pull/1422

## Why this should be merged

Contribution to the libevm effort.
Note there are the same changes for both coreth and subnet-evm, but they should not be part of libevm since these are extra features specific to subnet-evm/coreth, even if they're the same for both.

## How this works

Comparing the following:

- subnet-evm `master` vs coreth `master`: subnet-evm is missing the extra file https://github.com/ava-labs/coreth/blob/b6b4dfbc4bfc7f17e4cfe5f3c0fb44944176c884/metrics/init_test.go (which is present in geth as well)
- subnet-evm `master` vs geth `v1.13.14`: subnet-evm has many deletions, very few minor unneeded changes and the new non-conflicting file [`metrics/prometheus/prometheus.go`](https://github.com/ava-labs/subnet-evm/blob/387bbc1cab40b40cd06d87d38db71b425df2d01e/metrics/prometheus/prometheus.go)

Therefore:

- file kept and refactored `metrics/prometheus/prometheus.go` with the `Gatherer` implementation we use
- file kept and refactored `metrics/prometheus/prometheus_test.go`
- new file `metrics/prometheus/interfaces.go` added for refactoring
- geth global variable `metrics.Enabled` is set to true in `plugin/evm.VM.initializeMetrics`

Note for the two files refactored (with fixes as well), this was done in some intermediary step in https://github.com/ava-labs/libevm/pull/103, so I decided to bring this over here so it doesn't get trashed.

## How this was tested

CI passing

## Need to be documented?

No

## Need to update RELEASES.md?

Not really, since it *should* not change anything 🙏

PR #880, opened 2025-03-20
fix: use `libevm/metrics` package

ARR4N:
## Why this should be merged

#745 migrated to upstream metrics and added `metrics.Enabled` inside `plugin/evm`. However the flag name is misleading and actually disables metric _constructors_, thus disabling all metrics.

## How this works

Uses `libevm/metrics` instead of `go-ethereum/metrics` as ava-labs/libevm#165 enabled constructors by default.

## How this was tested

New unit test.

## Need to be documented?

No

No


qdm12 (approved):
I think we need this in the end, just need to add an exclusion for libevm/metrics in the goleak detector.

metrics/prometheus/enabled_test.go
```diff
@@ -0,0 +1,29 @@
+// (c) 2021-2025, Ava Labs, Inc. All rights reserved.
```
qdm12, line 1: nit

```suggestion
// (c) 2025, Ava Labs, Inc. All rights reserved.
```

qdm12 (approved):

darioush (approved):

PR #882, opened 2025-03-21
chore(ethclient): make code similar to geth code

qdm12:
## Why this should be merged

Easier geth sync of the code.

Work was done as a side effect, whilst investigating the coreth x avalanchego x subnet-evm ethclient dependency issue.

## How this works

- Move coreth specific code to ethclient_ext.go
- Unexport `ToBlockNumARg` and just copy it to `ethclient/corethclient`, to reduce diffs
- add json-ignored `Withdrawals` field to `rpcBlock` to reduce diffs
- Change client -> Client and remove Client interface: that requires avalanchego to change 3-4 lines max from `ethclient.Client` to `*ethclient.Client`. Anyway there is no point to have an interface defined here, it should return and expose a concrete type.
- add geth code client methods even if unused
- make client `SyncProgress` method a bit more similar to geth

## How this was tested

CI passing

## Need to be documented?

No

No


ethclient/ethclient.go
```diff
@@ -193,19 +103,26 @@ func (ec *client) BlockByHash(ctx context.Context, hash common.Hash) (*types.Blo
...
 	return uint64(result), err
 }
 
+// PeerCount returns the number of p2p peers as reported by the net_peerCount method.
+func (ec *Client) PeerCount(ctx context.Context) (uint64, error) {
+	var result hexutil.Uint64
+	err := ec.c.CallContext(ctx, &result, "net_peerCount")
+	return uint64(result), err
+}
+
```
ceyonur, line 123: Can we remove (comment out) this? I think it should always be returning an err

ethclient/ethclient.go
```diff
@@ -781,3 +693,55 @@ func toCallArg(msg ethereum.CallMsg) interface{} {
 	}
 	return arg
 }
+
+// rpcProgress is a copy of SyncProgress with hex-encoded fields.
+type rpcProgress struct {
```
ceyonur, line 698: IMO If removing these is not an option, we should at least keep them commented out

ethclient/ethclient_ext.go
ceyonur, line 1: noice 👍

ceyonur (commented):
generally LGTM, just not %100 sure about keeping unused/non-working code

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.23.9
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.12.1
-	github.com/ava-labs/avalanchego v1.13.1-rc.2
+	github.com/ava-labs/avalanchego v1.13.1-rc.2.0.20250523102010-9a1b43c7c765
```
ceyonur, line 7: why did we need to bump the avalanchego?

ethclient/ethclient.go
```diff
@@ -708,31 +632,19 @@ func (ec *client) EstimateGas(ctx context.Context, msg ethereum.CallMsg) (uint64
...
+func (ec *Client) SendTransaction(ctx context.Context, tx *types.Transaction) error {
 	data, err := tx.MarshalBinary()
 	if err != nil {
 		return err
 	}
 	return ec.c.CallContext(ctx, nil, "eth_sendRawTransaction", hexutil.Encode(data))
 }
 
-func ToBlockNumArg(number *big.Int) string {
+func toBlockNumArg(number *big.Int) string {
```
ceyonur, line 647: I slightly think it's an easier migration to have a diff that only exports the function rather than having a copied code.

plugin/evm/customethclient/ethclient_ext.go
ceyonur, line 1: can we rename this file since we have 2 same filenames now?

ethclient/ethclient.go
qdm12: This is geth unused code, let's keep it as is (that's how it'll be in libevm later on)

ethclient/ethclient.go
qdm12: Eh disagree, that'll make upgrading geth more complicated and the diffs with geth would be harder to see (if we keep on commenting things around). It doesn't hurt to have unused symbols in the code I would say

qdm12: Keeping it as is since it's unexported

ethclient/ethclient.go
qdm12: COmmented out!

PR #883, opened 2025-03-21
chore(interfaces): use upstream code

qdm12:
## Why this should be merged

Use upstream code when possible and reduce `interfaces` package to coreth-specific symbols.

This was found:

- whilst investigating the coreth x avalanchego x subnet-evm ethclient dependency issue, I noticed `interfaces/interfaces.go` was very similar to geth code with just 2 extra interfaces added, so this is my 2 minutes of messing around to improve that file.
- when reviewing libevm [in this comment](https://github.com/ava-labs/coreth/pull/820#discussion_r2005585635). This just extends a tiny bit of work done in libevm to a tiny bit more, so we're done with the interfaces package now.

## How this works

- import interfaces from github.com/ava-labs/libevm directly
- import NotFound error from github.com/ava-labs/libevm directly
- Keep our 2 coreth-specific interfaces to interfaces/interfaces.go


interfaces/interfaces.go
ARR4N, line 1: Why not delete this file and import `libevm` everywhere? The two `Accepted*` types could stay in this `interfaces` package (which doesn't exist in `libevm` and `geth` so won't conflict). Since they're only used a few files, you could easily do a global find-replace and then revert just the files that shouldn't have been changed.

```shell
find -iname '*.go' -print0 | xargs -0 gsed -i 's|"github.com/ava-labs/coreth/interfaces"|interfaces "github.com/ava-labs/libevm"|'
```

Then fix these:
```
./accounts/abi/bind/backend.go
./accounts/abi/bind/backends/simulated.go
./accounts/abi/bind/base.go
./accounts/abi/bind/base_test.go
./ethclient/ethclient.go
./ethclient/simulated/backend.go
```

ARR4N (changes requested):

interfaces/interfaces.go
qdm12: Great idea, thanks!
I did, to reduce diffs:
- alias github.com/ava-labs/libevm as `interfaces` to reduce diffs
- alias github.com/ava-labs/coreth/interfaces as `custominterfaces`

Let me know if you feel we should have them as other names (i.e. `ethereum` and `interfaces`, or `libevm` and `interfaces`, etc.)

ARR4N (approved):
You could, optionally, alias `libevm` (the package) as `ethereum` (it's actually [still named that](https://github.com/ava-labs/libevm/blob/87a2d57b9686fc406276d0c65ed046bc3149ff27/interfaces.go#L18)) and remove the alias to `custominterfaces`.

More diffs but closer alignment to upstream.

qdm12:
> More diffs but closer alignment to upstream.

Agreed, done 👍 !

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.23.6
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.12.1
-	github.com/ava-labs/avalanchego v1.12.3-0.20250314173032-a3d150e0aad0
+	github.com/ava-labs/avalanchego v1.13.0-fuji-rc.2.0.20250320111141-fef07a6db96b
```
darioush, line 7: is this the right version to depend on?

qdm12: This is avalanchego from my libevm branch (based on v1.13.0-fuji tag), notably using ava-labs/libevm instead of ava-labs/coreth/interfaces, which is accidentally relevant in this PR.
However, good comment 💯 , this made me realise, we should not bump the dependency.
On top of this, subnet-evm's libevm branch already depends on that version, although its master branch go.mod avalanchego version is pretty different than coreth's one ([diffs](https://github.com/ava-labs/avalanchego/compare/v1.12.3-name-fortuna.0...v1.12.3-rc.1))

In conclusion, I will rebase that avalanchego libevm branch such that:

- coreth-specific avalanchego libevm branch is based off `v1.12.3-name-fortuna.0`
- subnet-evm -specific avalanchego libevm branch is based off `v1.12.3-rc.1`

darioush (approved):

go.mod
qdm12: This PR no longer modifies the go.mod dependencies

ceyonur (approved):

PR #894, opened 2025-03-26
nit: check genesis commit error

darioush:
I want to check this error since I spent some time debugging yesterday and this would have made it obvious.

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -301,7 +301,7 @@ func (g *Genesis) toBlock(db ethdb.Database, triedb *triedb.Database) *types.Blo
 	}
 
 	if _, err := statedb.Commit(0, false); err != nil {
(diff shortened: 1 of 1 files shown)


core/genesis.go
```diff
@@ -300,7 +300,9 @@ func (g *Genesis) toBlock(db ethdb.Database, triedb *triedb.Database) *types.Blo
 		}
 	}
 
-	statedb.Commit(0, false)
+	if _, err := statedb.Commit(0, false); err != nil {
+		panic(fmt.Sprintf("unable to commit genesis block: %v", err))
```
qdm12, line 304: nit just to differentiate from the same panic message below

```suggestion
		panic(fmt.Sprintf("unable to commit genesis block to statedb: %v", err))
```

qdm12 (approved):

ceyonur (approved):

PR #665, opened 2024-09-27
miner: upstream alignment

darioush:

PR #903, opened 2025-04-03
chore(miner): textual alignment with upstream

qdm12:
## Why this should be merged

Updated version of #665


qdm12:
Force push rebasing this from master to libevm

miner/worker.go
```diff
@@ -131,132 +324,415 @@ func (w *worker) setEtherbase(addr common.Address) {
...
-	// Note: in order to support asynchronous block production, blocks are allowed to have
-	// the same timestamp as their parent. This allows more than one block to be produced
-	// per second.
-	if parent.Time >= timestamp {
-		timestamp = parent.Time
-	}
+func (w *worker) setGasCeil(ceil uint64) {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	// w.config.GasCeil = ceil
```
qdm12, line 337: why is this commented compared to geth code?

miner/worker.go
```diff
@@ -131,132 +324,415 @@ func (w *worker) setEtherbase(addr common.Address) {
...
+// The returned values can be nil in case the pending block is not initialized.
+func (w *worker) pendingBlockAndReceipts() (*types.Block, types.Receipts) {
+	w.snapshotMu.RLock()
+	defer w.snapshotMu.RUnlock()
+	return w.snapshotBlock, w.snapshotReceipts
+}
+
+// start sets the running status as 1 and triggers new work submitting.
+func (w *worker) start() {
+	// w.running.Store(true)
```
qdm12, line 391: Why is this commented compared to geth code?

miner/worker.go
```diff
@@ -459,71 +972,374 @@ func (w *worker) commitTransactions(env *environment, plainTxs, blobTxs *transac
...
-func (w *worker) commit(env *environment) (*types.Block, error) {
-	if params.GetRulesExtra(env.rules).IsDurango {
+// generateParams wraps various of settings for generating sealing task.
+type generateParams struct {
+	timestamp   uint64            // The timestamp for sealing task
+	forceTime   bool              // Flag whether the given timestamp is immutable or not
+	parentHash  common.Hash       // Parent block hash, empty means the latest chain head
+	coinbase    common.Address    // The fee recipient address for including transaction
+	random      common.Hash       // The randomness generated by beacon chain, empty before the merge
+	withdrawals types.Withdrawals // List of withdrawals to include in block.
```
qdm12, line 1000: I've added that field now that we use libevm and it's present 🤷 
Reduces a bunch of diffs, even if totally unused...

plugin/evm/vm.go
```diff
@@ -641,7 +641,6 @@ func (vm *VM) Initialize(
 }
 
 func (vm *VM) initializeMetrics() error {
-	metrics.Enabled = true
```
qdm12, line 644: this was removed since it was causing data race conditions (duh global variable...);
It's not needed anymore since it's true by default and never disabled anywhere.

miner/worker.go
qdm12: Every code block running when w.running is true is geth specific, that's why we set this to always false.

miner/worker.go
```diff
@@ -1,13 +1,3 @@
-// (c) 2019-2020, Ava Labs, Inc.
```
qdm12, line 1: re-add this for now

PR #904, opened 2025-04-03
test(core/genesis): fix UT to be sensitive to `SetEthUpgrades`

darioush:
Prior to this change, the unit test will still pass if the line:
```
params.SetEthUpgrades(storedcfg)
```
is removed from genesis.go.

This is because `RewindToBlock` and `RewindToTime` are overloaded where 0 both means "rewind to genesis" and that "block number or timestamp is not used for setting the fork" (i.e, the other one is used).

If the line is removed it will conflict on the DAO fork block / flag, then rewind to genesis, then the UT will pass, which defeats the purpose of this UT.


ARR4N (approved):
Please can you add a comment along with the change, explaining why the values are set as they are?

michaelkaplan13 (approved):

PR #916, opened 2025-04-10
chore: upgrade version to v0.15.0

qdm12:
## Why this should be merged

Do not merge this, this is only to create the release tag v0.15.0


qdm12:
Not needed, we'll just use -rc releases

PR #924, opened 2025-04-17
Simplify `eth_suggestPriceOptions` handling of `baseFee`s

StephenButtolph:
## Why this should be merged

1. Fortuna has been activated, so the `.IsFortuna` check is no longer needed.
2. Previously, the base fee was set between [80%, 120%] of the chain values (with min and max values enforced). Post activation of Fortuna, the price has been extremely stable. This means that issuing a transaction at 80% of the base fee is all but guaranteed not to be included on chain. The base fee is now always set at 200% of the chain base fee.

## How this works

- Cleans up code that isn't needed.
- The base fee is not the fee that is actually paid by the end user, so it is safe to over-estimate.

## How this was tested

- [x] Locally
- [X] Unit tests

## Need to be documented?

I don't think the configs that are being changed are documented anywhere... But if they are, they should be updated.

Done.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -5,6 +5,8 @@
 - Major refactor to use [`libevm`](https://github.com/ava-labs/libevm) for EVM execution, database access, types & chain configuration. This improves maintainability and enables keeping up with upstream changes more easily.
 - Add metrics for ACP-176
 
+- Removed the `"price-options-max-base-fee"` config flag
+
 ## [v0.15.0](https://github.com/ava-labs/coreth/releases/tag/v0.15.0)
 
 - Bump golang version to v1.23.6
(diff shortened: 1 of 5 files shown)


RELEASES.md
```diff
@@ -13,7 +15,7 @@
 - Add `GasTarget` to the chain config to allow modifying the chain's `GasTarget` based on the ACP-176 rules
 
 - Added `eth_suggestPriceOptions` API to suggest gas prices (slow, normal, fast) based on the current network conditions
-- Added `"po-slow-fee-percentage"`, `"po-fast-fee-percentage"`, `"po-max-base-fee"`, and `"po-max-tip"` config flags to configure the new `eth_suggestPriceOptions` API
+- Added `"price-options-slow-fee-percentage"`, `"price-options-fast-fee-percentage"`, `"price-options-max-base-fee"`, and `"price-options-max-tip"` config flags to configure the new `eth_suggestPriceOptions` API
```
StephenButtolph, line 18: These were incorrect previously.

alarso16 (approved):
lgtm, not sure 2 * baseFee is necessary, but is certainly sufficient

internal/ethapi/api.coreth.go
```diff
@@ -62,38 +58,21 @@ func (s *EthereumAPI) SuggestPriceOptions(ctx context.Context) (*PriceOptions, e
...
+		new(big.Int).SetUint64(config.MaxTip),
+		new(big.Int).SetUint64(config.SlowFeePercentage),
+		new(big.Int).SetUint64(config.FastFeePercentage),
 	)
-	slowGasFee := new(big.Int).Add(baseFees.slow, gasTips.slow)
-	normalGasFee := new(big.Int).Add(baseFees.normal, gasTips.normal)
-	fastGasFee := new(big.Int).Add(baseFees.fast, gasTips.fast)
+
+	// Double the baseFee estimate without modifying the original variable.
+	baseFee = new(big.Int).Lsh(baseFee, 1)
```
qdm12, line 71: Should we name it `baseFeeDouble` to avoid any confusion?

qdm12 (approved):

ceyonur (approved):
overall lgtm

PR #937, opened 2025-04-22
Remove deadcode from old vm interfaces

aaronbuchwald:
## Why this should be merged

This PR removes deadcode from the Coreth VM that's left over from old VM interfaces.

## How this works

Deletes unused code.

## How this was tested

Existing test coverage.

## Need to be documented?

No

No

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -23,6 +23,7 @@ import (
 	"github.com/ava-labs/avalanchego/network/p2p/gossip"
 	"github.com/ava-labs/avalanchego/upgrade"
 	avalanchegoConstants "github.com/ava-labs/avalanchego/utils/constants"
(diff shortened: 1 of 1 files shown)


rkuris (approved):

plugin/evm/vm.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/ava-labs/avalanchego/network/p2p/gossip"
 	"github.com/ava-labs/avalanchego/upgrade"
 	avalanchegoConstants "github.com/ava-labs/avalanchego/utils/constants"
+	"github.com/ava-labs/avalanchego/utils/logging"
```
qdm12, line 26: super-nit you could move this back below where it was to have **zero** line added

PR #938, opened 2025-04-23
Update ref from appSender to vm.p2pSender to align with tests

aaronbuchwald:
## Why this should be merged

This PR updates a reference to the `appSender` argument passed in to `vm.Initialize(...)` to instead use `vm.p2pSender`.

This is because the tests explicitly override the value of `vm.p2pSender` [here](https://github.com/ava-labs/coreth/blob/6605c4d77123864f0b8957bae3c98479fbb0c184/plugin/evm/tx_gossip_test.go#L61) prior to `Initialize` and checking if the field has already been set [here](https://github.com/ava-labs/coreth/blob/6605c4d77123864f0b8957bae3c98479fbb0c184/plugin/evm/vm.go#L543).

This behavior is currently documented here: https://github.com/ava-labs/coreth/blob/6605c4d77123864f0b8957bae3c98479fbb0c184/plugin/evm/vm.go#L308

This PR changes it to make sure we use the potentially overridden value as an argument to constructing the p2p network helper as well. This does not change the test status and makes the existing behavior easier to follow ie. override all usage of `appSender` instead of only half the time.

## How this works

as described above

## How this was tested

Existing tests

## Need to be documented?

No

No


JonathanOppenheimer:
`p2pSender` has been factored out of `vm` making this PR moot, so I will close it. @aaronbuchwald let me know if you'd like to re-address it.

PR #941, opened 2025-04-29
Log transaction-chain mismatch

yacovm:
## Why this should be merged

If the user sends a transaction with the wrong chain ID, it is returned with a cryptic error.

This commit adds a logging event that informs if a user tries to send a transaction with the wrong chain ID.

## How this works

Adds a log entry in case the chain ID mismatches with the transaction's chain ID.

## How this was tested

Ran it locally with error logging level and then reverted back to trace logging level.


alarso16 (approved):

core/txpool/legacypool/legacypool.go
```diff
@@ -683,6 +683,9 @@ func (pool *LegacyPool) validateTxBasics(tx *types.Transaction, local bool) erro
 		opts.MinTip = new(big.Int)
 	}
 	if err := txpool.ValidateTransaction(tx, pool.currentHead.Load(), pool.signer, opts); err != nil {
+		if errors.Is(err, txpool.ErrInvalidSender) && pool.chainconfig.ChainID.Cmp(tx.ChainId()) != 0 {
```
michaelkaplan13, line 686: IIUC, this log condition will only be met for EIP-155 type transactions that were signed using an incorrect chain ID (where the chain ID isn't explicitly included in the tx itself, but using the wrong chain ID results in an invalid signature). 

Should we remove the "invalid sender" check, and instead just have the chain ID check? I think that way it will apply for any tx types using the wrong chain ID, including EIP-2930, EIP-1559, etc.

yacovm: > IIUC, this log condition will only be met for EIP-155 type transactions that were signed using an incorrect chain ID (where the chain ID isn't explicitly included in the tx itself, but using the wrong chain ID results in an invalid signature).
> 
> Should we remove the "invalid sender" check, and instead just have the chain ID check? I think that way it will apply for any tx types using the wrong chain ID, including EIP-2930, EIP-1559, etc.

Sure, done.

qdm12 (approved):

core/txpool/legacypool/legacypool.go
```diff
@@ -683,6 +683,9 @@ func (pool *LegacyPool) validateTxBasics(tx *types.Transaction, local bool) erro
 		opts.MinTip = new(big.Int)
 	}
 	if err := txpool.ValidateTransaction(tx, pool.currentHead.Load(), pool.signer, opts); err != nil {
+		if pool.chainconfig.ChainID.Cmp(tx.ChainId()) != 0 {
+			log.Trace("Transaction targets wrong chain", "chainID", tx.ChainId(), "expected", pool.chainconfig.ChainID)
+		}
```
StephenButtolph, line 688: I'm not sure if this is the right place to put this (I think we are trying to minimize the diff with geth in the core package.. Right? cc: @ceyonur)

If we do go with this, we shouldn't log this for unprotected transactions.
```suggestion
		if tx.Protected() && pool.chainconfig.ChainID.Cmp(tx.ChainId()) != 0 {
			log.Trace("Transaction targets wrong chain", "chainID", tx.ChainId(), "expected", pool.chainconfig.ChainID)
		}
```

yacovm: > If we do go with this, we shouldn't log this for unprotected transactions.


why?

JonathanOppenheimer:
Closed after conferring with @yacovm

PR #942, opened 2025-04-30
Update avalanchego dependency for new caches

StephenButtolph:
## Why this should be merged

Updates avalanchego version.


StephenButtolph:
Seems to have been implemented in a different PR

PR #943, opened 2025-05-04
reduce customtypes import

ceyonur:
## Why this should be merged

Importing `customtypes` and `params` pkg causes side-effects of `libevm`. We should reduce importing of those pkgs in custom-facing client code.

AvalancheGo Follow-up PR: https://github.com/ava-labs/avalanchego/pull/3918

## How this works

This pull request refactors logging functionality by consolidating and renaming the logging package, alongside minor code improvements and cleanup. The most significant changes include replacing the `log` and `customtypes` packages with a unified `customlog` package, updating imports and references accordingly, and simplifying variable declarations for consistency.

### Logging Refactor:

* Renamed the `log` package to `customlog` across multiple files, including `format.go`, `handler.go`, and `logger.go`, and updated their package declarations and references. 
* Replaced references to `customtypes.FlattenLogs` with `customlog.FlattenLogs` in `core/blockchain.go`, `eth/filters/filter.go`, and other files. 

### Code Cleanup and Simplification:

* Removed unused fields (`Version` and `BlockExtraData`) and associated logic from `ethclient/ethclient.go` to streamline block handling, to prevent importing `customtypes` and libevm side-effects.

## How this was tested

Added to `imports_test`

## Need to be documented?

No

Probably


ethclient/ethclient.go
```diff
@@ -70,7 +69,7 @@ var (
 )
 
 // Client defines interface for typed wrappers for the Ethereum RPC API.
-type Client interface {
+type Client interface { // Q: why do we need this?
```
ceyonur, line 72: @felipemadero Can you remember why we needed this? I'd like to remove this interface (or move it elsewhere) to revert extra diffs introduced here.

ethclient/ethclient.go
```diff
@@ -267,18 +264,10 @@ func (ec *client) getBlock(ctx context.Context, method string, args ...interface
...
-	block := types.NewBlockWithHeader(head).WithBody(
-		types.Body{
-			Transactions: txs,
-			Uncles:       uncles,
-		})
-	extra := &customtypes.BlockBodyExtra{
-		Version: body.Version,
-		ExtData: (*[]byte)(body.BlockExtraData),
-	}
-	customtypes.SetBlockExtra(block, extra)
```
ceyonur, line 280: This will remove extra fields in header + block body, but since with libevm they were already removed as the first-level fields, and moved to extra pseudo fields, if there was any breaking change in customer code we would have noticed. 

However if we still think we should maintain those extra fields, we can introduce a formatter which could introduce a custom `rpcBlock` to unmarshal the block from so that we can redefine it in `corethclient`. In that case `corethclient` should wrap the `ethclient` and set the extra field. It could still a bit problematic to import `corethclient` pkg in customer code, but at least we can unblock the `ethclient` pkg import.

On the other hand, I think we can completely remove this file and directly import it from libevm (or alias here), since we really don't do much here other than providing some shortcut APIs for `accepted` functions.

plugin/evm/imports_test.go
```diff
@@ -54,10 +56,12 @@ func TestMustNotImport(t *testing.T) {
...
 		// possible to do so for both coreth and subnet-evm, where the client may
 		// wish to connect to multiple chains.
-		"plugin/evm/atomic": {"core", "core/vm", "params"},
-		"plugin/evm/client": {"core", "core/vm", "params"},
-		"plugin/evm/config": {"core", "core/vm", "params"},
-		"plugin/evm/header": {"core", "core/vm", "params"},
+		"plugin/evm/atomic": {"plugin/evm/customtypes", "params"},
+		"plugin/evm/client": {"plugin/evm/customtypes", "params"},
+		"plugin/evm/config": {"plugin/evm/customtypes", "params"},
+		"plugin/evm/header": {"core", "core/vm", "params"}, // Q: why do we need this?
```
ceyonur, line 62: cc @qdm12, do you know why we need this?

I'm not sure if the header should be tested here as it's really the part of the custom code and should not be imported from outside.

ethclient/ethclient.go
qdm12: It's already removed in #882 😉 But we need to remove its import from avalanchego first I think (which is done in https://github.com/ava-labs/avalanchego/pull/3923)

plugin/evm/imports_test.go
qdm12: I don't 100% recall, but I would guess it's due to indirect imports. The header package is imported all over the place internally, so it's likely imported by other repositories

plugin/evm/imports_test.go
```diff
@@ -54,10 +56,12 @@ func TestMustNotImport(t *testing.T) {
...
+		"ethclient":         {"plugin/evm/customtypes", "params"},
+		"warp":              {"plugin/evm/customtypes", "params"},
```
qdm12, line 64: Why the warp addition?

plugin/evm/imports_test.go
```diff
@@ -15,7 +15,7 @@ import (
 // its recursive package imports (including itself) in the same format.
 func getDependencies(packageName string) (map[string]struct{}, error) {
 	// Configure the load mode to include dependencies
-	cfg := &packages.Config{Mode: packages.NeedDeps | packages.NeedImports | packages.NeedName | packages.NeedModule}
+	cfg := &packages.Config{Mode: packages.NeedImports | packages.NeedName}
```
qdm12, line 18: Why the change?

plugin/evm/imports_test.go
```diff
@@ -38,7 +38,9 @@ func getDependencies(packageName string) (map[string]struct{}, error) {
 	}
 
 	// Start collecting dependencies
-	collectDeps(pkgs[0])
+	for _, pkg := range pkgs {
+		collectDeps(pkg)
+	}
```
qdm12, line 43: Why were we collecting the first pkg only??

log/log_ext.go
```diff
@@ -1,6 +1,6 @@
 // (c) 2019-2025, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
-package customtypes
+package log
```
qdm12, line 3: If we do move this here, let's call it `customlog` since we import it as `customlog`

ethclient/ethclient.go
ceyonur: > I'm not sure I understand; we have extra logic here to stay retro-compatible despite the libevm change. I don't think we did introduce any change in coreth with the libevm branch?

If there were any custom client code using `ExtData`, after libevm transition their code must have broken (i.e `block.ExtraData => customtypes.GetExtraData(block)`. After all it's not possible to access the `ExtraData` same as before libevm.

> Given avalanchego needs custom methods such as EstimateBaseFee, we would have to import this corethclient, so this doesn't really unblock the avalanchego situation I think?

We can still keep them in `ethclient` but if it's a must to access extra fields in header and body from the client, we can introduce a formatter/custom `rpcBlock` in `corethclient` and wrap the `ethclient` in `corethclient`. At least this would unblock `ethclient` import itself, and if one wants to access extra fields they can use `corethclient` instead.

but yea given that we might need `EstimateBaseFee` etc we won't probably be able to remove `ethclient` entirely.

plugin/evm/imports_test.go
ceyonur: I think this test should ensure we're not directly/indirectly causing libevm side-effects in customer-facing pkgs. I don't think `evm/header` is one of these pkgs. The test should be already checking the indirect imports as well, so I think we can remove the case here.

plugin/evm/imports_test.go
ceyonur: Warp has a potential customer-facing client code.

plugin/evm/imports_test.go
ceyonur: I don't think we need to add modes for module and deps as we don't use them.

plugin/evm/imports_test.go
ceyonur: No idea. I don't think it was causing any problems as we directly referenced the pkgs itself. But I found this was not working with wildcards like `plugin/evm/...` etc to capture all subpackages.

plugin/evm/imports_test.go
```diff
@@ -54,10 +56,10 @@ func TestMustNotImport(t *testing.T) {
 		// Importing these packages configures libevm globally and it is not
 		// possible to do so for both coreth and subnet-evm, where the client may
 		// wish to connect to multiple chains.
-		"plugin/evm/atomic": {"core", "core/vm", "params"},
-		"plugin/evm/client": {"core", "core/vm", "params"},
-		"plugin/evm/config": {"core", "core/vm", "params"},
-		"plugin/evm/header": {"core", "core/vm", "params"},
```
qdm12, line 60: If this is pressing, let's keep these two in here.
There might be a reason for these to not import those packages, we should not remove them without investigating in depth their indirect dependencies (and a bit of PITA to do)

ceyonur: I've based my changes on this comment:
```
// The following sub-packages of plugin/evm must not import core, core/vm
// so clients (e.g., wallets, e2e tests) can import them without pulling in
// the entire VM logic.
// Importing these packages configures libevm globally and it is not
// possible to do so for both coreth and subnet-evm, where the client may
// wish to connect to multiple chains.
```

I'm pretty sure `config` was required by `client` pkg so that's why it was here. I restored the header pkg although it's a bit weird because it actually imports the `customtypes`

ethclient/ethclient.go
qdm12: Ideally interfaces should be defined in the "accepting side" (so the cli) instead of being returned. See [accepting interfaces, returning concrete types](https://duncanleung.com/go-idiom-accept-interfaces-return-types/) 😉

customlog/handler.go
```diff
@@ -1,4 +1,4 @@
-package log
```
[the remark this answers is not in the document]
qdm12: But when it comes to the existing `log` package, we should keep it to the `log` directory. I don't see the reason why we are moving `log/*.go` files to `plugin/evm/customlog`. These are quite similar to libevm/geth code and should be "offloaded to libevm" properly later and separately.

plugin/evm/imports_test.go
```diff
@@ -54,10 +56,11 @@ func TestMustNotImport(t *testing.T) {
...
 		// possible to do so for both coreth and subnet-evm, where the client may
 		// wish to connect to multiple chains.
-		"plugin/evm/atomic": {"core", "core/vm", "params"},
-		"plugin/evm/client": {"core", "core/vm", "params"},
-		"plugin/evm/config": {"core", "core/vm", "params"},
+		"plugin/evm/atomic": {"plugin/evm/customtypes", "params"},
+		"plugin/evm/client": {"plugin/evm/customtypes", "params"},
 		"plugin/evm/header": {"core", "core/vm", "params"},
+		"ethclient":         {"plugin/evm/customtypes", "params"},
+		"warp":              {"plugin/evm/customtypes", "params"},
```
michaelkaplan13, line 63: Should we still also be checking for "core" and "core/vm" not being imported by these?

ethclient/ethclient.go
ceyonur: I think we can delete/move this interface in your PR. (but we should indeed remove it from here and reduce the diff in client code)

ethclient/ethclient.go
michaelkaplan13: There hasn't been a release since the libevm changes were merged, so are we sure we're confident that there aren't any downstream breaking implications of the changes that just haven't been discovered yet?

plugin/evm/imports_test.go
ceyonur: I don't think the intention was ever to check `core` and `core/vm` but rather check the libevm registration in `params` and `customtypes` (indirectly)

customlog/handler.go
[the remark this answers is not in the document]
ceyonur: I was under impression that we were following convention of "customxxx" pkgs for those we extended. I'm fine with any naming/location for those files. They're kind-of extending the existing `logs` pkgs, so I kept it in the same level and renamed it to`customlog`. I don't think this is a blocker for this PR given that we have not offloaded them to libevm yet.

plugin/evm/imports_test.go
michaelkaplan13: Others would know better than me, but this comment seems to suggest the intention was to explicitly check that `core` and `core/vm` are not imported by these packages, right?
```
        // The following sub-packages of plugin/evm must not import core, core/vm
		// so clients (e.g., wallets, e2e tests) can import them without pulling in
		// the entire VM logic.
```

ethclient/ethclient.go
ceyonur: hmm we did release a Coreth but might not have released an accompanying AvalancheGo version with it. In any case libevm changes will already cause a breaking change since those fields will not be directly available in `types.Block`

1- I'm %99 percent these fields are not directly being used from client and even if they're, using an older coreth version would still work.
2- If we absolutely want to maintain accessibility of these extra fields from client code, we should still remove it from ethclient pkg and move it to `gethclient` or expect a formatter interface as arg passed to client. (which we can provide from another pkg). `ethclient` itself is being widely used for other non-breaking methods, so we should decouple libevm side-effect from it.

plugin/evm/imports_test.go
qdm12: Libevm registration happens in the following packages:

- `params`
- `plugin/evm/customtypes`
- `core`

and `core/vm` no longer exists.

So the list of packages to avoid importing for libevm double registration should be `{"core", "plugin/evm/customtypes", "params"}`.

qdm12: Regarding the removal of `plugin/evm/config`, it was originally added in https://github.com/ava-labs/coreth/pull/734 - maybe we should keep it? As long as it doesn't bother us as well 🤷

plugin/evm/imports_test.go
qdm12: Oh because for a single package specified, this is always of length 1. I suppose using `...` can return multiple ones, why not the change then!

ethclient/ethclient.go
michaelkaplan13: You'd know best on if it's being relied upon or not, so definitely defer to your judgement 👍 Just wanted to make sure we weren't basing that solely on the release already being out and broadly used.

plugin/evm/imports_test.go
ceyonur: I'm pretty sure `params` pkg was the reason, but I don't want to block this any further and reverting any extra changes.

ceyonur: and also by transition `core` cannot be ever imported because we already marked `customtypes` and `params`(which are already imported in `core`) as forbidden. but as I said I don't want to drag this conversation anymore.

tsachiherman (commented):
could you fix the 
```
no required module provides package github.com/ava-labs/coreth/plugin/evm/customlogs; to add it:
```
lint message ?

ceyonur:
> could you fix the
> 
> ```
> no required module provides package github.com/ava-labs/coreth/plugin/evm/customlogs; to add it:
> ```
> 
> lint message ?
> 
> maybe you forgot to add `customlogs` to this PR ?

it was gitignored

plugin/evm/imports_test.go
michaelkaplan13: Sounds like I was just misunderstanding the tests intent based on the comment or missing context potentially, sorry. But sounds good

michaelkaplan13 (approved):

.gitignore
```diff
@@ -23,7 +23,7 @@ awscpu
 # ignore GoLand metafiles directory
 .idea/
 
-*logs/
+logs/
```
qdm12, line 26: nit maybe

```suggestion
*logs/
!*customlogs/
```

so that other logs/ directories are ignored still?

plugin/evm/imports_test.go
```diff
@@ -54,10 +56,12 @@ func TestMustNotImport(t *testing.T) {
 		// Importing these packages configures libevm globally and it is not
 		// possible to do so for both coreth and subnet-evm, where the client may
 		// wish to connect to multiple chains.
-		"plugin/evm/atomic": {"core", "core/vm", "params"},
-		"plugin/evm/client": {"core", "core/vm", "params"},
-		"plugin/evm/config": {"core", "core/vm", "params"},
+		"plugin/evm/atomic": {"core", "core/vm", "plugin/evm/customtypes", "params"},
+		"plugin/evm/client": {"core", "core/vm", "plugin/evm/customtypes", "params"},
+		"plugin/evm/config": {"core", "core/vm", "params", "plugin/evm/customtypes"},
```
qdm12, line 61: nit we can remove core/vm, there is no go file there anymore

```suggestion
		"plugin/evm/atomic": {"core", "plugin/evm/customtypes", "params"},
		"plugin/evm/client": {"core", "plugin/evm/customtypes", "params"},
		"plugin/evm/config": {"core", "params", "plugin/evm/customtypes"},
```

qdm12 (approved):

PR #949, opened 2025-05-06
chore(plugin/evm): remove unneeded code `BuildGenesisTest` and associated

StephenButtolph:
## Why this should be merged

Code is only used in tests.

## How this works

Deletes it.

## How this was tested

Test can't fail if it doesn't exist.

## Need to be documented?

No

No


michaelkaplan13 (approved):

plugin/evm/vm_test.go
```diff
@@ -177,25 +176,6 @@ func newPrefundedGenesis(
 	}
 }
 
-// BuildGenesisTest returns the genesis bytes for Coreth VM to be used in testing
-func BuildGenesisTest(t *testing.T, genesisJSON string) []byte {
```
qdm12, line 181: Why did we have this encode->decode in the first place? 🤔

qdm12 (approved):

PR #951, opened 2025-05-07
add ethclient hooks and examples

ceyonur:
## Why this should be merged

Libevm transition changes the extra fields as top-level fields to psuedo fields which can only be retrieved with extra function calls from customtypes. #943 removes extra types from ethclient since using `customtypes` pkg causes side effects. 

This PR restores the extra type retrieval in `customclient` pkg by extending the ethclient.

## How this works

* Add a hook interface to modify the returned block
* Implement the hook in another pkg to prevent importing `customtypes` in `ethclient`

## How this was tested

Locally (with test in `example_test.go`)

## Need to be documented?

AFAIK we don't have any doc for clients. A reference in released.md should be sufficient.

Updated

### DIFF
--- a/ethclient/customclient/ethclient_ext.go
+++ b/ethclient/customclient/ethclient_ext.go
@@ -1,9 +1,10 @@
 // (c) 2019-2025, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
-package corethclient
+package customclient
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 
@@ -12,9 +13,36 @@ import (
 
 	"github.com/ava-labs/coreth/ethclient"
 	"github.com/ava-labs/coreth/plugin/evm/customtypes"
+	"github.com/ava-labs/coreth/rpc"
 )
 
 var _ ethclient.BlockHook = (*extBlockHook)(nil)
+
+// Client wraps the ethclient.Client interface to provide extra data types (in header, block body).
+// If you want to use the standardized Ethereum RPC functionality without extra types, use ethclient.Client instead.
+type Client struct {
+	c *rpc.Client
+	ethclient.Client
+}
+
+// New creates a client that uses the given RPC client.
+func New(c *rpc.Client) *Client {
+	return &Client{c: c, Client: ethclient.NewClientWithHook(c, &extBlockHook{})}
+}
+
+// Dial connects a client to the given URL.
+func Dial(rawurl string) (*Client, error) {
+	return DialContext(context.Background(), rawurl)
+}
+
+// DialContext connects a client to the given URL with context.
+func DialContext(ctx context.Context, rawurl string) (*Client, error) {
+	c, err := rpc.DialContext(ctx, rawurl)
+	if err != nil {
+		return nil, err
+	}
+	return New(c), nil
+}
 
 // extBlockHook is a hook that is called when a block is decoded.
 type extBlockHook struct{}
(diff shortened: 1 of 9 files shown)


ethclient/customclient/ethclient_ext.go
```diff
@@ -0,0 +1,70 @@
...
+// extRpcBlock is the structure of the block extra data in the RPC response.
+// It contains the version and the block extra data.
+type extRpcBlock struct {
+	Version        uint32         `json:"version"`
+	BlockExtraData *hexutil.Bytes `json:"blockExtraData"`
+}
+
+// OnBlockDecoded is called when a block is decoded. It unmarshals the
+// extra data from the RPC response and sets it in the block with libevm extras.
+func (h *extBlockHook) OnBlockDecoded(raw json.RawMessage, block *types.Block) error {
```
qdm12, line 59: that does modify the ethclient.go code; we should instead have a json unmarshal method as the libevm blockbody hooks (we have only RLP for now) to handle this, which would allow to keep ethclient.go as it is. Then registration could occur in this customclient package I suppose 🤔 
I mean, this is ideally/long term, we can have this in the meantime, although it's a bit misleading to have this as "hook" when we want later to libevm-hook it which would be at a different place (types.Block JSON unmarshal method)

ceyonur: I think we can wrap those methods without modifying the ethclient itself. I tried to avoid having duplicate code, but if the intention is to avoid modifying the client we can copy `getBlock` and `setSenderFromServer` and implement what we need in `customclient`.

ethclient/ethclient.go
```diff
@@ -133,7 +144,12 @@ func DialContext(ctx context.Context, rawurl string) (Client, error) {
 
 // NewClient creates a client that uses the given RPC client.
 func NewClient(c *rpc.Client) Client {
-	return &client{c}
+	return &client{c, nil}
```
qdm12, line 147: nit use named fields to reduce confusion and be able to move fields around later

```suggestion
	return &client{
		c: c,
	}
```

plugin/evm/customethclient/ethclient_ext.go
```diff
@@ -0,0 +1,70 @@
+// (c) 2019-2025, Ava Labs, Inc. All rights reserved.
```
qdm12, line 1: this can be renamed to `ethclient.go` without the `_ext` suffix

plugin/evm/customethclient/example_test.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+const FujiAPIURI = "https://api.avax-test.network"
+
+func ExampleClient() {
+	ethC, err := Dial(FujiAPIURI + "/ext/bc/C/rpc")
+	if err != nil {
+		panic(err)
+	}
+	bc, err := ethC.BlockByNumber(context.Background(), nil)
```
qdm12, line 17: Maybe let's target a particular block number and convert this to a test with assertions? If it's the same block, the results should remain the same I think

plugin/evm/customethclient/ethclient_ext.go
ceyonur: let's keep it as it's actually extending the ethclient

plugin/evm/customethclient/example_test.go
ceyonur: I'm not sure about adding a UT that hits external server (api.avax). I've looked for a possible mock server but could not really find one. Given that the original ethclient does not also have any test, I think this is fine. The intention here is not to test/show how client is being called, but rather how the extra types can be extracted.

StephenButtolph (approved):

PR #952, opened 2025-05-08
Fix `SetEthUpgrades` with unscheduled activations

StephenButtolph:
## Why this should be merged

Ran into this case when trying to properly provide `upgrade.Config`s in our unit tests.

## How this works

1. Doesn't populate `ShanghaiTime` or `CancunTime` when provided `upgrade.Config`s where they are unscheduled.
2. Reports an error rather than silently dropping invalid configurations in `SetEthUpgrades`
3. Refactors `ChainConfig` initialization in `vm.Initialize` to be performed in a single location

## How this was tested

CI

## Need to be documented?

No

No


plugin/evm/vm_test.go
```diff
@@ -3774,16 +3774,16 @@ func TestSkipChainConfigCheckCompatible(t *testing.T) {
 	// use the block's timestamp instead of 0 since rewind to genesis
 	// is hardcoded to be allowed in core/genesis.go.
 	genesisWithUpgrade := &core.Genesis{}
-	require.NoError(t, json.Unmarshal([]byte(genesisJSONApricotPhase1), genesisWithUpgrade))
-	params.GetExtra(genesisWithUpgrade.Config).ApricotPhase2BlockTimestamp = utils.TimeToNewUint64(blk.Timestamp())
```
StephenButtolph, line 3778: This test previously specified a timestamp for AP2 that wasn't initially active or unscheduled... Which is explicitly not supported.

plugin/evm/vm.go
```diff
@@ -625,6 +600,42 @@ func (vm *VM) Initialize(
...
+	configExtra.AvalancheContext = extras.AvalancheContext{
+		SnowCtx: ctx,
+	}
+
+	// TODO: This check should be removed. The upgrade config is always provided
+	// in production. All tests should be updated to correctly provide the
+	// upgrade config.
+	if ctx.NetworkUpgrades != (upgrade.Config{}) {
+		configExtra.NetworkUpgrades = extras.GetNetworkUpgrades(ctx.NetworkUpgrades)
+	}
```
StephenButtolph, line 620: The removal of this check is what lead me down this rabbit hole.

params/config_extra.go
```diff
@@ -37,32 +46,48 @@ func SetEthUpgrades(c *ChainConfig) {
...
 		}
 	}
-	extra := GetExtra(c)
-	if extra.DurangoBlockTimestamp != nil {
+
+	// We only mark Shanghai and Cancun as enabled if we have marked them as
+	// scheduled.
+	if extra.DurangoBlockTimestamp != nil && *extra.DurangoBlockTimestamp < unscheduledActivation {
 		c.ShanghaiTime = utils.NewUint64(*extra.DurangoBlockTimestamp)
 	}
```
ARR4N, line 85:
```suggestion
	if durango := extra.DurangoBlockTimestamp; durango != nil && *durango < unscheduledActivation {
		c.ShanghaiTime = utils.NewUint64(*durango)
	}
```

I'd even go so far as to just use `d` for `durango` or `t` for `time` as the variable.

ARR4N (commented):
Other than my question in `parseGenesis()` re the order, the rest LGTM pending some readability improvements.

params/config_extra.go
```diff
@@ -37,32 +46,48 @@ func SetEthUpgrades(c *ChainConfig) {
...
+		// London at the initially active time. This corresponds to an intended
+		// block number of 0.
+		switch {
+		case extra.ApricotPhase2BlockTimestamp == nil:
+			return nil
+		case *extra.ApricotPhase2BlockTimestamp <= initiallyActive:
 			c.BerlinBlock = big.NewInt(0)
+		case *extra.ApricotPhase2BlockTimestamp < unscheduledActivation:
+			return fmt.Errorf("%w: AP2 must be either unscheduled or initially activated", errInvalidUpgradeTime)
 		}
```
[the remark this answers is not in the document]
alarso16: I agree with @ARR4N - defining the shorter variable makes it way easier to read

plugin/evm/vm.go
```diff
@@ -625,6 +600,42 @@ func (vm *VM) Initialize(
...
+
+	// If Durango is scheduled, schedule the Warp Precompile at the same time.
+	if configExtra.DurangoBlockTimestamp != nil {
```
StephenButtolph, line 623: Should this handle the unscheduled time?

plugin/evm/vm.go
```diff
@@ -625,6 +600,42 @@ func (vm *VM) Initialize(
...
+		})
+	}
+	if err := configExtra.Verify(); err != nil {
+		return nil, fmt.Errorf("invalid chain config: %w", err)
+	}
+
+	// Align all the Ethereum upgrades to the Avalanche upgrades
+	if err := params.SetEthUpgrades(g.Config); err != nil {
+		return nil, fmt.Errorf("setting eth upgrades: %w", err)
+	}
```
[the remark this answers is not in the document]
StephenButtolph: It doesn't matter, but I felt like verifying the `configExtra` made the most sense right after populating the `configExtra` rather than after an unrelated operation.

plugin/evm/vm.go
StephenButtolph: I don't think it matters. This PR is only really needed because of some enforced ordering in the eth config.

params/config_extra_test.go
```diff
@@ -0,0 +1,164 @@
...
+				EIP158Block:         genesisBlock,
+				ByzantiumBlock:      genesisBlock,
+				ConstantinopleBlock: genesisBlock,
+				PetersburgBlock:     genesisBlock,
+				IstanbulBlock:       genesisBlock,
+				MuirGlacierBlock:    genesisBlock,
+				BerlinBlock:         genesisBlock,
+				LondonBlock:         genesisBlock,
+				ShanghaiTime:        genesisTimestamp,
+				CancunTime:          genesisTimestamp,
```
meaghanfitzgerald, line 140: [no action required] just curious, if the cancun upgrade was applied to coreth in etna, then why is cancuntime set to genesistime and not the canonical etnatime? same question for all other block/timestamps in these tests

StephenButtolph: These are for testing - not production. The test configs have upgrades either initially activated (at genesis) or disabled. Configuring custom times is handled separately

ARR4N (approved):

ceyonur (approved):
lgtm, just a nit

params/config_extra.go
```diff
@@ -37,32 +46,48 @@ func SetEthUpgrades(c *ChainConfig) {
...
-		// number of 0 as well.
-		initiallyActive := uint64(upgrade.InitiallyActiveTime.Unix())
-		extra := GetExtra(c)
-		if extra.ApricotPhase2BlockTimestamp != nil && *extra.ApricotPhase2BlockTimestamp <= initiallyActive && c.BerlinBlock == nil {
+		// In testing or local networks, we only support enabling Berlin and
+		// London at the initially active time. This corresponds to an intended
+		// block number of 0.
+		switch ap2 := extra.ApricotPhase2BlockTimestamp; {
+		case ap2 == nil:
+			return nil
```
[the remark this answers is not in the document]
StephenButtolph: @ceyonur do we ever allow optional upgrades? I wouldn't expect that in coreth, but maybe in subnet-evm?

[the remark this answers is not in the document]
ceyonur: I was hoping to merge these (draft in https://github.com/ava-labs/subnet-evm/pull/1562)

PR #956, opened 2025-05-09
Explain duplicated UTXO check

StephenButtolph:
## Why this should be merged

Block verification currently reads imported UTXOs twice. Once in `verify`, and once in `onExtraStateChange`.

I was originally confused why this verification existed in `verify` at all, as it doesn't fully verify the UTXOs (the full verification is performed in `onExtraStateChange`). After looking back at the original introduction of the code to `verify`, the rationale made sense.

This PR improves the documentation in the current code so that we don't have to look back at old PRs to understand why this code is important.

Additionally, this aligns the behavior of `verifyUTXOsPresent` with `verifyPredicates`, so the bootstrapped check is only done once.

## How this works

Just improving comments + making the code a bit more uniform.

## How this was tested

CI

## Need to be documented?

No.

No/

### DIFF
--- a/plugin/evm/block.go
+++ b/plugin/evm/block.go
@@ -340,8 +340,8 @@ func (b *Block) verify(predicateContext *precompileconfig.PredicateContext, writ
 			return err
 		}
 
-		// Verify that all the warp messages are correctly marked as either
-		// valid or invalid.
+		// Verify that all the ICM messages are correctly marked as either valid
+		// or invalid.
 		if err := b.verifyPredicates(predicateContext); err != nil {
 			return fmt.Errorf("failed to verify predicates: %w", err)
 		}
--- a/plugin/evm/block.go
+++ b/plugin/evm/block.go
@@ -324,16 +324,24 @@ func (b *Block) verify(predicateContext *precompileconfig.PredicateContext, writ
 		return fmt.Errorf("syntactic block verification failed: %w", err)
 	}
 
-	// verify UTXOs named in import txs are present in shared memory.
-	if err := b.verifyUTXOsPresent(); err != nil {
-		return err
-	}
-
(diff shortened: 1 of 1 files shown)


plugin/evm/block.go
```diff
@@ -394,20 +402,15 @@ func (b *Block) verifyPredicates(predicateContext *precompileconfig.PredicateCon
...
-// verifyUTXOsPresent returns an error if any of the atomic transactions name UTXOs that
-// are not present in shared memory.
+// verifyUTXOsPresent verifies all atomic UTXOs consumed by the block are
+// present in shared memory.
 func (b *Block) verifyUTXOsPresent() error {
 	blockHash := common.Hash(b.ID())
 	if b.vm.atomicBackend.IsBonus(b.Height(), blockHash) {
 		log.Info("skipping atomic tx verification on bonus block", "block", blockHash)
 		return nil
 	}
```
StephenButtolph, line 412: This code doesn't need to be called during bootstrapping either.

plugin/evm/block.go
```diff
@@ -324,16 +324,24 @@ func (b *Block) verify(predicateContext *precompileconfig.PredicateContext, writ
...
+		//
+		// This does not fully verify that this block can spend these UTXOs.
+		// However, it guarantees that any block that fails the later checks was
+		// built by an incorrect block proposer. This ensures that we only mark
+		// blocks as BAD BLOCKs if they were incorrectly generated.
+		if err := b.verifyUTXOsPresent(); err != nil {
+			return err
+		}
+
+		// Verify that all the warp messages are correctly marked as either
```
michaelkaplan13, line 343: Wen "ICM messages"? 😏

michaelkaplan13 (approved):

qdm12 (approved):

PR #962, opened 2025-05-15
refactor: use the built-in max/min to simplify the code

cuinix:
## Why this should be merged

In Go 1.21, the standard library includes built-in [max/min](https://pkg.go.dev/builtin@go1.21.0#max) function, which can greatly simplify the code.


JonathanOppenheimer:
Hello @cuinix - thank you for your contribution. We have a policy of not modifying upstream code (the code you're modifying is from GoEthereum), so we will not be making changes directly.

PR #963, opened 2025-05-15
Atomic sync pkg

ceyonur:
## Why this should be merged

Moves atomic sync related logic to atomic/sync pkg and provides a customizable/extendible way to support extra state handling in syncers

Split from #885 

## How this works

This pull request introduces significant updates to the atomic trie functionality and synchronization logic in the EVM plugin. It includes new features, enhancements to existing tests, and code refactoring for better modularity and maintainability. Below are the most important changes grouped by theme.

### Atomic Trie and Synchronization Enhancements

* Introduced `atomicLeafHandler` in `plugin/evm/atomic/sync/atomic_leaf_handler.go`, a wrapper for `LeafRequestHandler` to facilitate initialization with an atomic trie database.
* Added `atomicSummaryProvider` in `plugin/evm/atomic/sync/atomic_summary_provider.go` to provide state summaries at specific block heights using the atomic trie.
* Implemented `atomicSyncExtender` in `plugin/evm/atomic/sync/atomic_sync_extender.go`, which manages the synchronization of the atomic trie and applies updates to shared memory.
* Created `atomicSyncSummaryParser` in `plugin/evm/atomic/sync/atomic_sync_parser.go` to parse synchronization summaries from bytes.
* Added `AtomicSyncSummary` in `plugin/evm/atomic/sync/atomic_sync_summary.go`, which encapsulates block and atomic trie state for synchronization purposes.

### Testing Enhancements

* Added a unit test in `plugin/evm/atomic/sync/atomic_sync_summary_test.go` to validate the marshaling and parsing of `AtomicSyncSummary` objects.
* Updated `plugin/evm/atomic/sync/atomic_syncer_test.go` to use the new `atomicLeafHandler` and `atomicSyncSummary` components for testing atomic trie synchronization.

### Refactoring and Code Cleanup

* Renamed `plugin/evm/atomic_syncer.go` and `plugin/evm/atomic_syncer_test.go` to `plugin/evm/atomic/sync/atomic_syncer.go` and `plugin/evm/atomic/sync/atomic_syncer_test.go`, respectively, for better organization. [[1]](diffhunk://#diff-9ed7286d57f7743e525f35f195746fd7af4d4b09eb881f4aaf65f2b904788be4L4-R4) [[2]](diffhunk://#diff-7a6aef107a03f7060ba600e008a9005c26d36f8823f256003042440521e8450dL4-R4)
* Added `AtomicTrieNode` constant and `AtomicTrie` interface to define the structure and operations of the atomic trie.
* Enhanced the `Block` struct in `plugin/evm/block.go` to implement the new `EthBlockWrapper` interface, allowing retrieval of Ethereum blocks for synchronization. [[1]](diffhunk://#diff-9a870b1572ea65697aea0a8f959e0ee20efa4ff3370c3e39dcc2f3ec436b4fc3R36-R45) [[2]](diffhunk://#diff-9a870b1572ea65697aea0a8f959e0ee20efa4ff3370c3e39dcc2f3ec436b4fc3R456-R459)

### Minor Fixes and Adjustments

* Removed descriptive error messages in `plugin/evm/atomic/state/atomic_trie_test.go` to simplify assertions. [[1]](diffhunk://#diff-4093f6807bb07d0bb0e6eea73a446cba30178decd33b818352d615122797b5aeL828-R828) [[2]](diffhunk://#diff-4093f6807bb07d0bb0e6eea73a446cba30178decd33b818352d615122797b5aeL839-R839) [[3]](diffhunk://#diff-4093f6807bb07d0bb0e6eea73a446cba30178decd33b818352d615122797b5aeL864-R864) [[4]](diffhunk://#diff-4093f6807bb07d0bb0e6eea73a446cba30178decd33b818352d615122797b5aeL873-R873)

## How this was tested

Existing UT

## Need to be documented?

No

No


plugin/evm/sync/syncervm_client.go
```diff
@@ -17,69 +18,91 @@ import (
...
+
+// EthBlockWrapper can be implemented by a concrete block wrapper type to
+// return *types.Block, which is needed to update chain pointers at the
+// end of the sync operation.
+type EthBlockWrapper interface {
+	GetEthBlock() *types.Block
+}
+
+// Extender is an interface that allows for extending the state sync process.
+type Extender interface {
```
alarso16, line 48: Eventually, I'm going to change the extension interface to be something like this:
```
type Syncer[T any] interface {
	Start(ctx context.Context, target T) error
	Wait(ctx context.Context) error
	Close() error
	UpdateSyncTarget(ctx context.Context, target T) error
}
```
where `T` is a summary.
I think this design is compatible, but I just wanted to put that here for reference.

alarso16 (commented):
I think this PR would be worth syncing to Fuji or mainnet to ensure everything works as expected.

plugin/evm/atomic/sync/atomic_sync_summary.go
```diff
@@ -0,0 +1,99 @@
...
+	summaryID, err := ids.ToID(crypto.Keccak256(bytes))
+	if err != nil {
+		return nil, fmt.Errorf("failed to compute summary ID: %w", err)
+	}
+	summary.summaryID = summaryID
+
+	return &summary, nil
+}
+
+func (a *AtomicSyncSummary) GetBlockHash() common.Hash {
```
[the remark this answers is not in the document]
ceyonur: Frankly I haven't tried to wrap the  block sync summary here and rather reimplemented whole thing (message.Syncable interface), it might make sense to wrap it.

plugin/evm/message/block_sync_summary.go
```diff
@@ -0,0 +1,119 @@
...
+	var err error
+	Codec, err = NewCodec(BlockSyncSummary{})
+	if err != nil {
+		panic(fmt.Errorf("failed to create codec manager: %w", err))
+	}
+}
+
+// BlockSyncSummary provides the information necessary to sync a node starting
+// at the given block.
+type BlockSyncSummary struct {
```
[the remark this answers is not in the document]
ceyonur: yes I think it makes sense, I was initially concerned with the codec but it should be doable now

plugin/evm/sync/syncervm_client.go
ceyonur: I'm fine with any solution that does not require atomic logic to be moved to evm pkg, we just need to be careful about committing stuff to DB.

plugin/evm/atomic/sync/atomic_sync_summary.go
```diff
@@ -0,0 +1,99 @@
...
+
+// AtomicSyncSummary provides the information necessary to sync a node starting
+// at the given block.
+type AtomicSyncSummary struct {
+	*message.BlockSyncSummary `serialize:"true"`
+	AtomicRoot                common.Hash `serialize:"true"`
+
+	summaryID  ids.ID
+	bytes      []byte
+	acceptImpl message.AcceptImplFn
```
michaelkaplan13, line 43: Is there a reason why we need to duplicate these fields from `message.BlockSyncSummary`? Since we're embedding `*message.BlockSyncSummary` already, it seems like it may be unnecessary.

We can likely also get rid of `GetBlockHash`, `GetBlockRoot`, etc, since they are already present in `BlockSyncSummary` and embedded here, right?

plugin/evm/vm.go
```diff
@@ -674,7 +672,55 @@ func (vm *VM) initializeChain(lastAcceptedHash common.Hash) error {
...
+		Handler:    leafHandler,
+	})
+
+	leafHandlers := make(LeafHandlers, len(leafHandlerConfigs))
+	for _, leafConfig := range leafHandlerConfigs {
+		if _, exists := leafHandlers[leafConfig.LeafType]; exists {
+			return fmt.Errorf("duplicate leaf type %v", leafConfig.LeafType)
+		}
+		leafHandlers[leafConfig.LeafType] = leafConfig.Handler
+	}
```
michaelkaplan13, line 712:
```suggestion
	leafHandlers := make(LeafHandlers, 2) // 2 default leaf handlers for state trie and atomic trie
	syncStats := handlerstats.GetOrRegisterHandlerStats(metrics.Enabled)

	// Register the leafs request handler for state trie
	leafHandlers[message.StateTrieNode] = handlers.NewLeafsRequestHandler(evmTrieDB,
		message.StateTrieKeyLength,
		vm.blockChain, vm.networkCodec,
		syncStats,
	)

	// Register the atomic trie leafs request handler
	leafHandlers[atomicsync.AtomicTrieNode] = atomicsync.NewAtomicLeafHandler(vm.atomicBackend.AtomicTrie().TrieDB(), atomicstate.TrieKeyLength, vm.networkCodec)
```

IIUC, this could simplify this initialization code.

plugin/evm/atomic/state/atomic_trie_test.go
```diff
@@ -825,7 +825,7 @@ func verifyOperations(t testing.TB, atomicTrie *AtomicTrie, codec codec.Manager,
 	fromBytes := make([]byte, wrappers.LongLen)
 	binary.BigEndian.PutUint64(fromBytes, from)
 	iter, err := atomicTrie.Iterator(rootHash, fromBytes)
-	require.NoError(t, err, "creating iterator")
```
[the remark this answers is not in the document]
ceyonur: Felt odd to have a description as we have marked these methods as helper

plugin/evm/atomic/sync/atomic_sync_summary.go
ceyonur: These are unexported fields, the bytes and summaryID here should include `AtomicRoot`, so we cannot use them from blockSyncSummary at best we can use another strut but feels like unnecessary. 

removed the methods

plugin/evm/vm.go
ceyonur: I kept this a bit "ugly" with the intention that we can register leaf configs externally (like in #885), these will be changed at the end.

ceyonur: removed in https://github.com/ava-labs/coreth/pull/985

michaelkaplan13: Sorry I don't quite follow...I may be missing something here.

Why do we create handlers in `leafHandlerConfigs` only to then loop over it to assign into `leafHandlers`? It looks like that pattern still exists in #985 also

plugin/evm/atomic/sync/atomic_sync_parser.go
```diff
@@ -0,0 +1,37 @@
...
+	"fmt"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/coreth/plugin/evm/message"
+	"github.com/ava-labs/libevm/crypto"
+)
+
+var _ message.SyncableParser = (*atomicSyncSummaryParser)(nil)
+
+type atomicSyncSummaryParser struct{}
```
StephenButtolph, line 15:
```suggestion
type summaryParser struct{}
```

plugin/evm/atomic/sync/atomic_sync_parser.go
```diff
@@ -0,0 +1,37 @@
...
+
+func NewAtomicSyncSummaryParser() *atomicSyncSummaryParser {
+	return &atomicSyncSummaryParser{}
+}
+
+func (a *atomicSyncSummaryParser) ParseFromBytes(summaryBytes []byte, acceptImpl message.AcceptImplFn) (message.Syncable, error) {
```
StephenButtolph, line 21: Do we ever parse something _not_ from bytes?
```suggestion
func (a *atomicSyncSummaryParser) Parse(summaryBytes []byte, acceptImpl message.AcceptImplFn) (message.Syncable, error) {
```

plugin/evm/atomic/sync/atomic_sync_parser.go
```diff
@@ -0,0 +1,37 @@
...
+	summary := AtomicSyncSummary{}
+	if codecVersion, err := Codec.Unmarshal(summaryBytes, &summary); err != nil {
+		return nil, fmt.Errorf("failed to parse syncable summary: %w", err)
+	} else if codecVersion != message.Version {
+		return nil, fmt.Errorf("failed to parse syncable summary due to unexpected codec version (got %d, expected %d)", codecVersion, message.Version)
+	}
```
StephenButtolph, line 27: Because we only register a single codec version on this codec, there is no reason to check the version. It is already guaranteed to be `message.Version`
```suggestion
	if _, err := Codec.Unmarshal(summaryBytes, &summary); err != nil {
		return nil, fmt.Errorf("failed to parse syncable summary: %w", err)
	}
```

plugin/evm/atomic/sync/atomic_sync_extender.go
```diff
@@ -0,0 +1,82 @@
...
+		atomicSyncSummary.BlockNumber,
+		a.stateSyncRequestSize,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create atomic syncer: %w", err)
+	}
+	if err := atomicSyncer.Start(ctx); err != nil {
+		return fmt.Errorf("failed to start atomic syncer: %w", err)
+	}
+	err = <-atomicSyncer.Done()
```
StephenButtolph, line 56: It seems like every instance we utilize the syncer we make it synchronous with this blocking pattern. Is there a reason that we expose `Start` as an async interface at all?

plugin/evm/atomic/sync/atomic_sync_summary.go
```diff
@@ -0,0 +1,87 @@
...
+	var err error
+	Codec, err = message.NewCodec(AtomicSyncSummary{})
+	if err != nil {
+		panic(fmt.Errorf("failed to create codec manager: %w", err))
+	}
+}
+
+// AtomicSyncSummary provides the information necessary to sync a node starting
+// at the given block.
+type AtomicSyncSummary struct {
```
StephenButtolph, line 37:
```suggestion
type Summary struct {
```

plugin/evm/atomic/sync/atomic_sync_summary.go
```diff
@@ -0,0 +1,87 @@
...
+var _ message.Syncable = (*AtomicSyncSummary)(nil)
+
+// Codec is the codec manager that contains the codec for AtomicSyncSummary and
+// other message types that are used in the network protocol. This is to ensure that the codec
+// version is consistent across all message types and includes the codec for AtomicSyncSummary.
+var Codec codec.Manager
+
+func init() {
+	var err error
+	Codec, err = message.NewCodec(AtomicSyncSummary{})
```
StephenButtolph, line 29: Why do we register `AtomicSyncSummary` here? We never seem to marshal/unmarshal these as interfaces.

plugin/evm/message/block_sync_summary.go
```diff
@@ -0,0 +1,95 @@
...
+var _ Syncable = (*BlockSyncSummary)(nil)
+
+// Codec is the codec manager that contains the codec for BlockSyncSummary and
+// other message types that are used in the network protocol. This is to ensure that the codec
+// version is consistent across all message types and includes the codec for BlockSyncSummary.
+var Codec codec.Manager
+
+func init() {
+	var err error
+	Codec, err = NewCodec(BlockSyncSummary{})
```
StephenButtolph, line 27: Similar comment as before, I don't see where we ever use this to marshal/unmarshal interfaces. Why do we register types for this codec?

plugin/evm/atomic/sync/atomic_sync_summary.go
```diff
@@ -0,0 +1,87 @@
...
+type AtomicSyncSummary struct {
+	*message.BlockSyncSummary `serialize:"true"`
+	AtomicRoot                common.Hash `serialize:"true"`
+
+	summaryID  ids.ID
+	bytes      []byte
+	acceptImpl message.AcceptImplFn
+}
+
+func NewAtomicSyncSummary(blockHash common.Hash, blockNumber uint64, blockRoot common.Hash, atomicRoot common.Hash) (*AtomicSyncSummary, error) {
```
StephenButtolph, line 46: I understand that we only generate summaries this way when we are serving them to peers... But it feels unusual to leave `acceptImpl` undefined here and for it to only be set out-of-band in `Parse`.

Perhaps we should move the definition of `Parse` to live here as well?

plugin/evm/message/block_sync_summary_parser.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/ava-labs/libevm/crypto"
+)
+
+type BlockSyncSummaryParser struct{}
+
+func NewBlockSyncSummaryParser() *BlockSyncSummaryParser {
+	return &BlockSyncSummaryParser{}
+}
+
+func (b *BlockSyncSummaryParser) ParseFromBytes(summaryBytes []byte, acceptImpl AcceptImplFn) (Syncable, error) {
```
StephenButtolph, line 18: Same comment as above
```suggestion
func (b *BlockSyncSummaryParser) Parse(summaryBytes []byte, acceptImpl AcceptImplFn) (Syncable, error) {
```

plugin/evm/message/block_sync_summary_parser.go
```diff
@@ -0,0 +1,34 @@
...
+	summary := BlockSyncSummary{}
+	if codecVersion, err := Codec.Unmarshal(summaryBytes, &summary); err != nil {
+		return nil, fmt.Errorf("failed to parse syncable summary: %w", err)
+	} else if codecVersion != Version {
+		return nil, fmt.Errorf("failed to parse syncable summary due to unexpected codec version (%d != %d)", codecVersion, Version)
+	}
```
StephenButtolph, line 24: Same as above
```suggestion
	if _, err := Codec.Unmarshal(summaryBytes, &summary); err != nil {
		return nil, fmt.Errorf("failed to parse syncable summary: %w", err)
	}
```

plugin/evm/sync/syncervm_client.go
```diff
@@ -1,13 +1,14 @@
 // Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
-
-package evm
+package sync
```
StephenButtolph, line 3: Don't we normally have a space between the copyright header and the package?

plugin/evm/sync/syncervm_client.go
```diff
@@ -17,69 +18,91 @@ import (
...
+type BlockAcceptor interface {
+	PutLastAcceptedID(ids.ID) error
+}
+
+// EthBlockWrapper can be implemented by a concrete block wrapper type to
+// return *types.Block, which is needed to update chain pointers at the
+// end of the sync operation.
+type EthBlockWrapper interface {
+	GetEthBlock() *types.Block
+}
```
StephenButtolph, line 45: This type is duplicated between here and `plugin/evm/block.go` - do we need both of these definitions?

StephenButtolph (commented):
Still reviewing

plugin/evm/atomic/sync/atomic_sync_extender.go
ceyonur: It looks like we don't, I'm not sure if we had any reason and it got obsolote. However it might come in handy for dynamic state sync that @alarso16 is preparing.

plugin/evm/atomic/sync/atomic_sync_summary.go
ceyonur: You mean why do we take a interface type in `NewCodec` or why do we register `AtomicSyncSummary{}` rather than `&AtomicSyncSummary{}`?

plugin/evm/message/block_sync_summary.go
ceyonur: This was a step to reduce diff between subnet-evm. In Subnet-EVM we only use `BlockSyncSummary`. We can argue that they both should have their own codec, but this seemed can be shared.

plugin/evm/atomic/sync/atomic_sync_summary.go
ceyonur: Parse is defined for `SummaryParser` which is a different type. Added comment

plugin/evm/sync/syncervm_client.go
ceyonur: removed the one in block

plugin/evm/atomic/sync/atomic_leaf_handler.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/ava-labs/libevm/triedb"
+)
+
+// atomicLeafHandler is a wrapper around handlers.LeafRequestHandler that allows for initialization after creation
+type atomicLeafHandler struct {
+	handlers.LeafRequestHandler
+}
+
+// NewAtomicLeafHandler initializes the atomicLeafHandler with the provided atomicTrieDB, trieKeyLength, and networkCodec
+func NewAtomicLeafHandler(atomicTrieDB *triedb.Database, trieKeyLength int, networkCodec codec.Manager) *atomicLeafHandler {
```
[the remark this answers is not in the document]
ceyonur: My intention was to rename them at the end, but it's probably just easier to rename them here. 

There are some variable names that can be confusing since it's a generic type of `triedb.Database` or `common.Hash` so I prefer to keep "atomic" name in them.

plugin/evm/atomic/sync/atomic_sync_extender.go
```diff
@@ -0,0 +1,82 @@
...
+		atomicTrie:           atomicTrie,
+		stateSyncRequestSize: stateSyncRequestSize,
+	}
+}
+
+func (a *atomicSyncExtender) Sync(ctx context.Context, client syncclient.LeafClient, verDB *versiondb.Database, syncSummary message.Syncable) error {
+	atomicSyncSummary, ok := syncSummary.(*AtomicSyncSummary)
+	if !ok {
+		return fmt.Errorf("expected *AtomicSyncSummary, got %T", syncSummary)
+	}
```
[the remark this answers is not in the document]
ceyonur: Not sure what to do here other than expect another interface that provides `AtomicRoot()`, `BlockNumber()` etc.

plugin/evm/atomic/sync/atomic_sync_extender.go
alarso16: I have a refactor of the interface waiting here #927, so I don't think there's much sense changing it right now.

plugin/evm/sync/syncervm_client.go
```diff
@@ -362,50 +370,54 @@ func (client *stateSyncerClient) finishSync() error {
...
-func (client *stateSyncerClient) updateVMMarkers() error {
+func (client *stateSyncerClient) commitVMMarkers() error {
 	// Mark the previously last accepted block for the shared memory cursor, so that we will execute shared
 	// memory operations from the previously last accepted block to [vm.syncSummary] when ApplyToSharedMemory
 	// is called.
-	if err := client.atomicBackend.MarkApplyToSharedMemoryCursor(client.lastAcceptedHeight); err != nil {
+	id, err := ids.ToID(client.syncSummary.GetBlockHash().Bytes())
+	if err != nil {
 		return err
 	}
```
StephenButtolph, line 413:
```suggestion
	id := ids.ID(client.syncSummary.GetBlockHash())
```

plugin/evm/sync/syncervm_server.go
```diff
@@ -11,65 +10,35 @@ import (
...
-	AtomicTrie *state.AtomicTrie
+var errProviderNotSet = fmt.Errorf("provider not set")
 
-	// SyncableInterval is the interval at which blocks are eligible to provide syncable block summaries.
-	SyncableInterval uint64
+type SummaryProvider interface {
+	StateSummaryAtBlock(ethBlock *types.Block) (block.StateSummary, error)
 }
 
 type stateSyncServer struct {
```
StephenButtolph, line 24:
```suggestion
type server struct {
```

plugin/evm/sync/syncervm_server.go
```diff
@@ -11,65 +10,35 @@ import (
...
 }
 
-type StateSyncServer interface {
+type Server interface {
 	GetLastStateSummary(context.Context) (block.StateSummary, error)
 	GetStateSummary(context.Context, uint64) (block.StateSummary, error)
 }
 
-func NewStateSyncServer(config *stateSyncServerConfig) StateSyncServer {
+func SyncServer(chain *core.BlockChain, provider SummaryProvider, syncableInterval uint64) Server {
```
StephenButtolph, line 36:
```suggestion
func NewServer(chain *core.BlockChain, provider SummaryProvider, syncableInterval uint64) Server {
```

plugin/evm/atomic/sync/atomic_sync_summary.go
StephenButtolph: why do we take a interface type in NewCodec?

ceyonur: https://github.com/ava-labs/coreth/pull/963#discussion_r2112789420

PR #968, opened 2025-05-17
Add linters

ceyonur:


peer/client.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 var (
-	_ NetworkClient = &client{}
+	_ NetworkClient = (*client)(nil)
```
alarso16, line 16: Why did we change of all these? Is this better practice?

plugin/evm/atomic_trie_test.go
```diff
@@ -726,13 +726,11 @@ func TestAtomicTrie_AcceptTrie(t *testing.T) {
 				encoder.WriteBytes(make([]byte, 32)) // value
 				encoder.ListEnd(offset)
 				testBlob := encoder.ToBytes()
-				err := encoder.Flush()
-				require.NoError(t, err)
+				require.NoError(t, encoder.Flush())
```
alarso16, line 729: nit: I think it's more readable before

plugin/evm/message/block_request.go
```diff
@@ -12,9 +12,7 @@ import (
 	"github.com/ava-labs/libevm/common"
 )
 
-var (
-	_ Request = BlockRequest{}
-)
+var _ Request = BlockRequest{}
```
alarso16, line 15: Do you need to update this one too?
`var_Request = (*BlockRequest)(nil)`

plugin/evm/vm_test.go
```diff
@@ -211,7 +212,7 @@ func newVM(t *testing.T, config testVMConfig) *testVM {
 		CantSendAppGossip: true,
 		SendAppGossipF:    func(context.Context, commonEng.SendConfig, []byte) error { return nil },
 	}
-	err := vm.Initialize(
+	require.NoError(t, vm.Initialize(
```
alarso16, line 215: Which linter is recommending this?

peer/network_test.go
```diff
@@ -41,18 +41,18 @@ var (
...
+	_                 = (*GreetingResponse)(nil)
+	_                 = (*TestMessage)(nil)
 
-	_ message.RequestHandler = &HelloGreetingRequestHandler{}
-	_ message.RequestHandler = &testRequestHandler{}
+	_ message.RequestHandler = (*HelloGreetingRequestHandler)(nil)
+	_ message.RequestHandler = (*testRequestHandler)(nil)
 
-	_ common.AppSender = testAppSender{}
+	_ common.AppSender = (*testAppSender)(nil)
```
alarso16, line 53: This is technically different, right? I don't think it matters, but this diff specific line of diff is probably unnecessary.

sync/client/test_network.go
```diff
@@ -13,44 +13,44 @@ import (
 	"github.com/ava-labs/avalanchego/version"
 )
 
-var _ peer.NetworkClient = &mockNetwork{}
+var _ peer.NetworkClient = (*testNetwork)(nil)
 
-// TODO replace with gomock library
-type mockNetwork struct {
+// TODO replace with gotest library
```
alarso16, line 18: Is this a find and replace error? It should probably still be `gomock`, right?

alarso16: Actually, it looks like `gomock` isn't maintained. Maybe Uber's fork is still good, if you wanted to update the comment to that. As a side note, I think avalanchego's linter enforces using Uber's testing suites, which might be good in coreth too.

sync/statesync/sync_test.go
```diff
@@ -532,3 +533,91 @@ func testSyncerSyncsToNewRoot(t *testing.T, deleteBetweenSyncs func(*testing.T,
 		deleteBetweenSyncs(t, root1, clientDB)
 	})
 }
+
+// assertDBConsistency checks [serverTrieDB] and [clientTrieDB] have the same EVM state trie at [root],
+// and that [clientTrieDB.DiskDB] has corresponding account & snapshot values.
+// Also verifies any code referenced by the EVM state is present in [clientTrieDB] and the hash is correct.
+func assertDBConsistency(t testing.TB, root common.Hash, clientDB ethdb.Database, serverTrieDB, clientTrieDB *triedb.Database) {
```
alarso16, line 540: this function might go well in `statesynctest`

sync/statesync/sync_test.go
```diff
@@ -532,3 +533,91 @@ func testSyncerSyncsToNewRoot(t *testing.T, deleteBetweenSyncs func(*testing.T,
...
+
+		assert.Equal(t, storageTrieLeavesCount, snapshotStorageKeysCount)
+		return nil
+	})
+
+	// Check that the number of accounts in the snapshot matches the number of leaves in the accounts trie
+	assert.Equal(t, trieAccountLeaves, numSnapshotAccounts)
+}
+
+func fillAccountsWithStorage(t *testing.T, serverDB ethdb.Database, serverTrieDB *triedb.Database, root common.Hash, numAccounts int) common.Hash {
```
alarso16, line 605: This one could go in `statesynctest` too

alarso16 (commented):
My grep skills are really bad, I can't give a real approval for this

peer/client.go
ceyonur: yes https://go.dev/doc/effective_go#blank_implements

plugin/evm/atomic_trie_test.go
ceyonur: it's enforced by the linter

plugin/evm/message/block_request.go
ceyonur: yes good catch

plugin/evm/vm_test.go
ceyonur: https://github.com/ava-labs/coreth/blob/add-linters/scripts/lint.sh#L149-L156

sync/client/test_network.go
ceyonur: Please open a PR if you know how to add that and handle mocks. This PR is kind-of blocking others and already a huge one.

peer/network_test.go
ceyonur: not sure, but it looks like it should be changed in this way.

sync/statesync/sync_test.go
ceyonur: we use it only in this file

sync/statesync/sync_test.go
ceyonur: again we use it only here

sync/client/test_network.go
alarso16: I'll create an issue, and fix it soon

PR #960, opened 2025-05-14
move atomic state pkg

ceyonur:
## Why this should be merged

This PR moves atomic trie, state, repos and backend to atomic/state pkg

Split from https://github.com/ava-labs/coreth/pull/885

## How this works

This pull request introduces significant changes to the atomic backend and repository in the EVM plugin, focusing on refactoring, renaming, and adding utility functions to enhance maintainability and functionality. Key changes include the introduction of a new `SharedMemories` utility for testing, renaming and restructuring files and packages, and refactoring the atomic backend and repository for improved modularity and clarity.

### New Utilities for Testing:
* Added a `SharedMemories` struct in `plugin/evm/atomic/atomictest/shared_memories.go` to facilitate testing of shared memory operations. This includes helper methods such as `AddItemsToBeRemovedToPeerChain`, `AssertOpsApplied`, and `AssertOpsNotApplied`.
* Introduced `ConvertToAtomicOps` in `plugin/evm/atomic/atomictest/utils.go` to convert atomic transactions into a map of atomic operations for testing purposes.

### File and Package Restructuring:
* Renamed `plugin/evm/atomic_backend.go` to `plugin/evm/atomic/state/atomic_backend.go` and updated the package name to `state`. This reflects a more modular organization of atomic-related functionality.
* Renamed `plugin/evm/atomic_tx_repository.go` to `plugin/evm/atomic/state/atomic_repository.go` and updated the package name to `state`.

### Refactoring of Atomic Backend:
* Refactored the `atomicBackend` struct into `AtomicBackend` with updated field types and improved dependency injection, such as replacing direct database references with the `AtomicRepository`. [[1]](diffhunk://#diff-24781baefb5ddc9629bcf492b185273907de2e4e75236da9b262f482e1bb628cL24-R63) [[2]](diffhunk://#diff-24781baefb5ddc9629bcf492b185273907de2e4e75236da9b262f482e1bb628cL124-R82)
* Simplified methods in `AtomicBackend` to use the repository's database abstractions, improving modularity and reducing direct database interactions. [[1]](diffhunk://#diff-24781baefb5ddc9629bcf492b185273907de2e4e75236da9b262f482e1bb628cL178-R136) [[2]](diffhunk://#diff-24781baefb5ddc9629bcf492b185273907de2e4e75236da9b262f482e1bb628cL227-R186) [[3]](diffhunk://#diff-24781baefb5ddc9629bcf492b185273907de2e4e75236da9b262f482e1bb628cL300-R261)
[...]

PR #969, opened 2025-05-17
Add license linter

ceyonur:
## Why this should be merged

adds linter for licenses 

There are 2 licenses: one for files we created + upstream files we modified (or kept same). The script can handle the given folder and file paths within `upstream_licensed_folders`

## How this works

Added lint.sh (just the license part) from avalanchego

PR Updates copyright headers across multiple files in the `accounts/abi` package to reflect the new copyright range (2019-2025) and includes a reference to the `LICENSE` file for licensing terms.

## How this was tested

added Linter to CI

## Need to be documented?

No

No


qdm12:
See this comment: https://github.com/ava-labs/coreth/pull/960#discussion_r2095049437

I think the libevm header CI approach is better, but feel free to contradict!

ceyonur:
> See this comment: [#960 (comment)](https://github.com/ava-labs/coreth/pull/960#discussion_r2095049437)
> 
> I think the libevm header CI approach is better, but feel free to contradict me!

It certainly can replace the script in avalanchego as it uses only one license, but this one (coreth) requires a bit more complex handling as ((IMO) we need to handle the upstream license as well

alarso16 (approved):

PR #971, opened 2025-05-19
replace broken moby link  reexec.go

dizer-ti:
## Why this should be merged
Hey team—noticed a dead link, replaced it with a working URL

https://github.com/moby/moby/blob/master/pkg/reexec/reexec.go - old link
https://github.com/moby/moby/blob/master/pkg/reexec/reexec_deprecated.go - new link

### DIFF
--- a/internal/reexec/reexec.go
+++ b/internal/reexec/reexec.go
@@ -10,7 +10,7 @@
 // Much love to the original authors for their work.
 // **********
 // This file originates from Docker/Moby,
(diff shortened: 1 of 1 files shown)


alarso16:
Thanks for making the edit. However, this code is adapted from [`ethereum/go-ethereum`](https://github.com/ethereum/go-ethereum/blob/33c5031a2e6888dbac3dba0628c786442ceabc33/internal/reexec/reexec.go#L2), and this PR should be suggested there.

PR #974, opened 2025-05-21
chore(acp176): change MinTargetPerSecond to be a variable

qdm12:
## Why this should be merged

- Needed for load testing on the avalanchego side

## How this works

- Change constant to be a variable
- Add comment with what variables to update upon update of this variable
- Update associated constants to be variables

### DIFF
--- a/plugin/evm/upgrade/acp176/acp176.go
+++ b/plugin/evm/upgrade/acp176/acp176.go
@@ -19,8 +19,20 @@ import (
 	"github.com/holiman/uint256"
 )
 
+var (
+	// MinTargetPerSecond is a variable so that it can be changed by external Go code, notably
+	// for load testing in the avalanchego repository. You might want to also update the
+	// following variables value since they depend on MinMaxPerSecond:
+	// - MinTargetPerSecond
+	// - MinMaxPerSecond
+	// - MinMaxCapacity
+	// - eth/gasprice.DefaultMinGasUsed
+	MinTargetPerSecond = gas.Price(1_000_000) // P
+	MinMaxPerSecond    = MinTargetPerSecond * TargetToMax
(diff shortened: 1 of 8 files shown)


plugin/evm/upgrade/acp176/acp176.go
```diff
@@ -19,8 +19,20 @@ import (
...
 
+var (
+	// MinTargetPerSecond is a variable so that it can be changed by external Go code, notably
+	// for load testing in the avalanchego repository. You might want to also update the
+	// following variables value since they depend on MinMaxPerSecond:
+	// - MinTargetPerSecond
+	// - MinMaxPerSecond
+	// - MinMaxCapacity
+	// - eth/gasprice.DefaultMinGasUsed
+	MinTargetPerSecond = gas.Price(1_000_000) // P
```
ceyonur, line 30: I believe the intention is to increase this, right? Can't we accomplish the same thing by increasing the gas target in coreth config?

ceyonur: maybe not directly this variable, but we can increase the target gas per second ultimately

qdm12: Having a look, but so far it didn't seem to work. Changing this to be a draft for now

qdm12:
Note if this shall go forward, due to the cycle dependency between coreth and avalanchego, you would need to:

1. Create an avalanchego branch, use this coreth branch for the coreth dependency, and fix code compilation errors
2. Update the avalanchego dependency in this PR's branch to use that avalanchego branch created in step 1.
3. Now CI should pass. Review and merge this PR.
4. (Eventually) update avalanchego to use the coreth's master branch commit

PR #982, opened 2025-05-23
fix AvalancheGo version and nil configs for Granite

ceyonur:
## Why this should be merged

Fixes e2e tests that failing because the dependent commit does not exist anymore on avalanchego repo.

## How this works

bumps avalanchego to commit in https://github.com/ava-labs/avalanchego/pull/3977

## How this was tested

e2e

## Need to be documented?

no

no

### DIFF
--- a/plugin/evm/vm_test.go
+++ b/plugin/evm/vm_test.go
@@ -123,6 +123,8 @@ var (
 		upgradetest.Fortuna:           params.TestFortunaChainConfig,
 	}
 
+	latestKnownFork = upgradetest.Fortuna
+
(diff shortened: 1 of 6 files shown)


michaelkaplan13 (approved):

plugin/evm/vm_test.go
```diff
@@ -123,6 +123,8 @@ var (
 		upgradetest.Fortuna:           params.TestFortunaChainConfig,
 	}
 
+	latestKnownFork = upgradetest.Fortuna
```
StephenButtolph, line 126: Why don't we just add `Granite` to the chain params?

PR #987, opened 2025-05-27
Add granite configs

StephenButtolph:
## Why this should be merged

Aligns the avalanchego upgrades struct with the coreth upgrades structs.

## How this works

Boilerplate.

## How this was tested

CI

## Need to be documented?

No

No


michaelkaplan13 (approved):
Makes sense to me. Tagging @ceyonur for future reference since I know he had just added the `latestKnownFork`.

PR #988, opened 2025-05-27
Unexport transaction visitor

StephenButtolph:
## Why this should be merged

In avalanchego, we have started unexporting visitors because people found the usage of a normal function more readable than passing in visitors.

It might be less important here, but as visitors get more complex (including returning additional values other than just errors) this is more scalable.

## How this works

Unexports `atomic.SemanticVerifier`.

## How this was tested

CI

## Need to be documented?

No

No


StephenButtolph:
🤔

StephenButtolph:
No clue why things are failing, but will wait to address this

PR #995, opened 2025-05-30
Require min available capacity to build a block

aaronbuchwald:
## Why this should be merged

This PR updates the block building logic to return an error if the available gas capacity would be < min(80% of GasLimit, 15m).

This ensures that when the chain is quickly increasing fees due to all gas being consumed, block building will pause to allow the capacity to refill until it can guarantee that there is space for a transaction with a gas limit of at least 80% of the gas limit (currently configured on mainnet to 16m implying transactions of size 12.8M gas can be included).

The original heuristic was to ONLY check that at least 80% of the gas limit is available, but as the gas target scales up this would trade off delaying block production to guarantee greater than 15m gas is available. This check is added to ensure that transactions that could be included prior to ACP-176 can properly bid to be included, so there's no need to scale up as the gas limit increases further.

## How this works

After calculating the gas capacity, return an error if there's insufficient capacity. Block building will retry when capacity has refilled.

## How this was tested

Adds two unit tests:
- test block building fails with the expected error when there's insufficient capacity
- test when there's a large and small transaction to include, block building fails rather than include the small tx and starve the larger tx

## Need to be documented?

TODO: add a reference to https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -274,6 +274,15 @@ func (w *worker) createCurrentEnvironment(predicateContext *precompileconfig.Pre
 	// If Fortuna has activated, enforce we only build a block when there is a minimum
 	// built up gas capacity.
 	if chainConfigExtra.IsFortuna(header.Time) {
+		// ACP-176 maintains a gas capacity bucket that fills up over time separate from the gas limit.
+		// Since smaller transactions can consume the available gas capacity every second, this can lead
+		// to transactions above this size stalling.
+		// To mitigate this, the block builder returns an error and waits until the gas capacity bucket
+		// has refilled to the desired minimum.
+		// We set that minimum to MaxCapacity - GasCapacityPerSecond to be available before building a block.
+		// This avoids waiting past the point where the gas capacity bucket is already full. If we waited
+		// until the MaxCapacity, then waiting beyond that point would "overflow" the bucket. Since we do
+		// not allow the bucket to overflow, this would waste potential gas capacity.
 		capacityPerSecond := header.GasLimit / acp176.TimeToFillCapacity
 		minimumBuildableCapacity := header.GasLimit - capacityPerSecond
 		if capacity < minimumBuildableCapacity {
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -60,8 +60,7 @@ import (
 const (
 	// Leaves 256 KBs for other sections of the block (limit is 2MB).
 	// This should suffice for atomic txs, proposervm header, and serialization overhead.
-	targetTxsSize  = 1792 * units.KiB
-	minGasCapacity = 12_800_000 // 12.8M gas is the minimum gas capacity to attempt block building
+	targetTxsSize = 1792 * units.KiB
 )
 
 var ErrInsufficientGasCapacityToBuild = errors.New("insufficient gas capacity to build block")
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -59,8 +59,11 @@ import (
 const (
 	// Leaves 256 KBs for other sections of the block (limit is 2MB).
 	// This should suffice for atomic txs, proposervm header, and serialization overhead.
-	targetTxsSize = 1792 * units.KiB
+	targetTxsSize  = 1792 * units.KiB
+	minGasCapacity = 12_800_000 // 12.8M gas is the minimum gas capacity to attempt block building
 )
+
+var errInsufficientGasCapacity = errors.New("insufficient gas capacity to build block")
 
 // environment is the worker's current environment and holds all of the current state information.
 type environment struct {
@@ -267,6 +270,9 @@ func (w *worker) createCurrentEnvironment(predicateContext *precompileconfig.Pre
 	capacity, err := customheader.GasCapacity(chainConfigExtra, parent, header.Time)
 	if err != nil {
 		return nil, fmt.Errorf("calculating gas capacity: %w", err)
+	}
+	if capacity < minGasCapacity {
+		return nil, fmt.Errorf("%w: capacity (%d) < min capacity (%d)", errInsufficientGasCapacity, capacity, minGasCapacity)
 	}
 	numPrefetchers := w.chain.CacheConfig().TriePrefetcherParallelism
 	currentState.StartPrefetcher("miner", state.WithConcurrentWorkers(numPrefetchers))
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -271,8 +271,12 @@ func (w *worker) createCurrentEnvironment(predicateContext *precompileconfig.Pre
 	if err != nil {
 		return nil, fmt.Errorf("calculating gas capacity: %w", err)
 	}
-	if capacity < minGasCapacity {
-		return nil, fmt.Errorf("%w: capacity (%d) < min capacity (%d)", errInsufficientGasCapacity, capacity, minGasCapacity)
+	// If Fortuna has activated, enforce we only build a block when there is a minimum
+	// built up gas capacity.
+	if chainConfigExtra.IsFortuna(header.Time) {
+		if capacity < minGasCapacity {
+			return nil, fmt.Errorf("%w: capacity (%d) < min capacity (%d)", errInsufficientGasCapacity, capacity, minGasCapacity)
+		}
 	}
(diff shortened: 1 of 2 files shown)


miner/worker.go
```diff
@@ -59,9 +60,12 @@ import (
 const (
 	// Leaves 256 KBs for other sections of the block (limit is 2MB).
 	// This should suffice for atomic txs, proposervm header, and serialization overhead.
-	targetTxsSize = 1792 * units.KiB
+	targetTxsSize  = 1792 * units.KiB
+	minGasCapacity = 12_800_000 // 12.8M gas is the minimum gas capacity to attempt block building
```
StephenButtolph, line 64:
```suggestion
	targetTxsSize = 1792 * units.KiB
```

StephenButtolph (commented):
After this, `worker.go` lgtm

plugin/evm/vm_test.go
```diff
@@ -3881,3 +3883,83 @@ func TestNoBlobsAllowed(t *testing.T) {
...
+	require.NoError(err)
+
+	require.NoError(blk1.Verify(ctx))
+	require.NoError(tvm.vm.SetPreference(ctx, blk1.ID()))
+	require.NoError(blk1.Accept(ctx))
+
+	newHead := <-newTxPoolHeadChan
+	if newHead.Head.Hash() != common.Hash(blk1.ID()) {
+		t.Fatalf("Expected new block to match")
+	}
```
StephenButtolph, line 3923:
```suggestion
	require.Equal(common.Hash(blk1.ID()), newHead.Head.Hash())
```

plugin/evm/vm_test.go
```diff
@@ -3881,3 +3883,83 @@ func TestNoBlobsAllowed(t *testing.T) {
...
+	require.Len(errs, 1)
+	require.NoError(errs[0])
+
+	<-tvm.toEngine
+	// Expect block building to fail due to insufficient gas capacity
+	_, err = tvm.vm.BuildBlock(ctx)
+	require.ErrorIs(err, miner.ErrInsufficientGasCapacityToBuild)
+
+	// Wait to fill block capacity and retry block builiding
+	time.Sleep(acp176.TimeToFillCapacity * time.Second)
```
StephenButtolph, line 3957: Not relevant for this PR, but it might make sense to rename `TimeToFillCapacity` to `SecondsToFillCapacity`

plugin/evm/vm_test.go
```diff
@@ -3881,3 +3883,83 @@ func TestNoBlobsAllowed(t *testing.T) {
...
+	require.NoError(tvm.vm.mempool.AddLocalTx(importTx))
+
+	<-tvm.toEngine
+
+	blk1, err := tvm.vm.BuildBlock(ctx)
+	require.NoError(err)
+
+	require.NoError(blk1.Verify(ctx))
+	require.NoError(tvm.vm.SetPreference(ctx, blk1.ID()))
+	require.NoError(blk1.Accept(ctx))
```
StephenButtolph, line 3918: What is the purpose of this first block? Couldn't we just have this block exhaust all of the gas?

plugin/evm/vm_test.go
```diff
@@ -3881,3 +3883,180 @@ func TestNoBlobsAllowed(t *testing.T) {
...
+
+	newHead := <-newTxPoolHeadChan
+	if newHead.Head.Hash() != common.Hash(blk1.ID()) {
+		t.Fatalf("Expected new block to match")
+	}
+	gasAmount := uint64(9_000_000) // 9M gas
```
aaronbuchwald, line 3924: change to 8m

plugin/evm/vm_test.go
StephenButtolph, line 3957: Do we not have a normal way of mocking the clocks in our unit tests of coreth?

plugin/evm/vm_test.go
```diff
@@ -3881,3 +3883,180 @@ func TestNoBlobsAllowed(t *testing.T) {
...
+
+	<-tvm.toEngine
+	blk3, err := tvm.vm.BuildBlock(ctx)
+	require.NoError(err)
+
+	require.NoError(blk3.Verify(ctx))
+	require.NoError(blk3.Accept(ctx))
+}
+
+func TestLargeTxStarvation(t *testing.T) {
```
aaronbuchwald, line 3967: improve name

miner/worker.go
```diff
@@ -268,6 +271,15 @@ func (w *worker) createCurrentEnvironment(predicateContext *precompileconfig.Pre
 	if err != nil {
 		return nil, fmt.Errorf("calculating gas capacity: %w", err)
 	}
+	// If Fortuna has activated, enforce we only build a block when there is a minimum
+	// built up gas capacity.
+	if chainConfigExtra.IsFortuna(header.Time) {
+		capacityPerSecond := header.GasLimit / acp176.TimeToFillCapacity
```
ceyonur, line 277: can we have some comments why we dont use a constant here

aaronbuchwald: Good point - we only really want to ensure that large transactions up to existing expectations are able to get in. This means that there's little to gain from ever going above 15M even if the target increases in the future.

aaronbuchwald: Added a comment explaining this logic https://github.com/ava-labs/coreth/pull/995/commits/eed1b56078cccae791113b5d9200c89faccb9fe5

plugin/evm/vm_test.go
```diff
@@ -3881,3 +3883,183 @@ func TestNoBlobsAllowed(t *testing.T) {
...
+
+	blk1, err := tvm.vm.BuildBlock(ctx)
+	require.NoError(err)
+
+	require.NoError(blk1.Verify(ctx))
+	require.NoError(tvm.vm.SetPreference(ctx, blk1.ID()))
+	require.NoError(blk1.Accept(ctx))
+
+	newHead := <-newTxPoolHeadChan
+	require.Equal(newHead.Head.Hash(), common.Hash(blk1.ID()))
```
StephenButtolph, line 3921:
```suggestion
	require.Equal(common.Hash(blk1.ID()), newHead.Head.Hash())
```

plugin/evm/vm_test.go
```diff
@@ -3881,3 +3883,183 @@ func TestNoBlobsAllowed(t *testing.T) {
...
+			i,
+			big.NewInt(0),
+			8_000_000, // 8M gas is max guaranteed capacity with default gas target of 10M
+			big.NewInt(ap0.MinGasPrice),
+			[]byte{0xfe}, // invalid opcode consumes all gas
+		)
+		signedTx, err := types.SignTx(tx, types.NewEIP155Signer(tvm.vm.chainID), testKeys[0].ToECDSA())
+		require.NoError(err)
+		txs = append(txs, signedTx)
+	}
```
StephenButtolph, line 3936:
```suggestion
	txs := make([]*types.Transaction, 2)
	for i := range txs {
		tx := types.NewContractCreation(
			uint64(i),
			big.NewInt(0),
			8_000_000, // 8M gas is max guaranteed capacity with default gas target of 10M
			big.NewInt(ap0.MinGasPrice),
			[]byte{0xfe}, // invalid opcode consumes all gas
		)
		txs[i], err = types.SignTx(tx, types.NewEIP155Signer(tvm.vm.chainID), testKeys[0].ToECDSA())
		require.NoError(err)
	}
```

plugin/evm/vm_test.go
```diff
@@ -3881,3 +3883,183 @@ func TestNoBlobsAllowed(t *testing.T) {
...
+	require.NoError(blk1.Accept(ctx))
+
+	newHead := <-newTxPoolHeadChan
+	if newHead.Head.Hash() != common.Hash(blk1.ID()) {
+		t.Fatalf("Expected new block to match")
+	}
+	gasAmount := uint64(8_000_000) // 8M gas
+
+	// Build a block consuming all of the available gas
+	maxSizeTxs := make([]*types.Transaction, 0, 2)
```
StephenButtolph, line 4014: Same suggestion around simplifying the initialization of the txs slice

StephenButtolph (approved):
tests and code lgtm, only comments are nit style things

michaelkaplan13 (approved):

PR #1005, opened 2025-06-05
Refactor AP1 `ExtData` Checks

StephenButtolph:
## Why this should be merged

Personally, I find the current code hard to reason about.

- The current code has multiple cases where AP1 is checked.
- The pre-AP1 check (which is required even after the chains have advanced past AP1) is only enforced for Fuji and Mainnet, even though all chains really would need this check.
- The pre-AP1 checks have a control flow based on untrusted data (rather than having the control flow based on trusted data).

## How this works

1. Refactors the pre-AP1 checks to perform validation based on the known `extDataHashes` rather than based on the claimed `extData`.
2. Enforces all non-Fuji/Mainnet chains with pre-AP1 blocks to not have any `extData`.
3. Combined both `!rules.IsApricotPhase1` cases into a single block.

## How this was tested

CI / visual inspection.

## Need to be documented?

No.

No

### DIFF
--- a/plugin/evm/atomic_block_extension.go
+++ b/plugin/evm/atomic_block_extension.go
@@ -91,12 +91,7 @@ func (be *blockExtension) SyntacticVerify(rules extras.Rules) error {
 	headerExtra := customtypes.GetHeaderExtra(ethHeader)
 
 	extData := customtypes.BlockExtData(ethBlock)
-	if rules.IsApricotPhase1 {
-		extDataHash := customtypes.CalcExtDataHash(extData)
-		if headerExtra.ExtDataHash != extDataHash {
-			return fmt.Errorf("extra data hash mismatch: have %x, want %x", headerExtra.ExtDataHash, extDataHash)
-		}
-	} else {
+	if !rules.IsApricotPhase1 {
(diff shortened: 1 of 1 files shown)


ceyonur (approved):

plugin/evm/atomic_block_extension.go
```diff
@@ -90,40 +90,36 @@ func (be *blockExtension) SyntacticVerify(rules extras.Rules) error {
...
+		// hash.
+		//
+		// Otherwise, if the extra data hashes map does not include the block
+		// hash, it is expected for no extra data to have been provided.
+		if expectedExtDataHash, ok := blockExtender.extDataHashes[blockHash]; ok {
+			extDataHash := customtypes.CalcExtDataHash(extData)
+			if extDataHash != expectedExtDataHash {
+				return fmt.Errorf("extra data hash in block (%s, %d): %s, did not match the expected extra data hash: %s", blockHash, b.Height(), extDataHash, expectedExtDataHash)
+			}
+		} else if len(extData) != 0 {
```
alarso16, line 115: I think the error case here is flipped

PR #1022, opened 2025-06-25
feat: enable configurable state history for pruning nodes

alarso16:
## Why this should be merged

To prep for Firewood, we should be able to configure the state history for performance reasons.

## How this works

Rather than using `core.TipToBuffer` to fix the query window to 32, this allows this number to be set in the chain config.

## How this was tested

Unit tests.

## Need to be documented?

Yes - where?

No.


eth/ethconfig/config.go
```diff
@@ -58,7 +57,7 @@ var DefaultConfig = NewDefaultConfig()
 func NewDefaultConfig() Config {
 	return Config{
 		NetworkId:                 0, // enable auto configuration of networkID == chainID
-		StateHistory:              params.FullImmutabilityThreshold,
+		StateHistory:              32,
```
ceyonur, line 60: do we need to change this? I think we already override it in vm config right?

ceyonur (approved):
just a nit

eth/ethconfig/config.go
alarso16: I thought it would effect the tests, but you're correct - I reverted this change.

PR #1027, opened 2025-07-02
feat: c-chain config markdown

owenwahlgren:
## Why this should be merged

switch https://build.avax.network/docs/nodes/chain-configs/c-chain to being remotely fetched from this repo.

## Need to be documented?

No

No


alarso16 (approved):
I think this is great to be in coreth. We may want something similar for subnet-evm.

PR #1029, opened 2025-07-02
feat: Add firewood as a database

alarso16:
## Why this should be merged

This adds much of the Firewood state database implementation, without adding it as a configurable option in coreth.

## How this works

This reimplements everything with `core/state/database.go` by adding the Firewood backend for both reading and hashing from the tries (the `Trie` implementation) and for writing changes (the `triedb.Backend` implementation).

Note: starting this PR, Windows can no longer be built on coreth.

## How this was tested

This is broken off from #959, which is currently bootstrapping to mainnet and Fuji. Additionally, there is a small amount of fuzz testing to ensure that hashing, reading, and committing works as expected.

## Need to be documented?

No.

No - this code is not yet available to consumers.


triedb/firewood/account_trie.go
```diff
@@ -0,0 +1,276 @@
...
+	a.updateValues = append(a.updateValues, []byte{}) // Empty value indicates deletion
+	a.hasChanges = true                               // Mark that there are changes to commit
+	return nil
+}
+
+// Hash implements state.Trie.
+func (a *AccountTrie) Hash() common.Hash {
+	hash, err := a.hash()
+	if err != nil {
+		log.Error("Failed to hash account trie", "error", err)
```
ceyonur, line 193: I think we should return an empty hash here?

ceyonur: ah we do return in a.hash(), but still it's weird to leave it error and not return

triedb/firewood/account_trie.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	return hash
+}
+
+func (a *AccountTrie) hash() (common.Hash, error) {
+	// If we haven't already hashed, we need to do so.
+	if a.hasChanges {
+		root, err := a.fw.getProposalHash(a.parentRoot, a.updateKeys, a.updateValues)
```
ceyonur, line 201: do we always want to keep appending to update keys/update values? or should we clear them once we get a intermediate hash?

core/state/database.go
```diff
@@ -28,24 +28,80 @@
...
+	return wrapIfFirewood(coredb)
+}
+
+func NewDatabaseWithNodeDB(db ethdb.Database, triedb *triedb.Database) Database {
+	coredb := ethstate.NewDatabaseWithNodeDB(db, triedb)
+	return wrapIfFirewood(coredb)
+}
+
+func wrapIfFirewood(db Database) Database {
+	fw, ok := db.TrieDB().Backend().(*firewood.Database)
```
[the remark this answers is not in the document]
alarso16: We don't check the scheme because of a weird quirk of statedb. I set the scheme of Firewood to `rawdb.HashScheme` to avoid iterating through storage slots on account deletions

triedb/firewood/account_trie.go
alarso16: yeah that seems gross, I'll fix it

triedb/firewood/database.go
```diff
@@ -0,0 +1,617 @@
...
+	}
+
+	return config, path, nil
+}
+
+// Scheme returns the scheme of the database.
+// This is only used in some API calls
+// and in StateDB to avoid iterating through deleted storage tries.
+func (db *Database) Scheme() string {
+	return rawdb.HashScheme
```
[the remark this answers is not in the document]
alarso16: I have checked every call site. It's only used in APIs and the one call to statedb. If you know a better way around this, that would be great.

triedb/firewood/account_trie.go
alarso16: We keep appending them. There may be an optimization later in Firewood to store the proposal and "add" additional values later, but for now it's not there.

triedb/firewood/account_trie.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	// No pending update found, read from the underlying reader
+	storageBytes, err := a.reader.Node(common.Hash{}, combinedKey[:], common.Hash{})
+	if err != nil {
+		return nil, err
+	}
+	if storageBytes == nil {
+		return nil, nil
+	}
```
rkuris, line 117: I guess this could be:
```suggestion
	if err != nil || storageBytes == nil {
		return nil, err
	}
```
at the cost of some readability

triedb/firewood/account_trie.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// Decode the storage value
+	_, decoded, _, err := rlp.Split(storageBytes)
```
rkuris, line 120: Same code as before, makes me wonder if we should be computing the thing to rlp.Split and then always doing it here.

triedb/firewood/account_trie.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+// UpdateAccount implements state.Trie.
+func (a *AccountTrie) UpdateAccount(addr common.Address, account *types.StateAccount) error {
+	// Queue the keys and values for later commit
+	key := crypto.Keccak256Hash(addr.Bytes()).Bytes()
+	data, err := rlp.EncodeToBytes(account)
+	if err != nil {
+		return err
+	}
```
rkuris, line 131: My go-fu is very weak but I don't understand how this could fail encoding the account bytes to a new buffer. Out of memory I guess?

triedb/firewood/account_trie.go
```diff
@@ -0,0 +1,275 @@
...
+	data, err := rlp.EncodeToBytes(value)
+	if err != nil {
+		return err
+	}
+
+	// Queue the keys and values for later commit
+	a.dirtyKeys[string(combinedKey[:])] = data
+	a.updateKeys = append(a.updateKeys, combinedKey[:])
+	a.updateValues = append(a.updateValues, data)
+	a.hasChanges = true // Mark that there are changes to commit
```
rkuris, line 156: why can't this be inferred from a.DirtyKeys.len()?

triedb/firewood/account_trie.go
rkuris: You can do this now by creating a new proposal on top of the old one, but proposals are always immutable.

triedb/firewood/database.go
```diff
@@ -0,0 +1,617 @@
...
+			return nil, "", fmt.Errorf("database file path is a directory: %s", path)
+		}
+		// File exists
+		log.Info("Database file found", "path", path)
+		exists = true
+	}
+
+	// Create the Firewood config from the provided config.
+	config := &ffi.Config{
+		Create:            !exists,                               // Use any existing file
```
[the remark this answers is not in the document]
rkuris: Definitely should make the change to firewood. Also several of the earlier checks will result in a firewood error if you only supply a directory.

triedb/firewood/database.go
```diff
@@ -0,0 +1,617 @@
...
+		for _, existing := range existingProposals {
+			// If the block hash is already tracked, we can skip proposing this again.
+			if _, exists := existing.Hashes[hash]; exists {
+				log.Debug("firewood: proposal already exists", "root", root.Hex(), "parent", parentRoot.Hex(), "block", block, "hash", hash.Hex())
+				return nil
+			}
+			// We already have this proposal, but should create a new context with the correct hash.
+			// This solves the case of a unique block hash, but the same underlying proposal.
+			if _, exists := existing.Parent.Hashes[parentHash]; exists {
+				log.Debug("firewood: proposal already exists, updating hash", "root", root.Hex(), "parent", parentRoot.Hex(), "block", block, "hash", hash.Hex())
```
rkuris, line 221: might want a metric here

triedb/firewood/database.go
```diff
@@ -0,0 +1,617 @@
...
+	}
+	return genesisRoot == root
+}
+
+// Update takes a root and a set of keys-values and creates a new proposal.
+// It will not be committed until the Commit method is called.
+// This function should be called even if there are no changes to the state to ensure proper tracking of block hashes.
+func (db *Database) Update(root common.Hash, parentRoot common.Hash, block uint64, nodes *trienode.MergedNodeSet, _ *triestate.Set, opts ...stateconf.TrieDBUpdateOption) error {
+	db.proposalLock.Lock()
+	defer db.proposalLock.Unlock()
```
rkuris, line 199: Is this too early? Can't we wait until we called ExtractTrieDBUpdatePayload?

Also when is this released? Can't we release it while firewood is processing the proposal?

plugin/evm/customrawdb/database_ext.go
```diff
@@ -61,3 +62,71 @@ func InspectDatabase(db ethdb.Database, keyPrefix, keyStart []byte) error {
...
+		return fmt.Errorf("Invalid state path %q", path)
+	}
+
+	// Write the state path to the database
+	return db.Put(chainDataPathKey, []byte(path))
+}
+
+// ReadChainDataPath reads the path to the database files.
+func ReadChainDataPath(db ethdb.KeyValueReader) (string, error) {
+	has, err := db.Has(chainDataPathKey)
```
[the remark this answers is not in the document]
ceyonur: I think we can:
* Use C-chain config file (configBytes, plugin/evm/config) to specify this
* Use `ChainDataDir` in ctx. 

I think we should use `ChainDataDir` if these won't be using the same folder as in avalanchegodb (~/.avalanchego/db` and we can configure the custom path with C-Chain config (like /{ChainDataDir}/{configuredPath}/firewood/) etc. I don't think the latter (having a configurable subpath) is super important though, but we should definetly not persist this to DB. For example the path provided here should be compatible with running C-chain in local, fuji and mainnet at the same time (so they won't conflict each other) and the best way to prevent that is to use `chainDataDir` as each chain will have a separate folder as their ID.

plugin/evm/customrawdb/database_ext.go
```diff
@@ -61,3 +62,71 @@ func InspectDatabase(db ethdb.Database, keyPrefix, keyStart []byte) error {
...
+	path, err := db.Get(chainDataPathKey)
+	if err != nil {
+		return "", err
+	}
+
+	return string(path), nil
+}
+
+// WriteFirewoodGenesisRoot writes the genesis root to the database.
+func WriteFirewoodGenesisRoot(db ethdb.KeyValueWriter, root common.Hash) error {
```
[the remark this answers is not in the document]
ceyonur: root should be available through the genesis block header

triedb/firewood/account_trie.go
```diff
@@ -0,0 +1,276 @@
...
+
+// GetAccount implements state.Trie.
+func (a *AccountTrie) GetAccount(addr common.Address) (*types.StateAccount, error) {
+	key := crypto.Keccak256Hash(addr.Bytes()).Bytes()
+
+	// First check if there's a pending update for this account
+	keyStr := string(key)
+	if updateValue, exists := a.dirtyKeys[keyStr]; exists {
+		// If the value is empty, it indicates deletion
+		if len(updateValue) == 0 {
```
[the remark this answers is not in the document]
ceyonur: seems blobs use the same invariant, let's comment that as a `Invariant: xxx`

triedb/firewood/database.go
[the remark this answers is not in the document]
ceyonur: I think it's definitely worth to have this in firewood

triedb/firewood/database.go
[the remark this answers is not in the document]
ceyonur: I was imagining we could use our own `FirewoodScheme` (but that might be a complete pain to handle them in every call site)

triedb/firewood/account_trie.go
alarso16: Yeah if the writer fails. There's also at least one "unencodable" things, like `interface{} = nil`

core/state/firewood_database.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"fmt"
+
+	"github.com/ava-labs/coreth/triedb/firewood"
+	"github.com/ava-labs/libevm/common"
+)
+
+var (
+	_ Database = (*firewoodAccessorDb)(nil)
```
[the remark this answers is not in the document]
alarso16: Pretty much. My understanding is that this is good practice in general.

triedb/firewood/database.go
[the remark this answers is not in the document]
alarso16: I would much prefer to use `customrawdb.FirewoodScheme`, but because of the weird way statedb.go is written, it will attempt to iterate through the storage keys on account deletion, which is unnecessary.

plugin/evm/customrawdb/database_ext.go
```diff
@@ -61,3 +62,71 @@ func InspectDatabase(db ethdb.Database, keyPrefix, keyStart []byte) error {
 
 	return rawdb.InspectDatabase(db, keyPrefix, keyStart, options...)
 }
+
+// ParseStateSchemeExt parses the state scheme from the provided string.
+// It checks the rawdb package for HashDB or PathDB schemes.
+// If it's neither of those, it checks to see if the string is set to FirewoodScheme.
```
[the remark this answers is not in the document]
alarso16: It's a perfectly reasonable name, but I think it should attempt to match the eth-specific name, like our other overloaded methods from libevm in these `custom*` packages. It does exactly the same thing as `ParseStateScheme`, except also checks for Firewood

core/state/database.go
```diff
@@ -28,24 +28,80 @@
...
-func NewDatabaseWithNodeDB(db ethdb.Database, triedb *triedb.Database) ethstate.Database {
-	return ethstate.NewDatabaseWithNodeDB(db, triedb)
+// CopyTrie returns a deep copy of the given trie.
+// It can be altered by the caller.
+func (db *firewoodAccessorDb) CopyTrie(t Trie) Trie {
+	switch t := t.(type) {
+	case *firewood.AccountTrie:
+		return t.Copy()
+	case *firewood.StorageTrie:
+		return nil // The storage trie just wraps the account trie, so we don't need to copy it separately.
```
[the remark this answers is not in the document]
alarso16: I added some comments (and an explicit method).

plugin/evm/customrawdb/database_ext.go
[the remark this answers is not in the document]
alarso16: This is now completely removed. The database is initialized when opened, so if it is open, then it is initialized.

core/state/database.go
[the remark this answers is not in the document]
alarso16: This could still use the `Copy` method and return nil, but I think this is more clear.

triedb/firewood/database.go
```diff
@@ -190,16 +190,17 @@ func (db *Database) Initialized(root common.Hash) bool {
...
 		log.Error("firewood: no block hash provided for block %d", block)
 	}
 
+	// The rest of the operations except key-value arranging must occur with a lock
+	db.proposalLock.Lock()
+	defer db.proposalLock.Unlock()
+
 	// Check if this proposal already exists.
 	// During reorgs, we may have already created this proposal.
 	// Additinoally, we may have already created this proposal with a different block hash.
```
ceyonur, line 206: nit typo 
```suggestion
	// Additionally, we may have already created this proposal with a different block hash.
```

triedb/firewood/database.go
```diff
@@ -0,0 +1,604 @@
...
+// to recognize the Firewood database.
+func (db *Database) Scheme() string {
+	return rawdb.HashScheme
+}
+
+// Initialized indicates whether the root provided is the genesis root of the database.
+func (db *Database) Initialized(root common.Hash) bool {
+	// If the database is open, then the file is necessarily initialized.
+	// TODO: This would be more informative to have a flag to check that we've state synced later.
+	return db.fwDisk != nil
```
ceyonur, line 186: why don't we check `fw.Root() != common.Hash`?, this looks like we actually always return `Initialized=true` since fwDisk is never nil after `New` but if we haven't commited anything `fw.Root` still can be empty (and actually not initialized)

ceyonur (commented):
few nits and more cleanups + `Initialized` seems always returning `true` even for "uninitialized" dbs.

triedb/firewood/database.go
```diff
@@ -121,38 +116,34 @@ func New(diskdb ethdb.Database, config *Config) *Database {
...
+	if err != nil {
+		return nil, fmt.Errorf("error checking database directory: %w", err)
 	}
-	path := filepath.Join(trieConfig.FilePath, trieConfig.FileName)
 
 	// Check if the file exists
-	info, err = os.Stat(path)
+	info, err = os.Stat(trieConfig.FilePath)
 	exists := false
 	if err == nil {
```
ceyonur, line 141: should we check the specific err here i.e ` os.IsNotExist(err)`?

triedb/firewood/database.go
```diff
@@ -99,17 +101,17 @@ func New(config *Config) *Database {
...
 	fw, err := ffi.New(config.FilePath, fwConfig)
 	if err != nil {
-		log.Crit("firewood: error creating firewood database", "error", err)
+		fmt.Println("firewood: error creating firewood database", "error", err)
 	}
 
 	currentRoot, err := fw.Root()
 	if err != nil {
-		log.Crit("firewood: error getting current root", "error", err)
+		fmt.Println("firewood: error getting current root", "error", err)
```
ceyonur, line 114: we should not use printing for 2 reasons:
* this does not return in case of an error (despite log.Crit where it exits the program)
* rpcchainvms are not compatible with fmt.Println so this won't work with subnet-evm and we really should be using a logger for this.

triedb/firewood/database.go
```diff
@@ -121,38 +116,34 @@ func New(diskdb ethdb.Database, config *Config) *Database {
...
 
-	// Append the filename to the path
-	if trieConfig.FileName == "" {
-		return nil, "", errors.New("no filename provided")
+	// Check that the directory exists
+	dir := filepath.Dir(trieConfig.FilePath)
+	info, err := os.Stat(dir)
+	if err != nil {
+		return nil, fmt.Errorf("error checking database directory: %w", err)
 	}
```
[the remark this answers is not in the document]
alarso16: I put them distinctly to ensure the file can be created if it doesn't exist

triedb/firewood/database.go
```diff
@@ -0,0 +1,610 @@
...
+			return nil, fmt.Errorf("database file path is a directory: %s", trieConfig.FilePath)
+		}
+		// File exists
+		log.Info("Database file found", "path", trieConfig.FilePath)
+		exists = true
+	}
+
+	if err != nil && !os.IsNotExist(err) {
+		return nil, fmt.Errorf("unknown error checking database file: %w", err)
+	}
```
ceyonur, line 158:
```suggestion
        switch {
        case err == nil:
        	if info.IsDir() {
        		return nil, fmt.Errorf("database file path is a directory: %s", trieConfig.FilePath)
        	}
        	log.Info("Database file found", "path", trieConfig.FilePath)
        	exists = true
        case os.IsNotExist(err):
        	log.Info("Database file not found", "path", trieConfig.FilePath)
        default:
        	return nil, fmt.Errorf("error checking database file: %w", err)
        }
```

triedb/firewood/database.go
```diff
@@ -0,0 +1,610 @@
...
+	info, err := os.Stat(dir)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return nil, fmt.Errorf("error checking database directory: %w", err)
+		}
+		log.Info("Database directory not found, creating", "path", dir)
+		if err := os.MkdirAll(dir, 0755); err != nil {
+			return nil, fmt.Errorf("error creating database directory: %w", err)
+		}
+	}
```
ceyonur, line 142: just bc it's hard to follow `!os.IsNotExist` (double negative)
```suggestion
      	if os.IsNotExist(err) {
      		log.Info("Database directory not found, creating", "path", dir)
      		if err := os.MkdirAll(dir, 0755); err != nil {
      			return nil, fmt.Errorf("error creating database directory: %w", err)
      		}
      	} else {
      		return nil, fmt.Errorf("error checking database directory: %w", err)
      	}
```

ceyonur (approved):
just some nit changes

triedb/firewood/database.go
alarso16: Applied

triedb/firewood/database.go
alarso16: This wasn't quite right (didn't handle `err==nil`) so I fixed it locally

PR #1033, opened 2025-07-03
eth/filters: add address limit to filters (#31876)

ceyonur:
Cherry picks https://github.com/ethereum/go-ethereum/pull/31876:

The address filter was never checked against a maximum limit, which can be somewhat abusive for API nodes. This PR adds a limit similar to topics

This pull request introduces a new validation to enforce a maximum limit on the number of addresses allowed in filter criteria for Ethereum logs. It includes updates to the `FilterAPI` and `EventSystem` logic, as well as corresponding test cases to ensure the new constraint is properly enforced.

* **Validation for maximum addresses in filter criteria**:
- Added a new constant, `maxAddresses`, set to 100, to define the maximum allowable addresses in a filter.
- Introduced a new error, `errExceedMaxAddresses`, to handle cases where the number of addresses exceeds the limit.
- Updated the `GetLogs` method in `FilterAPI` to validate the number of addresses against `maxAddresses`.
- Modified the `UnmarshalJSON` method to return an error if the number of addresses in the input JSON exceeds `maxAddresses`.
- Added similar validation to the `SubscribeLogs` method in `EventSystem`.

* **New test cases for address limit validation**:
- Added a test in `TestUnmarshalJSONNewFilterArgs` to verify that exceeding the maximum number of addresses triggers the `errExceedMaxAddresses` error.
- Updated `TestInvalidLogFilterCreation` to include a test case for an invalid filter with more than `maxAddresses` addresses.
- Updated `TestInvalidGetLogsRequest` to test for invalid log requests with excessive addresses.

These changes ensure that the system enforces a reasonable limit on the number of addresses in filter criteria, improving robustness and preventing potential performance issues.

---------

## How this was tested

added UT

## Need to be documented?

no

yes

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -4,6 +4,7 @@
 
 - Removed legacy warp message handlers in favor of ACP-118 SDK handlers.
 - Use `state-history` eth config flag to designate the number of recent states queryable.
+- Added maximum number of addresses (1000) to be queried in a single filter.
 
 ## [v0.15.1](https://github.com/ava-labs/coreth/releases/tag/v0.15.1)
 
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -46,14 +46,19 @@ import (
 )
 
 var (
-	errInvalidTopic      = errors.New("invalid topic(s)")
-	errFilterNotFound    = errors.New("filter not found")
-	errInvalidBlockRange = errors.New("invalid block range params")
-	errExceedMaxTopics   = errors.New("exceed max topics")
+	errInvalidTopic       = errors.New("invalid topic(s)")
+	errFilterNotFound     = errors.New("filter not found")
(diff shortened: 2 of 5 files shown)


alarso16 (approved):
Seems good. I'm not sure if you want to wait until your test fix rolls in. Also, do we need the cherry-picked commit, or can you squash and merge?

michaelkaplan13 (approved):
LGTM. Sanity check for myself: libevm will include these changes when we do a Geth update for it, but this cherrypick does not need to be applied there since the APIs from it are not currently used by coreth/subnet-evm, right?

ceyonur:
> LGTM. Sanity check for myself: libevm will include these changes when we do a Geth update for it, but this cherrypick does not need to be applied there since the APIs from it are not currently used by coreth/subnet-evm, right?

yes

PR #1039, opened 2025-07-08
Remove SnowmanAPI

StephenButtolph:
## Why this should be merged

This API is almost certainly not used by anyone.

## How this works

Deletes code that should have been deleted years ago.

## How this was tested

CI

## Need to be documented?

The API doc has been updated.

✅


plugin/evm/config/config.go
```diff
@@ -98,7 +98,6 @@ type Config struct {
 	GasTarget *gas.Gas `json:"gas-target,omitempty"`
 
 	// Coreth APIs
-	SnowmanAPIEnabled     bool   `json:"snowman-api-enabled"`
```
alarso16, line 101: devrel just moved the public docs to be fetched from `config.md`. You should remove it there too

alarso16 (dismissed):

PR #1044, opened 2025-07-10
Register secp256r1 precompile

### LINKED ISSUE #1042: Register ACP-204 secp256r1 Precompile
[ACP-204](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/204-precompile-secp256r1) proposes adding a secp256r1 precompile, which has already been added to `libevm` in this [PR](https://github.com/ava-labs/libevm/pull/184).

Register the precompile at the address specified in the ACP, and gate it's activation behind a Granite upgrade check.

JonathanOppenheimer:
## Why this should be merged
ACP-204 implemented secp256r1 Curve Support as a precompile to allow cheaper verification of "modern device security systems" that use this curve. It was already added to libevm in https://github.com/ava-labs/libevm/pull/184.

Close #1042 

## How this works
This PR registers the libevm p256verify precompiles as a stateless precompile gatekept by the Granite release

## How this was tested
`precompile/contracts/p256verify/contract_test.go` and standard CI. 

## Need to be documented?
I'm not sure... 

Probably? I'm  also not sure... are users informed when a new precompile is provided?

### DIFF
--- a/params/hooks_libevm.go
+++ b/params/hooks_libevm.go
@@ -113,11 +113,7 @@ func (r RulesExtra) currentPrecompiles() map[common.Address]vm.PrecompiledContra
 // These were only active historically and are not active in the current network.
 func (r RulesExtra) precompileOverrideBuiltin(addr common.Address) (libevm.PrecompiledContract, bool) {
 	precompile, ok := r.currentPrecompiles()[addr]
-	if !ok {
-		return nil, false
-	}
-
-	return precompile, true
+	return precompile, ok
 }
 
 func makePrecompile(contract contract.StatefulPrecompiledContract) libevm.PrecompiledContract {


precompile/registry/registry.go
```diff
@@ -16,6 +17,7 @@ import (
 // Note: it is important that none of these addresses conflict with each other or any other precompiles
 // in /coreth/contracts/contracts/**.
 
-// WarpMessengerAddress = common.HexToAddress("0x0200000000000000000000000000000000000005")
+// WarpMessengerAddress  = common.HexToAddress("0x0200000000000000000000000000000000000005")
+// P256VerifyAddress     = common.HexToAddress("0x0000000000000000000000000000000000000100")
```
JonathanOppenheimer, line 21: Updated address in line with https://github.com/avalanche-foundation/ACPs/pull/222

precompile/modules/registerer.go
```diff
@@ -30,6 +30,11 @@ var (
 			Start: common.HexToAddress("0x0300000000000000000000000000000000000000"),
 			End:   common.HexToAddress("0x03000000000000000000000000000000000000ff"),
 		},
+		// Individual reserved address for p256verify precompile
+		{
+			Start: common.HexToAddress("0x0000000000000000000000000000000000000100"),
+			End:   common.HexToAddress("0x0000000000000000000000000000000000000100"),
+		},
```
ceyonur, line 37: This reserved range is for stateful precompile modules, so this does not apply stateless precompiles. We should not reserve a range for this, otherwise stateful precompiles can register in this reserve.

precompile/registry/registry.go
ceyonur: again this is only for a reference to stateful precompile addresses. we should not include this one here.

params/hooks_libevm.go
```diff
@@ -82,6 +83,12 @@ func (r RulesExtra) ActivePrecompiles(existing []common.Address) []common.Addres
 
 	var addresses []common.Address
 	addresses = slices.AppendSeq(addresses, maps.Keys(precompiles))
+
+	// Add p256verify address if Granite is enabled
+	if r.AvalancheRules.IsGranite {
+		addresses = append(addresses, common.HexToAddress("0x0000000000000000000000000000000000000100"))
+	}
```
ceyonur, line 90: can we keep the same pattern as above where we have each specific contracts for upgrades?

ceyonur: btw we should not redefine the address here but instead use it from libevm's `PrecompiledContractsP256Verify` var.

params/hooks_libevm.go
```diff
@@ -141,6 +148,16 @@ func (r RulesExtra) PrecompileOverride(addr common.Address) (libevm.PrecompiledC
 	if p, ok := r.precompileOverrideBuiltin(addr); ok {
 		return p, true
 	}
+
+	// Handle p256verify as a stateless precompile
+	if addr == common.HexToAddress("0x0000000000000000000000000000000000000100") {
```
ceyonur, line 153: same comment as above: 
* we should not define the address for the precompile
* we should follow a similar approach  to register these as in libevm/go-ethereum


probably best to introduce a new helper function to register these like we do with `precompileOverrideBuiltin`.

params/hooks_libevm.go
```diff
@@ -141,6 +148,16 @@ func (r RulesExtra) PrecompileOverride(addr common.Address) (libevm.PrecompiledC
...
+		// Check if p256verify is enabled at the current timestamp
+		if r.AvalancheRules.IsGranite {
```
ceyonur, line 155: Also we should make sure that this won't break once we have the actual Osaka.

ceyonur (dismissed):

params/hooks_libevm.go
```diff
@@ -149,7 +166,7 @@ func makePrecompile(contract contract.StatefulPrecompiledContract) libevm.Precom
 			}
 		}
 
-		return contract.Run(accessibleState, env.Addresses().Caller, env.Addresses().Self, input, suppliedGas, env.ReadOnly())
+		return contract.Run(accessibleState, env.Addresses().EVMSemantic.Caller, env.Addresses().EVMSemantic.Self, input, suppliedGas, env.ReadOnly())
```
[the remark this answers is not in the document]
JonathanOppenheimer: The code will not compile if I don't change it do to using the new libevm - I'll change it to raw.

JonathanOppenheimer:
> we need to have this is subnet-evm as well

Yup when this is finalized / merged I will sync it to subnet-evm

params/hooks_libevm.go
```diff
@@ -51,33 +51,49 @@ func (RulesExtra) MinimumGasConsumption(x uint64) uint64 {
...
+}
+
+var PrecompiledContractsGranite = map[common.Address]vm.PrecompiledContract{
+	nativeasset.GenesisContractAddr:    makePrecompile(&nativeasset.DeprecatedContract{}),
+	nativeasset.NativeAssetBalanceAddr: makePrecompile(&nativeasset.DeprecatedContract{}),
+	nativeasset.NativeAssetCallAddr:    makePrecompile(&nativeasset.DeprecatedContract{}),
+}
+
+func init() {
+	for addr, precompile := range vm.PrecompiledContractsP256Verify {
```
michaelkaplan13, line 85: `vm.PrecompiledContractsP256Verify` has a comment "This is exported for testing purposes". Seems like we should either remove that comment in libevm, or define the address/precompile mapping ourselves here as the consumer of libevm.

params/hooks_libevm.go
```diff
@@ -149,7 +166,7 @@ func makePrecompile(contract contract.StatefulPrecompiledContract) libevm.Precom
 			}
 		}
 
-		return contract.Run(accessibleState, env.Addresses().Caller, env.Addresses().Self, input, suppliedGas, env.ReadOnly())
+		return contract.Run(accessibleState, env.Addresses().Raw.Caller, env.Addresses().Raw.Self, input, suppliedGas, env.ReadOnly())
```
michaelkaplan13, line 169: It's a bit sub-optimal that this change needs to be bunched in within this PR because it is an important behavior change in of itself, but understand that it's necessary to specify when bumping `libevm`. 

My understanding is that:
1. Currently (i.e. in `v1.13.4`), the `Caller` and `Self` addresses passed to precompiles are the EVM semantic versions.
2. In the Warp precompile, the `Self` address is unused, and except for `delegatecall` or `callcode` call types (which are disallowed), the `Raw` vs `EVMSemantic` `Caller` addresses are equivalent.

Given the above, would it make more sense to pass the `EVMSemantic` addresses here? It shouldn't make a difference for existing precompiles, but future ones could choose to support more functionality. I realize this likely relates to the precompile interface changes, so totally fine if it needs to be deferred for now, but want to make sure what we set here is correct for the time being of course.

params/hooks_libevm.go
michaelkaplan13: Would be curious what @ARR4N thinks when he's back next week. On one hand, defining the address in as few places as possible the better. On the other, since Geth does not register the precompile at a specific address upstream, it's a little odd for libevm to do so for all of its consumers. 

Either way, most importantly I think it just needs to be clear/well documented so that it is clear wherever it is defined that that is the source of truth used to register it by chains.

params/hooks_libevm.go
```diff
@@ -115,7 +131,7 @@ func (r RulesExtra) precompileOverrideBuiltin(addr common.Address) (libevm.Preco
 		return nil, false
 	}
 
-	return makePrecompile(precompile), true
```
ARR4N, line 118: Removing this forces every line of the map-literals to use it. What is the advantage of it not just being done once, here?

params/hooks_libevm.go
ARR4N: > define the address/precompile mapping ourselves here as the consumer of libevm

I think this is the best option. When `libevm` doesn't absolutely need to change it shouldn't and, as @JonathanOppenheimer said, the address isn't going to change on C-Chain. 

> defining the address in as few places as possible the better

Agreed, but that should be kept in `coreth` IMO as `libevm` is nothing more than a provider of a VM in this instance. How it is configured (i.e. the precompile address) is the responsibility of the consumer.

params/hooks_libevm.go
```diff
@@ -98,8 +112,10 @@ func (r RulesExtra) ActivePrecompiles(existing []common.Address) []common.Addres
 // dynamic precompile activation registry.
 // These were only active historically and are not active in the current network.
```
ARR4N, line 113: This comment is out of date.

params/hooks_libevm.go
```diff
@@ -51,33 +51,47 @@ func (RulesExtra) MinimumGasConsumption(x uint64) uint64 {
...
+
+func init() {
+	for addr, precompile := range vm.PrecompiledContractsP256Verify {
+		PrecompiledContractsGranite[addr] = precompile
+	}
 }
 
 func (r RulesExtra) ActivePrecompiles(existing []common.Address) []common.Address {
-	var precompiles map[common.Address]contract.StatefulPrecompiledContract
+	var precompiles map[common.Address]vm.PrecompiledContract
```
ARR4N, line 91: This switch is identical to the one in `precompileOverrideBuiltin` so both should be abstracted into a common method.

params/hooks_libevm.go
ARR4N: It looks like it's because of the mix of `vm.PrecompiledContract` (the one we're registering) and `contract.StatefulPrecompiledContract` (everything else) types. It's messy but I'm not sure how we can get around it without a major refactor elsewhere.

params/hooks_libevm.go
ARR4N: The `init` function will probably become unnecessary.

params/hooks_libevm.go
JonathanOppenheimer: I take it this is no longer true? I'll remove the comment.

ARR4N: Sorry, I think I misunderstood because of the use of a `DeprecatedContract` type that does its own deprecation. Pretend I didn't say anything 🤐

params/hooks_libevm.go
```diff
@@ -79,15 +82,18 @@ var PrecompiledContractsGranite = map[common.Address]vm.PrecompiledContract{
 	nativeasset.GenesisContractAddr:    makePrecompile(&nativeasset.DeprecatedContract{}),
 	nativeasset.NativeAssetBalanceAddr: makePrecompile(&nativeasset.DeprecatedContract{}),
 	nativeasset.NativeAssetCallAddr:    makePrecompile(&nativeasset.DeprecatedContract{}),
+	P256VerifyAddress:                  vm.PrecompiledContractsP256Verify[P256VerifyAddress],
```
ARR4N, line 85: This is risky. If the upstream code changes its address (unlikely but currently still possible) then it will result in the precompile here being nil. I've sent you https://github.com/ava-labs/libevm/pull/222 for review so you can just use `&vm.P256Verify{}` directly once it's merged.

params/hooks_libevm.go
```diff
@@ -79,15 +82,18 @@ var PrecompiledContractsGranite = map[common.Address]vm.PrecompiledContract{
...
-		PrecompiledContractsGranite[addr] = precompile
-	}
+// ActivePrecompiles implements the RulesHooks interface
+func (r RulesExtra) ActivePrecompiles(existing []common.Address) []common.Address {
+	addresses, _, _ := r.ActivePrecompilesWithContract(existing, nil)
+	return addresses
 }
 
-func (r RulesExtra) ActivePrecompiles(existing []common.Address) []common.Address {
+// ActivePrecompiles specifies precompiles that were activated prior to the
```
ARR4N, line 94: Purely informational as I think this method should be removed.

This comment is unclear and doesn't seem to mirror what the code actually does. It leaves a number of open questions:

1. What do you mean by "specifies"?
2. What does the `addr` argument do?
3. What are the returned arguments?

While you don't have to always specify everything explicitly (I'm not expecting JSDoc-style comments), in this case it's not clear to a caller what anything does.

params/hooks_libevm.go
```diff
@@ -79,15 +82,18 @@ var PrecompiledContractsGranite = map[common.Address]vm.PrecompiledContract{
...
+	P256VerifyAddress:                  vm.PrecompiledContractsP256Verify[P256VerifyAddress],
 }
 
-func init() {
-	for addr, precompile := range vm.PrecompiledContractsP256Verify {
-		PrecompiledContractsGranite[addr] = precompile
-	}
+// ActivePrecompiles implements the RulesHooks interface
+func (r RulesExtra) ActivePrecompiles(existing []common.Address) []common.Address {
+	addresses, _, _ := r.ActivePrecompilesWithContract(existing, nil)
```
ARR4N, line 90: The fact that you're dropping the last 2 return arguments and you're passing a `nil` argument into the function are suggestive of an inappropriate abstraction (a.k.a. a [code smell](https://en.wikipedia.org/wiki/Code_smell)), which forces the caller to "massage" the code into doing what's necessary instead of it just flowing.

Have a look at https://github.com/ava-labs/coreth/commit/79032da35 for an alternative abstraction (the `currentPrecompiles()` method). It's based off the commit that I originally reviewed, not this current one, so you can more clearly see what I've done.

ARR4N (changes requested):

params/hooks_libevm.go
JonathanOppenheimer: resolving in favor of other comment.

PR #1050, opened 2025-07-15
refactor: re-exec loop check

alarso16:
## Why this should be merged

`reprocessState` unnecessarily re-executes up to `bc.commitInterval` blocks in some cases, when the most recent commit occurs at the acceptor tip.

## How this works

Instead of immediately decrementing the block number to check, the current block is checked first.

## How this was tested

Additional unit tests were created to ensure equivalent functionality

## Need to be documented?

No.

No?

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1806,8 +1806,14 @@ func (bc *BlockChain) reprocessState(current *types.Block, reexec uint64) error 
 		}
 	}
 
-	for i := 0; i < int(reexec); i++ {
+	for i := 0; i <= int(reexec); i++ {
 		// TODO: handle canceled context
+
+		// Once we have a state, we can reprocess from there.
+		_, err = bc.stateCache.OpenTrie(current.Root())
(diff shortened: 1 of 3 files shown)


core/blockchain.go
```diff
@@ -1806,9 +1806,15 @@ func (bc *BlockChain) reprocessState(current *types.Block, reexec uint64) error
 		}
 	}
 
-	for i := 0; i < int(reexec); i++ {
+	for i := 0; i <= int(reexec); i++ {
 		// TODO: handle canceled context
 
+		// Once we have a state, we can reprocess from there.
+		_, err = bc.stateCache.OpenTrie(current.Root())
```
ceyonur, line 1813: It seems we can actually use `HasState` over this and once the loop ends and if there is no parent found with state we can just return a crafted error (like below). I think this would simplify this even further as we don't need to do this extra error handling for no reason.

ceyonur (approved):

PR #1052, opened 2025-07-17
chore: Update Firewood to v0.0.9

alarso16:
## Why this should be merged

New version of Firewood released, has more features for metrics (coming soon).

## How this works

The only breaking change in the Firewood interface removed the need to explicitly state whether a file should be created or not - now Firewood will internally determine whether to create a file or read it, even if the file is empty (e.g. `filepath.Join()` creates an empty file)

## How this was tested

Existing UT

## Need to be documented?

No.

No?


JonathanOppenheimer (approved):
LGTM

RodrigoVillar (approved):
LGTM. Regarding the change in `prometheus_test.go`, I think this is because of an upgrade in Protobuf versions somewhere here: https://github.com/prometheus/client_golang/compare/v1.16.0...v1.22.0

PR #1053, opened 2025-07-18
Remove custom `state.StateDB`

StephenButtolph:
## Why this should be merged

This allows `core.wrapStateDB` to be called with a `libevm` `state.StateDB` which is required for SAE.

## How this works

Currently there are 4 `StateDB` implementations:
1. libevm's `state.StateDB` implementation.
2. coreth's `state.StateDB` implementation.
3. coreth's `core.StateDbAP1` implementation.
4. coreth's `extstate.StateDB` implementation.

coreth's `state.StateDB` and `extstate.StateDB` are similar in that they add additional functionality to the `StateDB` they are wrapping. The additional functionality is merged into a single `StateDB` (as `extstate.StateDB`).

Additionally, coreth's `state.StateDB` included state key normalization, this is migrated into the libevm `StateDBHooks` and registered in `extstate/statedb.go`.

Finally the `core.StateDbAP1` implementation is replaced with a `core.StateDBAP0` implementation, to correctly handle the `libevm` hook usage.

## How this was tested

Pretty minimally tbh.


go.mod
```diff
@@ -4,9 +4,9 @@ go 1.23.9
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.12.1
-	github.com/ava-labs/avalanchego v1.13.3-rc.1
+	github.com/ava-labs/avalanchego v1.13.2-rc.0
```
StephenButtolph, line 7: I don't know how to get the libevm version updated without downgrading avalanchego...

core/state/statedb.go
```diff
@@ -84,96 +47,9 @@ func (wp *workerPool) Done() {
...
-}
-
-// SetTxContext sets the current transaction hash and index which are
-// used when the EVM emits new state logs. It should be invoked before
-// transaction execution.
-func (s *StateDB) SetTxContext(thash common.Hash, ti int) {
-	s.thash = thash
-	s.txIndex = ti
-	s.StateDB.SetTxContext(thash, ti)
-}
```
StephenButtolph, line 148: This function was the only reason this struct needed to exist.

Specifically, this function could not be moved to the `extstate.StateDB` implementation because the statedb is wrapped _after_ `SetTxContext` is called.

This function supported the `GetTxHash()` function below.

LibEVM added a `TxHash()` function (to align with the `TxIndex()` function) which allows for us to use the `TxHash()` function rather than the `GetTxHash()` function, which meant that we could remove this entirely.

core/evm.go
```diff
@@ -77,19 +78,19 @@ func (hooks) OverrideEVMResetArgs(rules params.Rules, args *vm.EVMResetArgs) *vm
...
-	extstate.VmStateDB
+type StateDBAP0 struct {
+	*extstate.StateDB
 }
 
-func (s *StateDbAP1) GetCommittedState(addr common.Address, key common.Hash) common.Hash {
-	state.NormalizeStateKey(&key)
-	return s.VmStateDB.GetCommittedState(addr, key)
+func (s *StateDBAP0) GetCommittedState(addr common.Address, key common.Hash, _ ...stateconf.StateDBStateOption) common.Hash {
+	return s.StateDB.GetCommittedState(addr, key, stateconf.SkipStateKeyTransformation())
```
alarso16, line 93: It's unintuitive that the default behavior in libevm is to transform the state key. I guess it's not in this repo and you already merged it, but worth mentioning

core/extstate/state_object_test.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"testing"
+
+	"github.com/ava-labs/libevm/common"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestStateObjectPartition(t *testing.T) {
+	keys := []common.Hash{
+		{},
```
alarso16, line 15: Why are these the only keys tested? Seems like something like `0x00FFFFFFFFFFF` (and all other conbos for the first byte) would be worth adding. If not, then maybe add a comment why. It's not obvious to me

core/evm.go
alarso16: Edit: Ah this is because you already registered the extra state transform in the `init()` function of `extstate`? This isn't obvious behavior, i would doc that here somewhere

core/extstate/statedb_multicoin_test.go
```diff
@@ -1,133 +1,147 @@
...
+	"github.com/ava-labs/libevm/core/types"
 	"github.com/ava-labs/libevm/crypto"
 	"github.com/ava-labs/libevm/libevm/stateconf"
 	"github.com/holiman/uint256"
 	"github.com/stretchr/testify/require"
 )
 
 func TestMultiCoinOperations(t *testing.T) {
-	s := newStateEnv()
+	rawdb := rawdb.NewMemoryDatabase()
```
alarso16, line 23: This naming is weird (not conventional in coreth). Instead of `rawdb`, maybe do `db` or `ethdb` or `memdb`

core/extstate/statedb_multicoin_test.go
```diff
@@ -1,133 +1,147 @@
...
-	root, err := s.state.Commit(0, false)
-	require.NoError(t, err, "committing state")
-	s.state, err = New(root, s.state.db, s.state.snaps)
-	require.NoError(t, err, "creating statedb")
-
-	s.state.AddBalance(addr, new(uint256.Int))
+	stateDB.AddBalance(addr, new(uint256.Int))
 
-	balance := s.state.GetBalanceMultiCoin(addr, assetID)
+	state := New(stateDB)
```
alarso16, line 33: nit: These overlapping names is hard
```suggestion
	wrappedStateDB := New(stateDB)
```

core/extstate/statedb_multicoin_test.go
```diff
@@ -1,133 +1,147 @@
...
 	// Create empty [snapshot.Tree] and [StateDB]
 	root := common.HexToHash("0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")
 	// Use the root as both the stateRoot and blockHash for this test.
-	snapTree := snapshot.NewTestTree(db, root, root)
+	snapTree := snapshot.NewTestTree(rawdb, root, root)
 
 	addr := common.Address{1}
 	assetID1 := common.Hash{1}
 	assetID2 := common.Hash{2}
+	assertBalances := func(t *testing.T, state *StateDB, regular, multicoin1, multicoin2 int64) {
```
alarso16, line 57: nit:
```suggestion
	assertBalances := func(t *testing.T, sdb *StateDB, regular, multicoin1, multicoin2 int64) {
```

core/txpool/blobpool/blobpool_test.go
```diff
@@ -61,8 +61,8 @@ import (
 
 var (
 	emptyBlob          = kzg4844.Blob{}
-	emptyBlobCommit, _ = kzg4844.BlobToCommitment(emptyBlob)
-	emptyBlobProof, _  = kzg4844.ComputeBlobProof(emptyBlob, emptyBlobCommit)
+	emptyBlobCommit, _ = kzg4844.BlobToCommitment(&emptyBlob)
```
alarso16, line 64: Side effect of libevm update?

nativeasset/contract_test.go
```diff
@@ -129,9 +125,10 @@ func TestStatefulPrecompile(t *testing.T) {
 				// Set balance to pay for gas fee
 				statedb.SetBalance(userAddr1, u256Hundred)
 				// Initialize multicoin balance and set it back to 0
-				statedb.AddBalanceMultiCoin(userAddr1, assetID, bigHundred)
-				statedb.SubBalanceMultiCoin(userAddr1, assetID, bigHundred)
-				statedb.Finalise(true)
+				state := extstate.New(statedb)
```
alarso16, line 128: nit:
```suggestion
				wrappedStateDB := extstate.New(statedb)
```

plugin/evm/export_tx_test.go
```diff
@@ -2000,8 +2009,9 @@ func TestNewExportTxMulticoin(t *testing.T) {
 			if stdb.GetBalance(addr).Cmp(uint256.NewInt(test.bal*units.Avax)) != 0 {
 				t.Fatalf("address balance %s equal %s not %s", addr.String(), stdb.GetBalance(addr), new(big.Int).SetUint64(test.bal*units.Avax))
 			}
-			if stdb.GetBalanceMultiCoin(addr, common.BytesToHash(tid[:])).Cmp(new(big.Int).SetUint64(test.balmc)) != 0 {
-				t.Fatalf("address balance multicoin %s equal %s not %s", addr.String(), stdb.GetBalanceMultiCoin(addr, common.BytesToHash(tid[:])), new(big.Int).SetUint64(test.balmc))
+			state := extstate.New(stdb)
```
alarso16, line 2012: Why not use the same `exstdb` as above? Seems like this instance only needs to be created once

alarso16 (changes requested):
Mostly looks good, but some relatively nit changes and requests for inline comments

core/txpool/blobpool/blobpool_test.go
StephenButtolph: Yeah this is because of the libevm update. This is specifically because of the `kzg4844` library version bump to `v1` which was added because our usage of the older version broke people that depended on both geth and libevm

core/evm.go
StephenButtolph: I tried to add some docs on it, lmk if there is somewhere specific you'd like it added.

alarso16: yeah this is def sufficient

alarso16 (approved):

PR #1056, opened 2025-07-18
Replace coreth state package with libevm state package

StephenButtolph:
## Why this should be merged

My understanding is that whenever we can just use `libevm`, we are better off.


JonathanOppenheimer (approved):
🤩🤩

Q: can we also delete `coreth/core/state` entirely? Looks like you've replaced it in 34/40 of its total usages? 

edit: nvm maybe it's not that easy

alarso16 (commented):
The test failure is weird - that is definitely not a flaky test.
Is it because of `core/state/firewood_database.go`?

Edit: test is flaky, I'm investigating

core/state/state_test.go
StephenButtolph, line 1: This file doesn't seem to have been testing anything specific about our modifications to the statedb struct, so it is just removed entirely.

alarso16 (approved):
So much better!

PR #1059, opened 2025-07-21
fix: Ensure Firewood does not return data on unimplemented APIs

alarso16:
## Why this should be merged

Some of the eth APIs use `trie.StateTrie`, which is specific to geth's two-tier in-memory trie. These are not safe to be used with Firewood, since the intermediate nodes will not be found as expected. 

A long-term fix, if these APIs are necessary, would be to implement a Firewood-specific iterator in the FFI, and an easily-parsable proof method.

## How this works

This just returns an error if Firewood is used for these nodes.

## How this was tested

This is really hard to test. It can only occur in a production environment.

## Need to be documented?

Yes.

Yes.

### DIFF
--- a/eth/api_debug.go
+++ b/eth/api_debug.go
@@ -34,6 +34,7 @@ import (
 	"time"
 
 	"github.com/ava-labs/coreth/internal/ethapi"
+	"github.com/ava-labs/coreth/plugin/evm/customrawdb"
 	"github.com/ava-labs/coreth/rpc"
 	"github.com/ava-labs/libevm/common"
 	"github.com/ava-labs/libevm/common/hexutil"
@@ -45,6 +46,8 @@ import (
 	"github.com/ava-labs/libevm/rlp"
 	"github.com/ava-labs/libevm/trie"
 )
+
+var errFirewoodNotSupported = errors.New("firewood triedb scheme does not yet support this operation")
 
 // DebugAPI is the collection of Ethereum full node APIs for debugging the
 // protocol.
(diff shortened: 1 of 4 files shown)


eth/api_debug.go
```diff
@@ -173,8 +176,11 @@ type storageEntry struct {
 
 // StorageRangeAt returns the storage at the given block height and transaction index.
 func (api *DebugAPI) StorageRangeAt(ctx context.Context, blockNrOrHash rpc.BlockNumberOrHash, txIndex int, contractAddress common.Address, keyStart hexutil.Bytes, maxResult int) (StorageRangeResult, error) {
-	var block *types.Block
+	if api.eth.blockchain.CacheConfig().StateScheme == customrawdb.FirewoodScheme {
```
powerslider, line 179: Maybe you can extract this repetitive boolean condition to a method and reuse it in the other methods?

alarso16: I could, but is it worth it? I'm not sure. If you think it is, I can add it

JonathanOppenheimer: I don't really think it's worth it tbh

eth/api_debug.go
```diff
@@ -173,8 +176,11 @@ type storageEntry struct {
...
+		return StorageRangeResult{}, errFirewoodNotSupported
+	}
 
+	var block *types.Block
 	block, err := api.eth.APIBackend.BlockByNumberOrHash(ctx, blockNrOrHash)
```
JonathanOppenheimer, line 184: Is adding the additional block declaration necessary? I know `err` has to be declared and assigned to but I've not seen this pattern before.

JonathanOppenheimer (changes requested):
Additionally, If the bloom test is flaky, but fine I would add it to `scripts/known_flakes.txt` so you don't get the test failures.

eth/api_debug.go
powerslider: ok, all good

plugin/evm/config/config.md
```diff
@@ -186,6 +186,11 @@ Adds the following RPC calls to the `debug_*` namespace. Defaults to `false`.
 - `debug_getModifiedAccountsByHash`
 - `debug_getAccessibleState`
 
+The following RPC calls are disabled for any nodes with `state-scheme = firewood`:
```
JonathanOppenheimer, line 189: Should this be modified to plug into buildershub? I think this is the old style of documentation, no?

JonathanOppenheimer: my bad!

JonathanOppenheimer (approved):

plugin/evm/config/config.md
alarso16: This is the plug to builders-hub

PR #1065, opened 2025-07-23
Optimize build_test.sh to not rereun tests on flake failure   

JonathanOppenheimer:
## Why this should be merged
Currently, if a single test fails, the entire test-suite is reran (up to 4 possible times). This is extremely inefficient as if a flaky test fails, the test suite will rerun, but then maybe a different flaky test fails, so the whole test-suite is rerun, etc. This leads to long time waiting for CI jobs to pass in GitHub, which is a large waste of time. 

The prior claim that re-running all prior tests was necessary is copied below:

>  Note the absence of unexpected failures cannot be indicative that we only need to run the tests that failed, or example a test may panic and cause subsequent tests in that package to not run. 

However, **_there may not be an unexpected failure that causes subsequent tests not to run_** -- if a single flaky test fails there is no need to rerun the whole test-suite. Additionally, even if there is a panic that causes subsequent tests not to run, there's still no reason to re-run the whole test-suite -- just the tests that failed, and the tests that never ran. 

Scenario | Before Behavior | After Behavior
-- | -- | --
All tests pass | Exit immediately | Exit immediately 
Unexpected failures | Exit immediately | Exit immediately
All tests run, only known flakes fail | Rerun ALL tests  | Retry only failed tests 
Some tests don't run due to panics | Rerun ALL tests  | Rerun all tests in packages in which tests panicked + failed flaky tests

## How this works
1. Get list of all expected tests for the package using and store for comparison
2. Run tests with `go test -json -shuffle=on` for structured output
3. Parse test output to extract failed tests using regex patterns for both failures and panics
4. Compare expected vs ran tests to find tests that didn't run due to panics
5.  Categorize failures into known flakes vs unexpected failures
6. Depending on whether tests failed, or tests were skipped, target retry relevant tests. 

## How this was tested
This modifies the test framework directly. 

## Need to be documented?
No

No


scripts/build_test.sh
```diff
@@ -21,38 +21,187 @@ if [[ -n "${NO_RACE:-}" ]]; then
...
+    
+    # Run the specific test
+    go test -shuffle=on ${race:-} -timeout="${TIMEOUT:-600s}" -run "^${test_name}$" "$package" 2>&1 | tee "$output_file"
+    return "${PIPESTATUS[0]}"
+}
+
+# Initial test run
+echo "Running initial test suite..."
+# shellcheck disable=SC2046
+go test -shuffle=on ${race:-} -timeout="${TIMEOUT:-600s}" -coverprofile=coverage.out -covermode=atomic "$@" $(go list ./... | grep -v github.com/ava-labs/coreth/tests) | tee test.out || command_status=$?
```
JonathanOppenheimer, line 100: Not sure if it's possible to write this line so it both 
A. works
B. passes lint

scripts/known_flakes.txt
```diff
@@ -3,6 +3,7 @@ TestExpDecaySampleNanosecondRegression
 TestGolangBindings
 TestMempoolAtmTxsAppGossipHandlingDiscardedTx
 TestMempoolEthTxsAppGossipHandling
+TestPruningBlockChainSnapsDisabled
```
alarso16, line 6: This one is fixed - it should not be included here.

JonathanOppenheimer: Got it!

JonathanOppenheimer:
> Why did you decide to write this in bash? The complexity involved would really benefit from a real programming language (i.e. golang).

Not really any particular reason besides sticking with the current structure (order remains the same, `run_task.sh` and so on). If you find this overly complex, I can rewrite it in Python (or golang is no Python)

scripts/build_test.sh
JonathanOppenheimer: Interesting! That's definitely something that is worth bringing over to the evm repositories in my opinion.

PR #1068, opened 2025-07-24
refactor(sync): unify atomic and state sync under common interface

### LINKED ISSUE #1067: Move atomic trie sync code closely to state trie sync code and standardize interfaces
Currently atomic trie sync code can be found under `plugin/evm/atomic` and state trie sync code is at `sync/statesync`. Given that they have lots in common when it comes to shared functionality and interfaces it makes more sense to move them together and also share a common interface for `Start()` and `Wait()`.

powerslider:
## Why this should be merged

This change unifies all sync-related functionality under a single parent package
with consistent interfaces and patterns, improving code organization and
maintainability.

## How this works

  - Move atomic sync from plugin/evm/atomic/sync to sync/atomic for unified package structure.
  - Create common sync.Syncer interface with Start() and Wait() methods.
  - Update atomic state sync to implement unified interface.
  - Update state trie sync to implement unified interface.
  - Update code syncer to implement unified interface.
  - Remove deprecated Done() method in favor of Wait() method across all sync implementations.
  - Add ErrWaitBeforeStart for consistent error handling.
  - Ensure all sync implementations respect context cancellation.
  - Maintain backward compatibility with existing atomic functionality in plugin/evm/atomic.

## How this was tested

Existing tests are updated with the new unified interface.

## Need to be documented?
no

no

resolves #1067

### DIFF
--- a/triedb/firewood/database.go
+++ b/triedb/firewood/database.go
@@ -503,6 +503,9 @@ type reader struct {
 // Node retrieves the trie node with the given node hash. No error will be
 // returned if the node is not found.
 func (reader *reader) Node(_ common.Hash, path []byte, _ common.Hash) ([]byte, error) {
+	// TODO: remove these locks once Firewood supports concurrent reads and commits.
+	reader.db.proposalLock.RLock()
+	defer reader.db.proposalLock.RUnlock()
 	// This function relies on Firewood's internal locking to ensure concurrent reads are safe.
 	// This is safe even if a proposal is being committed concurrently.
 	start := time.Now()


sync/statesync/state_syncer.go
```diff
@@ -224,13 +224,14 @@ func (t *stateSync) storageTrieProducer(ctx context.Context) error {
...
 func (t *stateSync) Start(ctx context.Context) error {
-	// Create a cancellable context for the sync operations
-	syncCtx, cancel := context.WithCancel(ctx)
-	t.cancelFunc = cancel
+	ctx, t.cancelFunc = context.WithCancel(ctx)
 
 	// Start the code syncer and leaf syncer.
-	eg, egCtx := errgroup.WithContext(syncCtx)
-	t.codeSyncer.start(egCtx) // start the code syncer first since the leaf syncer may add code tasks
+	eg, egCtx := errgroup.WithContext(ctx)
```
alarso16, line 230: What's with these changes?

sync/statesync/sync_test.go
```diff
@@ -68,13 +68,16 @@ func testSync(t *testing.T, test syncTest) {
 	})
 	require.NoError(t, err, "failed to create state syncer")
 	// begin sync
-	s.Start(ctx)
-	waitFor(t, context.Background(), s.Wait, test.expectedError, testSyncTimeout)
-	if test.expectedError != nil {
-		return
+	if err := s.Start(ctx); err != nil {
```
alarso16, line 71: nit: `require` instead of `t.Fatal`

sync/statesync/sync_test.go
```diff
@@ -68,13 +68,16 @@ func testSync(t *testing.T, test syncTest) {
...
+		t.Fatal(err)
 	}
 
-	assertDBConsistency(t, root, clientDB, serverTrieDB, triedb.NewDatabase(clientDB, nil))
+	waitFor(t, context.Background(), s.Wait, test.expectedError, testSyncTimeout)
+
+	// Only assert database consistency if the sync was expected to succeed
```
alarso16, line 77: I don't think this should be flipped.

sync/vm/client.go
alarso16, line 1: Not sure we should move these files, definitely worth a discussion

sync/statesync/sync_test.go
powerslider: The `assertDBConsistency` is being called before the sync completes. The sync is starts with `s.Start(ctx)` but then immediately `assertDBConsistency` is called, and only after that does it wait for the sync to complete with `waitFor(t, context.Background(), s.Wait, test.expectedError, testSyncTimeout)`. If they are not flipped the test fails.

sync/vm/client.go
alarso16: I think this is ok. We should be able to separate the syncing logic from the client and server, so it makes sense it could be shared with subnet-evm

sync/statesync/state_syncer.go
JonathanOppenheimer: Yeah this is the only functional change I see - it's also a different return as this wasn't caught prior?

JonathanOppenheimer (approved):

JonathanOppenheimer:
but yeah this is very helpful - especially to move more code packages out of coreth - as I told @powerslider over Slack, we'll need to replicate these changes in `subnet-evm,` which I hope will not be too difficult

sync/statesync/state_syncer.go
powerslider: It's the same, it is just using the common Syncer interface's Start method.

PR #1078, opened 2025-07-28
refactor(sync): add `blockSyncer` as a new syncer implementation

alarso16:
## Why this should be merged

Requesting recent blocks should follow the same format as all other syncers.

## How this works

Wraps the old call in a `Start` and `Wait` call with cancel logic

## How this was tested

This code was completely untested before, so I added basic unit tests with a mock client to test functionality. These tests are not too thorough though

## Need to be documented?

No

No

### DIFF
--- a/sync/blocksync/syncer.go
+++ b/sync/blocksync/syncer.go
@@ -1,0 +1,137 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package blocksync
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	synccommon "github.com/ava-labs/coreth/sync"
+	statesyncclient "github.com/ava-labs/coreth/sync/client"
+	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/libevm/core/rawdb"
+	"github.com/ava-labs/libevm/ethdb"
+	"github.com/ava-labs/libevm/log"
+)
+
+var (
+	_ synccommon.Syncer = (*blockSyncer)(nil)
+
+	errNilClient       = errors.New("Client cannot be nil")
+	errNilDatabase     = errors.New("Database cannot be nil")
+	errInvalidFromHash = errors.New("FromHash cannot be empty")
+)
+
+type BlockSyncerConfig struct {
+	ChainDB      ethdb.Database
+	Client       statesyncclient.Client
+	FromHash     common.Hash
+	FromHeight   uint64
+	ParentsToGet uint64
+}
+
+func (c *BlockSyncerConfig) Validate() error {
+	if c.ChainDB == nil {
+		return errNilDatabase
+	}
+	if c.Client == nil {
+		return errNilClient
+	}
+	if c.FromHash == (common.Hash{}) {
+		return errInvalidFromHash
+	}
+	return nil
+}
+
+type blockSyncer struct {
+	*BlockSyncerConfig
+
+	err    chan error
+	cancel context.CancelFunc
+}
+
+func NewBlockSyncer(config *BlockSyncerConfig) (*blockSyncer, error) {
+	if err := config.Validate(); err != nil {
+		return nil, err
+	}
+	return &blockSyncer{
+		BlockSyncerConfig: config,
+		err:               make(chan error),
+	}, nil
+}
+
+func (syncer *blockSyncer) Start(ctx context.Context) error {
+	cancelCtx, cancel := context.WithCancel(ctx)
+	syncer.cancel = cancel
+	go func() {
+		syncer.err <- syncer.syncBlocks(cancelCtx)
+	}()
+	return nil
+}
+
+func (syncer *blockSyncer) Wait(ctx context.Context) error {
+	if syncer.cancel == nil {
+		return synccommon.ErrWaitBeforeStart
+	}
+
+	select {
+	case err := <-syncer.err:
+		return err
+	case <-ctx.Done():
+		syncer.cancel()
+		<-syncer.err // wait for the syncer to finish
+		return ctx.Err()
+	}
+}
+
+// syncBlocks fetches (up to) [parentsToGet] blocks from peers
+// using [client] and writes them to disk.
+// the process begins with [fromHash] and it fetches parents recursively.
+// fetching starts from the first ancestor not found on disk
+func (syncer *blockSyncer) syncBlocks(ctx context.Context) error {
+	nextHash := syncer.FromHash
+	nextHeight := syncer.FromHeight
+	parentsPerRequest := uint16(32)
+
+	// first, check for blocks already available on disk so we don't
+	// request them from peers.
+	for syncer.ParentsToGet >= 0 {
+		blk := rawdb.ReadBlock(syncer.ChainDB, nextHash, nextHeight)
+		if blk != nil {
+			// block exists
+			nextHash = blk.ParentHash()
+			nextHeight--
+			syncer.ParentsToGet--
+			continue
+		}
+
+		// block was not found
+		break
+	}
+
+	// get any blocks we couldn't find on disk from peers and write
+	// them to disk.
+	batch := syncer.ChainDB.NewBatch()
+	for i := syncer.ParentsToGet - 1; i >= 0 && (nextHash != common.Hash{}); {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		blocks, err := syncer.Client.GetBlocks(ctx, nextHash, nextHeight, parentsPerRequest)
+		if err != nil {
+			return fmt.Errorf("could not get blocks from peer: err: %w, nextHash: %s, remaining: %d", err, nextHash, i+1)
+		}
+		for _, block := range blocks {
+			rawdb.WriteBlock(batch, block)
+			rawdb.WriteCanonicalHash(batch, block.Hash(), block.NumberU64())
+
+			i--
+			nextHash = block.ParentHash()
+			nextHeight--
+		}
+		log.Info("fetching blocks from peer", "remaining", i+1, "total", syncer.ParentsToGet)
+	}
+	log.Info("fetched blocks from peer", "total", syncer.ParentsToGet)
+	return batch.Write()
+}
--- a/sync/blocksync/syncer.go
+++ b/sync/blocksync/syncer.go
@@ -19,12 +19,13 @@ import (
 var (
 	_ synccommon.Syncer = (*blockSyncer)(nil)
 
-	errNilClient       = errors.New("Client cannot be nil")
-	errNilDatabase     = errors.New("Database cannot be nil")
-	errInvalidFromHash = errors.New("FromHash cannot be empty")
+	errNilClient                     = errors.New("Client cannot be nil")
+	errNilDatabase                   = errors.New("Database cannot be nil")
+	errInvalidFromHash               = errors.New("FromHash cannot be empty")
+	errParentsToGetExceedsFromHeight = errors.New("ParentsToGet cannot exceed FromHeight")
 )
 
-type BlockSyncerConfig struct {
+type Config struct {
 	ChainDB      ethdb.Database
 	Client       statesyncclient.Client
 	FromHash     common.Hash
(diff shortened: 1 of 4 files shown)


sync/blocksync/syncer.go
```diff
@@ -0,0 +1,137 @@
...
+	if err := config.Validate(); err != nil {
+		return nil, err
+	}
+	return &blockSyncer{
+		BlockSyncerConfig: config,
+		err:               make(chan error),
+	}, nil
+}
+
+func (syncer *blockSyncer) Start(ctx context.Context) error {
```
powerslider, line 65: `syncer` is too long for a method receiver, use `s` or `b`.

powerslider (approved):

sync/vm/client.go
```diff
@@ -34,6 +34,7 @@ const (
 	ParentsToFetch = 256
 
 	atomicStateSyncOperationName = "Atomic State Sync"
+	blockStateSyncOperationName  = "Block State Sync"
```
StephenButtolph, line 37: The blocks really aren't "state".
```suggestion
	blockSyncOperationName       = "Block Sync"
```

sync/blocksync/syncer.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	if c.Client == nil {
+		return errNilClient
+	}
+	if c.FromHash == (common.Hash{}) {
+		return errInvalidFromHash
+	}
+	if c.ParentsToGet > c.FromHeight {
+		return errParentsToGetExceedsFromHeight
+	}
```
StephenButtolph, line 48: Is this a new constraint? It seems like the old code (while implemented weirdly) seemed to handle this case

sync/blocksync/syncer.go
```diff
@@ -0,0 +1,145 @@
...
+// not available.
+func (s *blockSyncer) sync(ctx context.Context) error {
+	nextHash := s.config.FromHash
+	nextHeight := s.config.FromHeight
+	parentsToGet := int64(s.config.ParentsToGet)
+	parentsPerRequest := uint16(32)
+
+	// first, check for blocks already available on disk so we don't
+	// request them from peers.
+	for parentsToGet >= 0 {
```
StephenButtolph, line 109: If `parentsToGet == 0`, then `i := parentsToGet - 1` will equal `-1`, which is not `i >= 0`, so there is no reason to enter this loop when `parentsToGet == 0`.
```suggestion
	for parentsToGet > 0 {
```

Additionally, this will prevent the very odd log:
```
fetched blocks from peer total=-1
```
when all the blocks are locally available.

sync/blocksync/syncer.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+		for _, block := range blocks {
+			rawdb.WriteBlock(batch, block)
+			rawdb.WriteCanonicalHash(batch, block.Hash(), block.NumberU64())
+
+			i--
+			nextHash = block.ParentHash()
+			nextHeight--
+		}
+		log.Info("fetching blocks from peer", "remaining", i+1, "total", parentsToGet)
```
StephenButtolph, line 141: This log doesn't really make much sense here. We should move this to be before the call to `GetBlocks`.

sync/blocksync/syncer.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	// get any blocks we couldn't find on disk from peers and write
+	// them to disk.
+	batch := s.config.ChainDB.NewBatch()
+	for i := parentsToGet - 1; i >= 0 && (nextHash != common.Hash{}); {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		blocks, err := s.config.Client.GetBlocks(ctx, nextHash, nextHeight, parentsPerRequest)
```
StephenButtolph, line 129: While I don't think this is broken, it is interesting that we can end up requesting more blocks than we actually need.

We still verify that they are the correct blocks, but we may end up writing more blocks than `ParentsToGet` would make it seem like we will.

Should we put a TODO to make `parentsPerRequest` be `maxBlocksPerRequest` and then we set `blocksToRequest := min(maxBlocksPerRequest, blocksRemaining)`? Or we could just do that now.

sync/blocksync/syncer.go
```diff
@@ -0,0 +1,145 @@
...
+		parentsToGet--
+	}
+
+	// get any blocks we couldn't find on disk from peers and write
+	// them to disk.
+	batch := s.config.ChainDB.NewBatch()
+	for i := parentsToGet - 1; i >= 0 && (nextHash != common.Hash{}); {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
```
StephenButtolph, line 128: No action required, but we probably don't need to actually check the context here, as the client implementation should be handling the cancelled contexts correctly

sync/vm/client.go
```diff
@@ -195,6 +201,16 @@ func (client *client) registerSyncers(ctx context.Context, registry *SyncerRegis
 	return nil
 }
 
+func (client *client) createBlockSyncer(ctx context.Context, fromHash common.Hash, fromHeight uint64, parentsToGet int) (synccommon.Syncer, error) {
```
StephenButtolph, line 204:
```suggestion
func (client *client) createBlockSyncer(ctx context.Context, fromHash common.Hash, fromHeight uint64, numToFetch uint64) (synccommon.Syncer, error) {
```

sync/blocksync/syncer.go
```diff
@@ -0,0 +1,145 @@
...
+	errInvalidFromHash               = errors.New("FromHash cannot be empty")
+	errParentsToGetExceedsFromHeight = errors.New("ParentsToGet cannot exceed FromHeight")
+)
+
+type Config struct {
+	ChainDB      ethdb.Database
+	Client       statesyncclient.Client
+	FromHash     common.Hash
+	FromHeight   uint64
+	ParentsToGet uint64
```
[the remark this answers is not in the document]
alarso16: You're right, I was also confused making this (requires reading the implementation)

sync/blocksync/syncer.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// sync fetches (up to) [ParentsToGet] blocks from peers
+// using [Client] and writes them to disk.
+// the process begins with [FromHash] and it fetches parents recursively.
+// fetching starts from the first ancestor not found on disk
+//
+// TODO: we will always fetch all the blocks, even if some blocks were
+// available on disk. This can happen if only the most recent block is
+// not available.
```
StephenButtolph, line 100: The bracket notation doesn't work here, as the variables aren't fully qualified.
```suggestion
// sync fetches (up to) ParentsToGet blocks from peers and writes them to disk.
// The process begins with FromHash and it fetches parents recursively.
// Fetching starts from the first ancestor not found on disk.
//
// TODO: we will always fetch all the blocks, even if some blocks were available
// on disk. This can happen if only the most recent block is not available.
```

sync/blocksync/syncer.go
alarso16: I think it's a logical constraint, and the old code does handle it. I think maybe the interesting point you're making here is what if it's a small chain? Probably the consumer doesn't want to worry about whether the minimum 256 blocks (or whatever the number is) to be achieved. I'll remove it

StephenButtolph: Yeah I'm interested in the small chain case.

fwiw I don't think we would ever perform state sync in that case, but it seems odd for this code to make that assumption.

sync/blocksync/syncer_test.go
```diff
@@ -0,0 +1,319 @@
...
+
+// newTestEnvironment creates a new test environment with generated blocks
+func newTestEnvironment(t *testing.T, numBlocks int) *testEnvironment {
+	t.Helper()
+
+	var (
+		key1, _        = crypto.HexToECDSA("b71c71a67e1177ad4e901695e1b4b9ee17ae16c6668d313eac2f96dbcda3f291")
+		key2, _        = crypto.HexToECDSA("8a1f9a8f95be41cd7ccb6168179afb4504aefe388d1e14474d32c45c72ce7b7a")
+		addr1          = crypto.PubkeyToAddress(key1.PublicKey)
+		addr2          = crypto.PubkeyToAddress(key2.PublicKey)
```
StephenButtolph, line 214: `key2` and `addr2` are really not needed. Any place that uses `addr2` could just use `addr1`.

sync/blocksync/syncer.go
```diff
@@ -0,0 +1,145 @@
...
+	cancel context.CancelFunc
+}
+
+func NewSyncer(config *Config) (*blockSyncer, error) {
+	if err := config.Validate(); err != nil {
+		return nil, err
+	}
+	return &blockSyncer{
+		config: config,
+		err:    make(chan error),
```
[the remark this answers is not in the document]
alarso16: Didn't think of that...

sync/blocksync/syncer_test.go
```diff
@@ -0,0 +1,319 @@
...
+		})
+	}
+}
+
+func TestBlockSyncer_ContextCancellation(t *testing.T) {
+	// Allow some time for startup for cancellation
+	env := newTestEnvironment(t, 10).withCustomBlockProvider(func(hash common.Hash, height uint64) *types.Block {
+		time.Sleep(100 * time.Millisecond)
+		return nil
+	})
```
StephenButtolph, line 184: The sleep (and custom block provider) don't seem required/relevant for this test.
```suggestion
	env := newTestEnvironment(t, 10).withDefaultBlockProvider()
```
As it is currently written, it seems like a racy test, but it actually isn't. This sleep isn't required.

sync/blocksync/syncer_test.go
```diff
@@ -0,0 +1,319 @@
...
+	batch := e.chainDB.NewBatch()
+	for _, height := range blockHeights {
+		if height <= len(e.blocks) {
+			// blocks[0] is block number 1, blocks[1] is block number 2, etc.
+			block := e.blocks[height]
+			rawdb.WriteBlock(batch, block)
+			rawdb.WriteCanonicalHash(batch, block.Hash(), block.NumberU64())
+		}
+	}
+	require.NoError(nil, batch.Write())
```
StephenButtolph, line 284: We should not be passing `nil` in to any of these functions.

We could (in preferred order)
- return the error to the caller
- pass in the testing struct
- explicitly panic

Having a nil pointer in the require library isn't ideal.

sync/blocksync/syncer_test.go
```diff
@@ -0,0 +1,319 @@
...
+		block := e.blocks[height]
+		if block.Hash() != hash {
+			return nil
+		}
+		return block
+	})
+}
+
+// prePopulateBlocks writes some blocks to the database before syncing (by block height)
+func (e *testEnvironment) prePopulateBlocks(blockHeights ...int) *testEnvironment {
```
StephenButtolph, line 274: Why does this return the `testEnvironment`? It doesn't seem like this is used as part of the constructor chaining pattern
```suggestion
func (e *testEnvironment) prePopulateBlocks(blockHeights ...int) {
```

sync/blocksync/syncer.go
alarso16: I checked, it seems handled correctly (and I have a test for this)

sync/blocksync/syncer.go
alarso16: I'm going to put it as a TODO, since this is a more invasive change than the other naming things. It seems like this should be addressed when fixing the decisions about which blocks will be requested

sync/vm/client.go
alarso16: It's actually a constant anyway, so I removed this as parameter to the function.

sync/blocksync/syncer.go
alarso16: How do I know whether it will work (it's fixed btw)

sync/blocksync/syncer.go
```diff
@@ -0,0 +1,145 @@
...
+		// block exists
+		nextHash = blk.ParentHash()
+		nextHeight--
+		parentsToGet--
+	}
+
+	// get any blocks we couldn't find on disk from peers and write
+	// them to disk.
+	batch := s.config.ChainDB.NewBatch()
+	for i := parentsToGet - 1; i >= 0 && (nextHash != common.Hash{}); {
```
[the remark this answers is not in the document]
alarso16: It is to handle genesis - I'll add a comment

sync/blocksync/syncer.go
StephenButtolph: If you hold Command + click on the doc, vscode will take you to the declaration. Alternatively you could run a godoc server locally and see how it renders

StephenButtolph (approved):
final nits then lgtm

PR #1083, opened 2025-07-30
feat(sync): make SyncerRegistry thread-safe and execute syncers concurrently

### LINKED ISSUE #1082: Start and execute syncers concurrently
For more context: #1074 

This feature enables starting the state trie syncer and the atomic trie syncer in a concurrent fashion to optimize for sync time.

powerslider:
## Why this should be merged

This change makes the registry production-ready for concurrent environments where multiple goroutines may register syncers simultaneously, and ensures all syncers execute concurrently for better performance.

## How this works

- Replace slice-based storage with sync.Map for concurrent access.
- Use LoadOrStore for atomic duplicate prevention during registration.
- Execute all registered syncers concurrently using errgroup.
- Update tests to work with concurrent model and verify thread safety.

### Implement Cancellation-Aware Result 
- Add `synccommon.WaitResult` type to unify return values for both successful completion, clean cancellation, and real errors.
- Implement `WaitForCompletion(ctx, doneCh, cancelFunc)` helper to DRY up cancellation versus completion logic across syncers.
- Refactor `stateSyncer.Wait` and `atomicSyncer.Wait` to use the new helper and return `WaitResult` instead of raw `error`.
- Update `RunSyncerTasks` to interpret `WaitResult` semantics:
  - Treat `Cancelled=true` or `Err=context.Canceled` as clean aborts.
  - Fail fast on non-cancellation errors.
- Adjust tests to validate the three-way (`Success` / `Cancelled` / `Error`) contract exposed by `WaitResult`.

## How this was tested

Lots of tests were added to `sync/vm/registry_test.go` to cover all possible concurrency edge cases.

## Need to be documented?
no

no

resolves #1082


sync/vm/registry_test.go
```diff
@@ -113,10 +132,31 @@ func TestSyncerRegistry_Register(t *testing.T) {
...
+		},
+		{
+			name: "nil syncer registration",
+			registrations: []struct {
+				name   string
+				syncer *mockSyncer
+			}{
+				{"Syncer1", nil},
+			},
+			expectedCount: 1, // Nil syncer should be allowed
```
alarso16, line 153: Should it be allowed?

sync/vm/registry_test.go
```diff
@@ -260,7 +422,39 @@ func TestSyncerRegistry_RunSyncerTasks(t *testing.T) {
...
 
 			// Use custom assertion function for each test case.
-			tt.assertState(t, mockSyncers, tt.expectedError)
+			tt.assertState(t, mockSyncers, err)
 		})
 	}
 }
+
+// TestSyncerRegistry_ConcurrentRegistration tests that the registry handles concurrent registration safely.
+func TestSyncerRegistry_ConcurrentRegistration(t *testing.T) {
```
alarso16, line 431: I don't think `SyncerRegistry` needs to be thread-safe. All the initialization should happen on a single thread anyway, so no need for `sync.Map`

alarso16 (commented):
Didn't fully review the tests, but I'm not sure I agree with the strategy here. `SyncerRegistry` doesn't need to be thread-safe, and if one of the syncers fail, we should immediately exit

sync/vm/registry.go
```diff
@@ -17,56 +19,84 @@ type SyncerTask struct {
...
-		registeredNames: make(map[string]bool),
-	}
+	return &SyncerRegistry{}
 }
 
 // Register adds a syncer to the registry.
 // Returns an error if a syncer with the same name is already registered.
 func (r *SyncerRegistry) Register(name string, syncer synccommon.Syncer) error {
-	if r.registeredNames[name] {
+	task := SyncerTask{name, syncer}
```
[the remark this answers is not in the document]
powerslider: That's a great catch. I already fixed this.

sync/vm/registry.go
```diff
@@ -17,57 +29,106 @@ type SyncerTask struct {
...
-		return fmt.Errorf("syncer with name '%s' is already registered", name)
+	// If the syncer registry has already been started, return an error.
+	if r.started.Load() {
+		return errCannotRegisterNewSyncer
+	}
+
+	// Use reflection to check for nil because in Go, a nil concrete type is not equal to a nil interface [synccommon.Syncer].
+	// When a nil concrete type is assigned to an interface, the interface contains type information even though the value is nil.
+	// reflect.ValueOf(syncer).IsNil() properly detects nil concrete types.
+	if syncer == nil || reflect.ValueOf(syncer).IsNil() {
```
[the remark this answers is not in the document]
powerslider: You can check here for explanations: 
https://mangatmodi.medium.com/go-check-nil-interface-the-right-way-d142776edef1

sync/types.go
```diff
@@ -50,3 +63,32 @@ type Extender interface {
...
+
+	select {
+	case err := <-doneCh:
+		// Background errgroup finished.
+		return WaitResult{Err: err, Cancelled: false}
+	case <-ctx.Done():
+		// Caller cancelled: tell the syncer to shut down.
+		cancelFunc()
+		// Wait for the syncer to finish.
+		<-doneCh
```
alarso16, line 90: This isn't correct - there's a race between the context cancelling and the the syncer finished

PR #1085, opened 2025-07-31
chore(sync, plugin/evm): simplify syncer configuration 

### LINKED ISSUE #1084: Simplify configuration for different syncer implementations
Extract in `Config` structs only those constructor parameters which are tunable either internally or externally. Also provide sane defaults for them when to explicit values are set.

powerslider:
## How this works

- Remove unnecessary validation on all parameters for syncers.
- All `Config` structs for the syncers now contain only tunable config option that could be internal or user facing and set sane default values to them if necessary.

## How this was tested

Modified tests in `plugin/evm/atomic/sync` and `sync/statesync`.

## Need to be documented?
no

no

resolves #1084


alarso16 (changes requested):
Validate is never called...

sync/statesync/sync_test.go
```diff
@@ -682,3 +682,160 @@ func TestStateSyncer_MultipleStart(t *testing.T) {
...
+		},
+		// Multiple field validation tests
+		{
+			name: "multiple invalid fields",
+			configModifyFunc: func(c *Config) {
+				c.Client = nil
+				c.DB = nil
+				c.BatchSize = 0
+			},
+			expectedErr: errNilClient, // Should return first error encountered
```
alarso16, line 819: nit: I don't think we need this test. It peeks into implementation of the the validation - something like reordering the checks shouldn't cause test failures imo

sync/statesync/sync_test.go
```diff
@@ -682,3 +682,160 @@ func TestStateSyncer_MultipleStart(t *testing.T) {
...
+			description:      "should reject empty root",
+		},
+		{
+			name:             "zero batch size",
+			configModifyFunc: func(c *Config) { c.BatchSize = 0 },
+			expectedErr:      errInvalidBatchSize,
+			description:      "should reject zero batch size",
+		},
+		{
+			name:             "negative batch size",
```
alarso16, line 751: We could also make these `uint`

sync/statesync/sync_test.go
```diff
@@ -682,3 +682,160 @@ func TestStateSyncer_MultipleStart(t *testing.T) {
...
+			expectedErr: errNilClient, // Should return first error encountered
+			description: "should return first validation error when multiple fields are invalid",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			// Create a copy of the valid config.
+			config := validConfig
```
alarso16, line 828: This feels race-y (prone to misuse). Maybe make an initialization function tliek `createValidConfig`

sync/statesync/sync_test.go
```diff
@@ -682,3 +682,160 @@ func TestStateSyncer_MultipleStart(t *testing.T) {
...
+
+			tt.configModifyFunc(&config)
+
+			err := config.Validate()
+
+			if tt.expectedErr != nil {
+				require.ErrorIs(t, err, tt.expectedErr, tt.description)
```
alarso16, line 835: I think you can just use `require.ErrorIs` for both cases

sync/vm/client.go
```diff
@@ -212,14 +212,8 @@ func (client *client) createBlockSyncer(ctx context.Context, fromHash common.Has
...
 func (client *client) createEVMSyncer() (synccommon.Syncer, error) {
-	return statesync.NewSyncer(&statesync.Config{
-		Client:                   client.Client,
-		Root:                     client.summary.GetBlockRoot(),
-		BatchSize:                ethdb.IdealBatchSize,
-		DB:                       client.ChainDB,
-		MaxOutstandingCodeHashes: statesync.DefaultMaxOutstandingCodeHashes,
-		NumCodeFetchingWorkers:   statesync.DefaultNumCodeFetchingWorkers,
-		RequestSize:              client.RequestSize,
+	return statesync.NewSyncer(client.Client, client.ChainDB, client.summary.GetBlockRoot(), &statesync.Config{
```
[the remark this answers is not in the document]
ceyonur: also the BatchSize

alarso16 (approved):

sync/statesync/code_syncer.go
```diff
@@ -62,19 +54,27 @@ type codeSyncer struct {
...
 // newCodeSyncer returns a code syncer that will sync code bytes from the network in a separate thread.
-func newCodeSyncer(db ethdb.Database, client statesyncclient.Client, maxOutstandingCodeHashes, numCodeFetchingWorkers int) (*codeSyncer, error) {
+func newCodeSyncer(client statesyncclient.Client, db ethdb.Database, config CodeSyncerConfig) (*codeSyncer, error) {
+	if config.MaxOutstandingCodeHashes == 0 {
+		config.MaxOutstandingCodeHashes = defaultMaxOutstandingCodeHashes
+	}
+
+	if config.NumCodeFetchingWorkers == 0 {
+		config.NumCodeFetchingWorkers = defaultNumCodeFetchingWorkers
+	}
```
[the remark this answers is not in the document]
ceyonur: we already have a `NewDefaultConfig` and we only use `MaxOutstandingCodeHashes` and `NumCodeFetchingWorkers` to pass it here, so I don't understand why we need to have 2 defaults specified in two different places.

It appears that the config itself is also a duplicate. We should either remove `CodeSyncerConfig` and use `statesync.Config` type as a parameter in this constructor or embed `CodeSyncerConfig` into the `Config`.  Currently it kind-of implies we have two different config but actually we don't.

ceyonur (approved):

PR #1099, opened 2025-08-07
feat(sync): enable thread-safe and de-duplicated syncer registration

### LINKED ISSUE #4601: Enable thread-safe syncer registration
**Design Goals**
- Coming soon: Start all registered syncers concurrently.
- Coming soon: add a method on the `Syncer` interface (e.g., `UpdateSyncTarget`) that requires iterating all syncers repeatedly while they’re running.
- Guarantee correctness and safety under concurrent registration, start, iteration, and shutdown. 

**Requirements**
- Exactly-once start: only one `RunSyncerTasks` may win - others must fail deterministically.
- Late-register rejection - no new syncers once execution begins.
- Lock-free iteration while running - constant iteration over a stable snapshot of syncers without blocking workers or risking races.
- Concurrent-safe registration before start - allow many goroutines to `Register` without data races or duplicate names (potentially needed by dynamic state sync??? - not quite sure here).
- Fast failure propagation - first syncer failure cancels others.

**Concurrency Model**
- N syncers run concurrently.
- A controller iterates all syncers to push updates while they are running.
- The registry must safely serve read-mostly iteration under write constraints (no new syncers after starting the registry).

powerslider:
## Why this should be merged

Adds crucial validations that disallow misuse of the `SyncerRegistry`.

## How this works

- Replace slice-based storage with `sync.Map` for concurrent access.
- Use `LoadOrStore` for atomic duplicate prevention during registration.
- Add validation that you can call `RunSyncerTasks` only once.
- Update tests to work with a concurrent model and verify thread safety.
- Add validation for the presence of a client summary.
- Add frozen `atomic.Value` to hold immutable `[]SyncerTask` snapshot.
- On first `RunSyncerTasks`, collect from `sync.Map` once and store in `frozen`.
  Iterate `frozen` slice instead of ranging `sync.Map` to avoid per-call copying.
- This is mostly prep work for when introducing `UpdateSyncTarget` in
  the `Syncer` API to support Firewood's changesets.

## How this was tested
Test cases were added in `sync/registry_test.go`.

## Need to be documented?
no

no 

resolves ava-labs/avalanchego#4601

### DIFF
--- a/sync/vm/registry.go
+++ b/sync/vm/registry.go
@@ -72,7 +72,7 @@ func (r *SyncerRegistry) Register(name string, syncer synccommon.Syncer) error {
 
 	task := SyncerTask{name, syncer}
 	if _, loaded := r.syncers.LoadOrStore(name, task); loaded {
-		return fmt.Errorf("syncer with name '%s' is already registered", name)
+		return fmt.Errorf("syncer is already registered: %s", name)
 	}
 
 	return nil
--- a/sync/vm/registry.go
+++ b/sync/vm/registry.go
@@ -20,7 +20,6 @@ var (
 	errSyncerCannotBeNil          = errors.New("syncer cannot be nil")
 	errCannotRegisterNewSyncer    = errors.New("cannot register new syncer due to sync already running")
 	errCannotRunSyncerTasksTwice  = errors.New("cannot run syncer tasks again before previous run completes")
-	errEmptySyncerName            = errors.New("syncer name cannot be empty")
 )
 
 // SyncerTask represents a single syncer with its name for identification.
@@ -56,11 +55,6 @@ func (r *SyncerRegistry) Register(name string, syncer synccommon.Syncer) error {
 	// If the syncer registry has already been started, return an error.
 	if r.started.Load() {
 		return errCannotRegisterNewSyncer
-	}
-
-	// Validate that the name is not empty.
-	if name == "" {
-		return errEmptySyncerName
 	}
 
 	// Use reflection to check for nil because in Go, a nil concrete type is not equal to a nil interface [synccommon.Syncer].
--- a/sync/vm/registry.go
+++ b/sync/vm/registry.go
@@ -16,7 +16,7 @@ import (
 )
 
 var (
-	errClientCannotProvideSummary = errors.New("client cannot provide a summary")
+	errClientCannotProvideSummary = errors.New("client must provide a summary")
 	errSyncerCannotBeNil          = errors.New("syncer cannot be nil")
 	errCannotRegisterNewSyncer    = errors.New("cannot register new syncer due to sync already running")
 	errCannotRunSyncerTasksTwice  = errors.New("cannot run syncer tasks again before previous run completes")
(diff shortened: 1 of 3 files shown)


sync/vm/registry.go
```diff
@@ -19,37 +31,70 @@ type SyncerTask struct {
...
+	return &SyncerRegistry{}
 }
 
 // Register adds a syncer to the registry.
 // Returns an error if a syncer with the same name is already registered.
 func (r *SyncerRegistry) Register(name string, syncer synccommon.Syncer) error {
-	if r.registeredNames[name] {
-		return fmt.Errorf("syncer with name '%s' is already registered", name)
+	// If the syncer registry has already been started, return an error.
+	if r.started.Load() {
```
alarso16, line 47: I've thought about this more: I still don't think it's necessary for all of this code to be thread-safe, but parts of it do during dynamic state sync, so it does seem easier to guarantee correctness is if it's all thread-safe

sync/vm/registry.go
```diff
@@ -19,37 +31,70 @@ type SyncerTask struct {
...
+		return errCannotRunSyncerTasksTwice
+	}
+
+	if client == nil || client.summary == nil {
+		return errClientCannotProvideSummary
+	}
+
+	// Collect all syncers from the map.
+	var syncers []SyncerTask
+	r.syncers.Range(func(key, value any) bool {
```
alarso16, line 84: I don't think this is a good way to store the `syncers`. Eventually, we're also going to have another function like this:
```go
func (r *SyncerRegistry) UpdateSyncTarget(ctx context.Context, block /* unsure of type */) error {
    for name, task := syncTasks {
        if block.Height() % task.interval == 0 {
            log.Info("Updating sync target", "syncer", name, "height", block.Height())
            if err := task.syncer.UpdateSyncTarget(ctx, block) {
                // handle err somehow
            }
        }
    }
    return nil
}
```
I don't think it's a good idea to have to re-construct the array every time we call this (approx. every 2 seconds for a couple hours)

sync/vm/registry.go
```diff
@@ -5,12 +5,24 @@ package vm
...
+	"reflect"
+	"sync"
+	"sync/atomic"
 
 	synccommon "github.com/ava-labs/coreth/sync"
 	"github.com/ava-labs/libevm/log"
 )
 
+var (
+	errClientCannotProvideSummary = errors.New("client cannot provide a summary")
```
alarso16, line 19:
```suggestion
	errClientCannotProvideSummary = errors.New("client must provide a summary")
```
I think this will ultimately be whatever the most recent `target` is, but a sync summary is ok for now

sync/vm/registry_test.go
```diff
@@ -6,25 +6,33 @@ package vm
...
 	startDelay     time.Duration
 	waitDelay      time.Duration
 	startError     error
 	waitError      error
-	startCalled    bool
-	waitCalled     bool
-	startCallCount int
-	waitCallCount  int
-	started        bool // Track if already started
+	startCalled    atomic.Bool
```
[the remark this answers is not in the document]
powerslider: We are not testing syncers here, this is just to test the behaviour of the registry. Making the `mockSyncer` thread safe is just because of running the tests in parallel. This has no connection to the syncers themselves.

sync/vm/registry.go
powerslider: Changes solving this problem you can check [here](c632a4dc9dfa3d7d6e76430974a3c35cd98e37b5)

ceyonur (commented):
Could you provide some more context why we need these in the PR description?

sync/vm/registry.go
```diff
@@ -19,37 +31,82 @@ type SyncerTask struct {
...
-	r.syncers = append(r.syncers, SyncerTask{name, syncer})
+	// Validate that the name is not empty.
+	if name == "" {
+		return errEmptySyncerName
+	}
+
+	// Use reflection to check for nil because in Go, a nil concrete type is not equal to a nil interface [synccommon.Syncer].
+	// When a nil concrete type is assigned to an interface, the interface contains type information even though the value is nil.
+	// reflect.ValueOf(syncer).IsNil() properly detects nil concrete types.
+	if syncer == nil || reflect.ValueOf(syncer).IsNil() {
```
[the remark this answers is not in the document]
powerslider: If I remove the `reflect` check, then I have to remove the entire check, because it will work for only some cases.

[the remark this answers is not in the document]
powerslider: I don't mind to remove it, but personally I would be more defensive especially in the registry, because we start counting way too much that externally the caller will validate everything properly for a component that should act as a black box to its consumer.

PR #1103, opened 2025-08-07
Do not Discard ErrMempoolFull transactions

StephenButtolph:
## Why this should be merged

We should only be discarding transactions that are invalid. If the mempool happens to be full, we should not mark a transaction as invalid.

## How this works

Specifies a slice of errors that should not be considered Discardable.

## How this was tested

Replaced `AddLocalTx` with `Add` to include `Discarded` checks.

## Need to be documented?

No.

No.

### DIFF
--- a/plugin/evm/atomic/txpool/mempool.go
+++ b/plugin/evm/atomic/txpool/mempool.go
@@ -72,6 +72,8 @@ func (m *Mempool) AddRemoteTx(tx *atomic.Tx) error {
 	defer m.lock.Unlock()
 
 	err := m.addTx(tx, false, false)
+	// If the transaction is already in the mempool, marking it as Discarded,
+	// could unexpectedly cause the transaction to have multiple statuses.
 	if err == nil || errors.Is(err, ErrAlreadyKnown) {
 		return err
 	}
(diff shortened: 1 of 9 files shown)


alarso16 (approved):

plugin/evm/atomic/txpool/mempool.go
```diff
@@ -22,6 +22,12 @@ var (
 	ErrConflict        = errors.New("conflict present")
 	ErrInsufficientFee = errors.New("insufficient fee")
 	ErrMempoolFull     = errors.New("mempool full")
+
+	errsNotToDiscard = []error{
+		nil,
```
ceyonur, line 27: nit: I have to admit `nil` here looks a bit weird

ceyonur (approved):

plugin/evm/atomic/txpool/mempool.go
StephenButtolph: Yeah I agree.. I slightly preferred the slice over:
`err == nil || errors.Is(err, ErrAlreadyKnown) || errors.Is(err, ErrMempoolFull)`
but I'm happy to go back to just specifying each if we feel like that is cleaner.

StephenButtolph: kept the slice but removed `nil` from it :+1:

PR #1104, opened 2025-08-07
refactor: use Sync instead of Start and Wait

alarso16:
## Why this should be merged

Due to the complexity of context handling and cancellation in the `Start` and `Wait` approach, this moves all syncers to the simpler `Sync` call sufficient to sync them.

One trickier part to move is the code syncer, since it needs to start prior to handling the account trie, 

## How this works

Removes channel issues (see #1096) since starting and waiting for syncers to finish are always done together.

## How this was tested

Existing UT

## Need to be documented?

No

No

### DIFF
--- a/plugin/evm/atomic/sync/syncer.go
+++ b/plugin/evm/atomic/sync/syncer.go
@@ -198,15 +198,10 @@ func newSyncer(config *Config) (*syncer, error) {
 }
 
 // Start begins syncing the target atomic root with the configured number of worker goroutines.
-func (s *syncer) Start(ctx context.Context) error {
-	if s.cancel != nil {
-		return synccommon.ErrSyncerAlreadyStarted
-	}
-
-	ctx, s.cancel = context.WithCancel(ctx)
+func (s *syncer) Sync(ctx context.Context) error {
 	s.syncer.Start(ctx, s.numWorkers, s.onSyncFailure)
 
-	return nil
+	return <-s.syncer.Done()
 }
 
 // onLeafs is the callback for the leaf syncer, which will insert the key-value pairs into the trie.
@@ -286,23 +281,6 @@ func (s *syncer) onFinish() error {
 // the atomic trie.
 func (s *syncer) onSyncFailure(error) error {
 	return nil
-}
-
-// Wait blocks until the sync operation completes and returns any error that occurred.
-// It respects context cancellation and returns ctx.Err() if the context is cancelled.
-// This method must be called after Start() has been called.
-func (s *syncer) Wait(ctx context.Context) error {
-	if s.cancel == nil {
-		return synccommon.ErrWaitBeforeStart
-	}
-
-	select {
-	case err := <-s.syncer.Done():
-		return err
-	case <-ctx.Done():
-		s.cancel()
-		return ctx.Err()
-	}
 }
 
 type syncerLeafTask struct {
--- a/plugin/evm/atomic/sync/syncer.go
+++ b/plugin/evm/atomic/sync/syncer.go
@@ -143,9 +143,6 @@ type syncer struct {
 
 	// numWorkers is the number of worker goroutines to use for syncing
 	numWorkers int
-
-	// cancel is used to cancel the sync operation
-	cancel context.CancelFunc
 }
 
 // addZeros adds [common.HashLenth] zeros to [height] and returns the result as []byte
@@ -193,15 +190,17 @@ func newSyncer(config *Config) (*syncer, error) {
 	tasks <- &syncerLeafTask{syncer: syncer}
 	close(tasks)
 
-	syncer.syncer = syncclient.NewCallbackLeafSyncer(config.Client, tasks, config.RequestSize)
+	syncer.syncer = syncclient.NewCallbackLeafSyncer(config.Client, tasks, &syncclient.LeafSyncerConfig{
+		RequestSize: config.RequestSize,
+		NumWorkers:  config.NumWorkers,
+		OnFailure:   syncer.onSyncFailure,
+	})
 	return syncer, nil
 }
 
 // Start begins syncing the target atomic root with the configured number of worker goroutines.
 func (s *syncer) Sync(ctx context.Context) error {
-	s.syncer.Start(ctx, s.numWorkers, s.onSyncFailure)
-
-	return <-s.syncer.Done()
+	return s.syncer.Sync(ctx)
 }
 
 // onLeafs is the callback for the leaf syncer, which will insert the key-value pairs into the trie.
--- a/plugin/evm/atomic/sync/syncer.go
+++ b/plugin/evm/atomic/sync/syncer.go
@@ -192,7 +192,7 @@ func newSyncer(config *Config) (*syncer, error) {
 
 	syncer.syncer = syncclient.NewCallbackLeafSyncer(config.Client, tasks, &syncclient.LeafSyncerConfig{
 		RequestSize: config.RequestSize,
-		NumWorkers:  config.NumWorkers,
+		NumWorkers:  numWorkers,
 		OnFailure:   syncer.onSyncFailure,
 	})
 	return syncer, nil
--- a/plugin/evm/atomic/sync/syncer.go
+++ b/plugin/evm/atomic/sync/syncer.go
@@ -193,7 +193,7 @@ func newSyncer(config *Config) (*syncer, error) {
 	syncer.syncer = syncclient.NewCallbackLeafSyncer(config.Client, tasks, &syncclient.LeafSyncerConfig{
 		RequestSize: config.RequestSize,
 		NumWorkers:  numWorkers,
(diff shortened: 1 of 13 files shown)


sync/statesync/sync_test.go
```diff
@@ -93,19 +89,24 @@ func testSyncResumes(t *testing.T, steps []syncTest, stepCallback func()) {
 // waitFor waits for a result on the [result] channel to match [expected], or a timeout.
 func waitFor(t *testing.T, ctx context.Context, resultFunc func(context.Context) error, expected error, timeout time.Duration) {
```
alarso16, line 90: It didn't actually use the ctx provided for anything... I think this is better behavior. This function could probably be simplified more

sync/vm/registry_test.go
```diff
@@ -9,56 +9,33 @@ import (
...
-	waitError      error
-	startCalled    bool
-	waitCalled     bool
-	startCallCount int
-	waitCallCount  int
-	started        bool // Track if already started
+	name       string
+	syncDelay  time.Duration
+	syncError  error
+	startCount int
```
StephenButtolph, line 20: `startCount` is unused.

sync/vm/registry_test.go
```diff
@@ -9,56 +9,33 @@ import (
...
-	waitDelay      time.Duration
-	startError     error
-	waitError      error
-	startCalled    bool
-	waitCalled     bool
-	startCallCount int
-	waitCallCount  int
-	started        bool // Track if already started
+	name       string
+	syncDelay  time.Duration
```
StephenButtolph, line 18: `syncDelay` seems to be unused.

sync/statesync/code_syncer.go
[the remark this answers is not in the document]
alarso16: Done - this was a little tricky

plugin/evm/atomic/sync/syncer.go
```diff
@@ -193,20 +190,17 @@ func newSyncer(config *Config) (*syncer, error) {
 	tasks <- &syncerLeafTask{syncer: syncer}
 	close(tasks)
 
-	syncer.syncer = syncclient.NewCallbackLeafSyncer(config.Client, tasks, config.RequestSize)
+	syncer.syncer = syncclient.NewCallbackLeafSyncer(config.Client, tasks, &syncclient.LeafSyncerConfig{
```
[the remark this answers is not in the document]
StephenButtolph: I would say we are very inconsistent here.

I think the best usage of `Config`s are when there are reasonable default values. If something doesn't have a default value, then it is probably better to force the caller to provide it with the type, rather than needing to handle an incorrectly populated config in the function.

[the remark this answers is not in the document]
alarso16: Ok, I removed the config since it's only like 4 things, and we can always add it later trivially

sync/statesync/code_syncer.go
```diff
@@ -47,96 +49,67 @@ type CodeSyncerConfig struct {
...
-	lock sync.Mutex
-
-	CodeSyncerConfig
+	db                       ethdb.Database
+	client                   statesyncclient.Client
+	maxOutstandingCodeHashes int
+	numCodeFetchingWorkers   int
 
+	lock                  sync.Mutex
 	outstandingCodeHashes set.Set[ids.ID]  // Set of code hashes that we need to fetch from the network.
```
StephenButtolph, line 58: This isn't related to the PR, but this should really be
```suggestion
	outstandingCodeHashes set.Set[common.Hash]  // Set of code hashes that we need to fetch from the network.
```

sync/statesync/code_syncer.go
```diff
@@ -47,96 +49,67 @@ type CodeSyncerConfig struct {
...
-		outstandingCodeHashes: set.NewSet[ids.ID](0),
-		errChan:               make(chan error, 1),
+func newCodeSyncer(db ethdb.Database, client statesyncclient.Client, maxOutstandingCodeHashes, numCodeFetchingWorkers int) (*codeSyncer, error) {
+	syncer := &codeSyncer{
+		db:                       db,
+		client:                   client,
+		maxOutstandingCodeHashes: maxOutstandingCodeHashes,
+		numCodeFetchingWorkers:   numCodeFetchingWorkers,
+		codeHashes:               make(chan common.Hash, maxOutstandingCodeHashes),
+		outstandingCodeHashes:    set.NewSet[ids.ID](0),
```
StephenButtolph, line 73: Also not related to the PR, but there is no need to initialize this set (with a zero length). The zero value of a set is valid to use.
```suggestion
```

sync/statesync/state_syncer.go
```diff
@@ -303,16 +275,15 @@ func (t *stateSync) removeTrieInProgress(root common.Hash) (int, error) {
...
-func (t *stateSync) onSyncFailure(error) error {
+func (t *stateSync) onSyncFailure() {
 	t.lock.RLock()
 	defer t.lock.RUnlock()
 
 	for _, trie := range t.triesInProgress {
 		for _, segment := range trie.segments {
 			if err := segment.batch.Write(); err != nil {
-				return err
+				log.Error("failed to write segment batch on sync failure", "err", err)
```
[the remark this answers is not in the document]
alarso16: Not intended - I'll add an empty return. I think in any case, this error will never be handled and the state is corrupted anyway.

sync/statesync/code_syncer.go
```diff
@@ -47,101 +48,77 @@ type CodeSyncerConfig struct {
...
+	close(c.open)
+
+	return eg.Wait()
 }
 
 // Clean out any codeToFetch markers from the database that are no longer needed and
 // add any outstanding markers to the queue.
-func (c *codeSyncer) addCodeToFetchFromDBToQueue() error {
-	it := customrawdb.NewCodeToFetchIterator(c.DB)
+func addCodeToFetchFromDBToQueue(db ethdb.Database) ([]common.Hash, error) {
```
StephenButtolph, line 107:
```suggestion
func getCodeToFetchFromDB(db ethdb.Database) ([]common.Hash, error) {
```

sync/statesync/code_syncer.go
```diff
@@ -47,101 +48,77 @@ type CodeSyncerConfig struct {
...
+		return fmt.Errorf("Unable to resume previous sync: %w", err)
 	}
+	c.dbCodeHashes = nil
+	close(c.open)
+
+	return eg.Wait()
 }
 
 // Clean out any codeToFetch markers from the database that are no longer needed and
 // add any outstanding markers to the queue.
```
StephenButtolph, line 106:
```suggestion
// return any outstanding markers.
```

StephenButtolph (approved):

powerslider (approved):

sync/statesync/code_syncer.go
```diff
@@ -47,101 +48,75 @@ type CodeSyncerConfig struct {
...
 
 // newCodeSyncer returns a code syncer that will sync code bytes from the network in a separate thread.
-func newCodeSyncer(config CodeSyncerConfig) *codeSyncer {
-	return &codeSyncer{
-		CodeSyncerConfig:      config,
-		codeHashes:            make(chan common.Hash, config.MaxOutstandingCodeHashes),
-		outstandingCodeHashes: set.NewSet[ids.ID](0),
-		errChan:               make(chan error, 1),
+func newCodeSyncer(db ethdb.Database, client statesyncclient.Client, maxOutstandingCodeHashes, numCodeFetchingWorkers int) (*codeSyncer, error) {
+	dbCodeHashes, err := getCodeToFetchFromDB(db)
```
ceyonur, line 66: any reason why we do this here? I think we can do it in Sync right?

sync/statesync/code_syncer.go
```diff
@@ -47,101 +48,75 @@ type CodeSyncerConfig struct {
 // codeSyncer syncs code bytes from the network in a seprate thread.
 // Tracks outstanding requests in the DB, so that it will still fulfill them if interrupted.
 type codeSyncer struct {
-	lock sync.Mutex
-
-	CodeSyncerConfig
```
ceyonur, line 52: should we also remove the type?

sync/client/leaf_syncer.go
```diff
@@ -48,12 +52,11 @@ type LeafClient interface {
 }
 
 // NewCallbackLeafSyncer creates a new syncer object to perform leaf sync of tries.
-func NewCallbackLeafSyncer(client LeafClient, tasks <-chan LeafSyncTask, requestSize uint16) *CallbackLeafSyncer {
+func NewCallbackLeafSyncer(client LeafClient, tasks <-chan LeafSyncTask, config *LeafSyncerConfig) *CallbackLeafSyncer {
```
ceyonur, line 55: The name suggest `Callback` but it's a bit weird to take the callback in the config. Can we revert the config and take full params instead?

PR #1110, opened 2025-08-11
chore: change dependabot config

ceyonur:
## Why this should be merged

This changed dependabot to alert only for security updates

## How this works

aligns avalanchego dependabot.yml file

## How this was tested

Dependabot notifications?

## Need to be documented?

No

No


StephenButtolph (approved):
Should we do the same thing with the `github-actions` ecosystem?

PR #1112, opened 2025-08-11
Atomic vm tests

ceyonur:
## Why this should be merged

Moves atomic tests to atomic package

## How this works

* Refactors vm tests and moves helpers to `vmtest` pkg
* Moves syncervm tests to `vmtest` package to be run from different VMs
* Moves atomic tests to atomic packages

## How this was tested

UT

## Need to be documented?

No

No


plugin/evm/atomic/vm/gossiper_atomic_gossiping_test.go
```diff
@@ -20,16 +18,20 @@ import (
...
 // the txID is unknown
 func TestMempoolAtmTxsAppGossipHandling(t *testing.T) {
 	assert := assert.New(t)
 
-	tvm := newVM(t, testVMConfig{})
+	vm := newAtomicTestVM()
+	tvm := vmtest.SetupTestVM(t, vm, vmtest.TestVMConfig{})
 	defer func() {
-		assert.NoError(tvm.vm.Shutdown(context.Background()))
+		assert.NoError(vm.Shutdown(context.Background()))
```
powerslider, line 34: Don't we have the convention to use `require` instead of `assert` (especially for error checks)?

plugin/evm/atomic/vm/api.go
```diff
@@ -190,7 +190,7 @@ func (service *AvaxAPI) GetAtomicTxStatus(r *http.Request, args *api.JSONTxID, r
 		// Since chain state updates run asynchronously with VM block acceptance,
 		// avoid returning [Accepted] until the chain state reaches the block
 		// containing the atomic tx.
-		lastAccepted := service.vm.InnerVM.Blockchain().LastAcceptedBlock()
+		lastAccepted := service.vm.InnerVM.Ethereum().BlockChain().LastAcceptedBlock()
```
JonathanOppenheimer, line 193: `Ethereum().Blockchain()` is not consistently used everywhere - still use vm.Blockchain in 19 different places by my count. 

For example in `vm_test.go,` I think stuff like 

```
if ethBlk1Root := ethBlk1.Root(); !vm.blockChain.HasState(ethBlk1Root) {
		t.Fatalf("Expected blk1 state root to not yet be pruned after blk2 was accepted because of tip buffer")
	}
```
	
should be replaced `Ethereum().Blockchain().HasState()`

plugin/evm/block_builder.go
```diff
@@ -23,7 +23,7 @@ import (
 const (
```
JonathanOppenheimer, line 23: nit: for single line constants, don't need a ( ) block

plugin/evm/atomic/vm/tx_gossip_test.go
```diff
@@ -0,0 +1,312 @@
...
+	require.NoError(err)
+	address := pk.EthAddress()
+	genesis := vmtest.NewPrefundedGenesis(100_000_000_000_000_000, address)
+	genesisBytes, err := genesis.MarshalJSON()
+	require.NoError(err)
+
+	responseSender := &enginetest.SenderStub{
+		SentAppResponse: make(chan []byte, 1),
+	}
+	vm := newAtomicTestVM()
```
JonathanOppenheimer, line 61: As you introduced a new TestVM function setup, should the use of `newAtomicTestVM()` be integrated with those functions? (e.g. `setupDefaultTestVM`)

plugin/evm/atomic/vm/vm_test.go
```diff
@@ -0,0 +1,2148 @@
...
+	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/libevm/core/types"
+	"github.com/ava-labs/libevm/rlp"
+	"github.com/ava-labs/libevm/trie"
+
+	"github.com/stretchr/testify/require"
+)
+
+func newAtomicTestVM() *VM {
+	return WrapVM(&evm.VM{})
```
JonathanOppenheimer, line 53: Here is the function - you had replaced most of these with `newDefaultTestVM()`

JonathanOppenheimer (changes requested):
I don't think my review was comprehensive - too large of a PR to be confident

plugin/evm/atomic/vm/tx_gossip_test.go
```diff
@@ -0,0 +1,312 @@
...
+	"github.com/ava-labs/avalanchego/vms/secp256k1fx"
+
+	"google.golang.org/protobuf/proto"
+
+	"github.com/ava-labs/coreth/plugin/evm/atomic"
+	"github.com/ava-labs/coreth/plugin/evm/config"
+	"github.com/ava-labs/coreth/plugin/evm/vmtest"
+	"github.com/ava-labs/coreth/utils"
+)
+
```
powerslider, line 40: One improvement I can suggest is instead of just initializing the context with `context.Background()`, you can derive a context from the test deadline (the `-timeout` option in `go test`) or a sane default timeout if no deadline is set. This will avoid a situation where the test hangs in CI and you can better manage error handling via context cancellation:

```suggestion
func testContext(t *testing.T) (context.Context, context.CancelFunc) {
	t.Helper()
	if d, ok := t.Deadline(); ok {
		return context.WithDeadline(context.Background(), d)
	}
	return context.WithTimeout(context.Background(), 30*time.Second)
}
```

plugin/evm/atomic/vm/tx_gossip_test.go
```diff
@@ -0,0 +1,312 @@
...
+func TestAtomicTxGossip(t *testing.T) {
+	require := require.New(t)
+	ctx := context.Background()
```
powerslider, line 43: Use the `testContext` here:
```suggestion
ctx, cancel := testContext(t)
defer cancel()
```

plugin/evm/atomic/vm/tx_gossip_test.go
```diff
@@ -0,0 +1,312 @@
...
+		gotTx, err := marshaller.UnmarshalGossip(response.Gossip[0])
+		require.NoError(err)
+		require.Equal(tx.ID(), gotTx.GossipID())
+
+		wg.Done()
+	}
+	require.NoError(client.AppRequest(ctx, set.Of(vm.Ctx.NodeID), requestBytes, onResponse))
+	require.NoError(vm.AppRequest(ctx, requestingNodeID, 3, time.Time{}, <-peerSender.SentAppRequest))
+	require.NoError(network.AppResponse(ctx, snowCtx.NodeID, 3, <-responseSender.SentAppResponse))
+	wg.Wait()
```
powerslider, line 173: use `waitGroupContext` here:
```suggestion
	waitGroupWaitContext(t, ctx, wg)
```

plugin/evm/atomic/vm/tx_gossip_test.go
```diff
@@ -0,0 +1,312 @@
...
+		100_000_000_000,
+		pk.Address(),
+	)
+	require.NoError(err)
+	tx, err := atomic.NewImportTx(vm.Ctx, vm.CurrentRules(), vm.clock.Unix(), vm.Ctx.XChainID, address, vmtest.InitialBaseFee, secp256k1fx.NewKeychain(pk), []*avax.UTXO{utxo})
+	require.NoError(err)
+	require.NoError(vm.AtomicMempool.AddLocalTx(tx))
+
+	// wait so we aren't throttled by the vm
+	time.Sleep(5 * time.Second)
```
powerslider, line 151: use `sleepContext` here:
```suggestion
sleepContext(ctx, 5*time.Second)
```

plugin/evm/atomic/vm/tx_gossip_test.go
```diff
@@ -0,0 +1,312 @@
...
+	require.NoError(client.AppRequest(ctx, set.Of(vm.Ctx.NodeID), requestBytes, onResponse))
+	require.NoError(vm.AppRequest(ctx, requestingNodeID, 3, time.Time{}, <-peerSender.SentAppRequest))
+	require.NoError(network.AppResponse(ctx, snowCtx.NodeID, 3, <-responseSender.SentAppResponse))
+	wg.Wait()
+}
+
+// Tests that a tx is gossiped when it is issued
+func TestAtomicTxPushGossipOutbound(t *testing.T) {
+	require := require.New(t)
+	ctx := context.Background()
```
powerslider, line 179: use `testContext` here:
```suggestion
ctx, cancel := testContext(t)
defer cancel()
```

plugin/evm/atomic/vm/tx_gossip_test.go
```diff
@@ -0,0 +1,312 @@
...
+	marshaller := atomic.TxMarshaller{}
+	gossipedTx, err := marshaller.UnmarshalGossip(outboundGossipMsg.Gossip[0])
+	require.NoError(err)
+	require.Equal(tx.ID(), gossipedTx.ID())
+}
+
+// Tests that a tx is gossiped when it is issued
+func TestAtomicTxPushGossipInbound(t *testing.T) {
+	require := require.New(t)
+	ctx := context.Background()
```
powerslider, line 248: use `testContext` here:
```suggestion
ctx, cancel := testContext(t)
defer cancel()
```

plugin/evm/atomic/vm/syncervm_test.go
```diff
@@ -0,0 +1,131 @@
...
+				require.True(t, ok)
+
+				alloc := map[ids.ShortID]uint64{
+					vmtest.TestShortIDAddrs[0]: importAmount,
+				}
+
+				for addr, avaxAmount := range alloc {
+					txID, err := ids.ToID(hashing.ComputeHash256(addr.Bytes()))
+					if err != nil {
+						t.Fatalf("Failed to generate txID from addr: %s", err)
```
powerslider, line 94: Could swap `t.Fatal` for `require.NoError(t, err, "Failed to generate txID from addr")`, though I do remember that the linter sometimes preferred `t.Fatal` for some strange reason, so take this with a grain of salt.

plugin/evm/atomic/vm/export_tx_test.go
```diff
@@ -95,16 +97,16 @@ func createExportTxOptions(t *testing.T, vm *atomicvm.VM, sharedMemory *avalanch
...
 		t.Fatal(err)
 	}
 
 	// Use the funds to create 3 conflicting export transactions sending the funds to each of the test addresses
 	exportTxs := make([]*atomic.Tx, 0, 3)
 	wrappedStateDB := extstate.New(statedb)
-	for _, addr := range testShortIDAddrs {
-		exportTx, err := atomic.NewExportTx(vm.Ctx, vm.CurrentRules(), wrappedStateDB, vm.Ctx.AVAXAssetID, uint64(5000000), vm.Ctx.XChainID, addr, initialBaseFee, []*secp256k1.PrivateKey{testKeys[0]})
+	for _, addr := range vmtest.TestShortIDAddrs {
+		exportTx, err := atomic.NewExportTx(vm.Ctx, vm.CurrentRules(), wrappedStateDB, vm.Ctx.AVAXAssetID, uint64(5000000), vm.Ctx.XChainID, addr, vmtest.InitialBaseFee, []*secp256k1.PrivateKey{vmtest.TestKeys[0]})
```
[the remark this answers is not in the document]
ceyonur: Previously we have used atomic txs to fund accounts (from x chain to c-chain), now we just use the balance allocation in genesis for test keys. It means we always fund those test keys, some tests (like `TestNewExportTx`) asumes they're not already funded so it uses fresh keys.

plugin/evm/vm_test.go
```diff
@@ -17,333 +16,86 @@ import (
...
-	}}}}); err != nil {
-		return nil, err
+	if err := vm.SetExtensionConfig(exts); err != nil {
+		panic(err)
 	}
+	return vm
+}
 
-	return utxo, nil
+func setupDefaultTestVM(t *testing.T, cfg vmtest.TestVMConfig) (*VM, *vmtest.TestVMSuite) {
```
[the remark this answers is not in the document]
ceyonur: I just kept it to reduce diffs, but yea probably better to have less of these helpers.

plugin/evm/atomic/vm/api.go
ceyonur: They're part of extensible VM interface, the evm/vm has a direct access to these fields.

plugin/evm/atomic/vm/tx_gossip_test.go
ceyonur: I reverted that, I think it's better to utilize first create then setup in everywhere.

plugin/evm/atomic/vm/tx_gossip_test.go
ceyonur: We should have a linter if we want to use them instead of plain context/wg/sleep

cc @alarso16

JonathanOppenheimer (approved):

powerslider (approved):

PR #1117, opened 2025-08-13
ci: Enable extra linters

alarso16:
## Why this should be merged

This copies the avalanchego linter to use on coreth for files with only the ava-labs license

This lints the ENTIRE section of the repo, with a limited list of linters.

## How this works

Uses the same list of files as license linting creating a temporary lint file. There might be better ways to get around the creation of a new file - feel free to suggest something.

This also applies all lints that could be automatically applied. If you want to recreate the diff locally, `./scripts/lint.sh` will print the path to the created lint file, so you could apply that on master with `golangci-lint --config /path/to/config --fix`.

## How this was tested

See #1119: I changed several files to see the lint failure, using all linters

## Need to be documented?

No.

No.


.extra-golangci.yml
alarso16, line 1: This is exactly the same as the avalanchego linter

powerslider (approved):

.extra-golangci.yml
alarso16: Except several of the linters are commented out

accounts/abi/abi_test.go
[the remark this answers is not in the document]
alarso16: Done

plugin/evm/customtypes/header_ext.go
```diff
@@ -168,6 +169,8 @@ func (h *HeaderSerializable) updateToExtras(extras *HeaderExtra) {
 // HeaderSerializable defines the header of a block in the Ethereum blockchain,
 // as it is to be serialized into RLP and JSON. Note it must be exported so that
 // rlpgen can generate the serialization code from it.
+//
+//nolint:tagalign
```
alarso16, line 173: If this is linted, the generated file is different

scripts/upstream_files.txt
```diff
@@ -6,6 +6,7 @@ internal/*
 log/*
 miner/*
 node/*
+rpc/*
```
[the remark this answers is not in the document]
alarso16: Stale

PR #1118, opened 2025-08-13
Refactor atomic tx gas price calculations

StephenButtolph:
## Why this should be merged

SAE needs to calculate the effective gas price of an atomic transaction (to align with eth transactions).

## How this works

The mempool currently calculates the gas price of an atomic transaction to sort the mempool. This changes the gas price to be denominated in aAVAX/gas rather than nAVAX/gas and exposes it as a helper for other parts of the code to be able to use.

## How this was tested

- [X] Added unit tests
- [X] Existing unit tests

## Need to be documented?

No.

No.

### DIFF
--- a/plugin/evm/atomic/tx.go
+++ b/plugin/evm/atomic/tx.go
@@ -40,6 +40,7 @@ var (
 	ErrNilTx          = errors.New("tx is nil")
 	errNoValueOutput  = errors.New("output has no value")
 	ErrNoValueInput   = errors.New("input has no value")
+	ErrNoGasUsed      = errors.New("no gas used")
 	errNilOutput      = errors.New("nil output")
 	errNilInput       = errors.New("nil input")
 	errEmptyAssetID   = errors.New("empty asset ID is not valid")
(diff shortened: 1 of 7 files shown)


plugin/evm/atomic/tx.go
```diff
@@ -288,6 +289,37 @@ func SortEVMInputsAndSigners(inputs []EVMInput, signers [][]*secp256k1.PrivateKe
...
 
+// EffectiveGasPrice returns the price per gas that the transaction is paying
+// denominated in aAVAX/gas.
+//
+// The result is rounded down to the nearest aAVAX/gas.
+func EffectiveGasPrice(
+	tx UnsignedTx,
+	avaxAssetID ids.ID,
+	isApricotPhase5 bool,
+) (uint256.Int, error) {
```
StephenButtolph, line 300: We pass around `uint256.Int` rather than `*uint256.Int` to avoid accidentally overwriting values. Additionally, it often avoids a memory allocation.

When the function returns a pointer:
```
goos: darwin
goarch: arm64
pkg: github.com/ava-labs/coreth/plugin/evm/atomic
cpu: Apple M2 Max
BenchmarkEffectiveGasPrice-12    	14471852	        78.22 ns/op	      32 B/op	       1 allocs/op
PASS
coverage: 9.0% of statements in all
ok  	github.com/ava-labs/coreth/plugin/evm/atomic	2.085s
```

When it returns the struct:
```
goos: darwin
goarch: arm64
pkg: github.com/ava-labs/coreth/plugin/evm/atomic
cpu: Apple M2 Max
BenchmarkEffectiveGasPrice-12    	38252162	        31.20 ns/op	       0 B/op	       0 allocs/op
PASS
coverage: 8.9% of statements in all
ok  	github.com/ava-labs/coreth/plugin/evm/atomic	2.874s
```

plugin/evm/atomic/tx.go
```diff
@@ -288,6 +289,37 @@ func SortEVMInputsAndSigners(inputs []EVMInput, signers [][]*secp256k1.PrivateKe
...
+	gasUsed, err := tx.GasUsed(isApricotPhase5)
```
michaelkaplan13, line 301: [No action] Is there a TLDR on the historic context of why `isApricotPhase5` added a fixed gas cost to atomic txs? Purely for my own context.

PR #1122, opened 2025-08-14
fix: license header file

alarso16:
## Why this should be merged

In a comment of #1117, it was pointed out that `upstream_files.txt` is missing several files, so this is corrected.

## How this works

All files that also exist in `go-ethereum` should have the license header crediting them.

## How this was tested

Lint test

## Need to be documented?

No

No

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -1,5 +1,14 @@
 // Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
+//
+// This file is a derived work, based on the go-ethereum library whose original
+// notices appear below.
+//
+// It is distributed under a license compatible with the licensing terms of the
+// original code from which it is derived.
+//
+// Much love to the original authors for their work.
+// **********
 
 package log
 
--- a/scripts/upstream_files.txt
+++ b/scripts/upstream_files.txt
@@ -14,8 +14,8 @@ triedb/*
 !accounts/abi/bind/bind_extra_test.go
 !consensus/dummy/*
 !core/blockchain_ext.go
-!core/blockchain_log_test.go
 !core/blockchain_ext_test.go
+!core/blockchain_log_test.go
 !core/bounded_buffer.go
 !core/coretest/*
 !core/extstate/*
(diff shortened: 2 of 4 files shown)


StephenButtolph (commented):
Are the `log/` files actually derived?

StephenButtolph:
> > Are the log/ files actually derived?
> 
> Yeah they look almost exactly the same - about 20 lines difference in each file

ooooh right, those files are missing their license header in upstream.

scripts/upstream_files.txt
```diff
@@ -1,9 +1,18 @@
+accounts/*
+consensus/*
 core/*
 eth/*
+log/*
+miner/*
 node/*
 internal/*
```
StephenButtolph, line 8: These look like they are supposed to be sorted, can we make sure they are all sorted as expected?

scripts/upstream_files.txt
```diff
@@ -1,9 +1,18 @@
...
+tests/*
+triedb/*
 
-!internal/ethapi/api_extra_test.go
+!accounts/abi/abi_extra_test.go
+!accounts/abi/bind/bind_extra_test.go
+!consensus/dummy/*
 !core/blockchain_ext.go
 !core/blockchain_log_test.go
 !core/blockchain_ext_test.go
```
StephenButtolph, line 18:
```suggestion
!core/blockchain_ext_test.go
!core/blockchain_log_test.go
```

scripts/upstream_files.txt
StephenButtolph: Yeah I also like it being sorted.

log/logger.go
```diff
@@ -1,5 +1,14 @@
...
 // See the file LICENSE for licensing terms.
+//
+// This file is a derived work, based on the go-ethereum library whose original
+// notices appear below.
+//
+// It is distributed under a license compatible with the licensing terms of the
+// original code from which it is derived.
+//
+// Much love to the original authors for their work.
+// **********
```
JonathanOppenheimer, line 11: I found a whole bunch of these files in subnet-evm when I was linting them, in which there just was not an original license and I found it kinda annoying (I added the geth license, but then Cey had me remove it as for for whatever reason it wasn't there originally).

Too much for this PR but I would if there could be some beter way to denote these files

JonathanOppenheimer (approved):

PR #1127, opened 2025-08-18
feat(plugin/evm): ACP-226 - set expected block gas cost to 0 in Granite

ceyonur:
## Why this should be merged

This PR intends to remove the block gas cost by setting and expecting a `0` value. 

Related ACP: https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/226-dynamic-minimum-block-times

## How this works 

Applies with `Granite`:

* Sets `BlockGasCost` header field to 0
* Sets `EstimateRequiredTip` to be 0 
* Skips block fee verification and expect `BlockGasCost` to be 0
* Adds verification to `wrapped_block`

## How this was tested

Added vm test and UT in consensus tests

## Need to be documented?

yes, as a part of granite doc (dunno if we have one)

Maybe?

### DIFF
--- a/eth/gasprice/gasprice_test.go
+++ b/eth/gasprice/gasprice_test.go
@@ -422,11 +422,11 @@ func TestSuggestTipCapMaxBlocksLookback(t *testing.T) {
 func TestSuggestTipCapIgnoresObservedTipsPostGranite(t *testing.T) {
 	applyGasPriceTest(t, suggestTipCapTest{
 		chainConfig:     params.TestChainConfig, // Granite active in TestChainConfig
-		numBlocks:       50,
+		numBlocks:       20,
 		extDataGasUsage: common.Big0,
 		// Generate blocks with very high on-chain tips to ensure they wouldn't bias the result
 		// if the oracle looked at observed tips. Expectation remains 1 wei.
-		genBlock:    testGenBlock(t, 100_000, 200),
+		genBlock:    testGenBlock(t, 100_000, 80),
 		expectedTip: big.NewInt(1),
 	}, defaultOracleConfig())
 }


consensus/dummy/consensus.go
```diff
@@ -311,6 +313,14 @@ func (eng *DummyEngine) verifyBlockFee(
 	if requiredBlockGasCost == nil || !requiredBlockGasCost.IsUint64() {
 		return fmt.Errorf("invalid block gas cost (%d) in apricot phase 4", requiredBlockGasCost)
 	}
+	// In Granite we don't need to verify the block fee
+	if rules.IsGranite {
+		if requiredBlockGasCost.Sign() != 0 {
```
michaelkaplan13, line 318: Same comment as [here.](https://platwizards.slack.com/archives/C0758P2745C/p1755271285674579)

plugin/evm/wrapped_block.go
```diff
@@ -421,6 +429,12 @@ func (b *wrappedBlock) syntacticVerify() error {
 		}
 	}
 
+	if rulesExtra.IsGranite {
+		if customtypes.BlockGasCost(b.ethBlock).Sign() != 0 {
```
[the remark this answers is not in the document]
ceyonur: `Sign` is a bit more efficient but it does not really matter too much, if you think this is easier I'm changing that.

plugin/evm/vm_test.go
```diff
@@ -2110,3 +2110,90 @@ func TestCreateHandlers(t *testing.T) {
...
+		tx := types.NewTransaction(uint64(0), vmtest.TestEthAddrs[1], big.NewInt(1), 21000, common.Big1, nil)
+		signedTx, err := types.SignTx(tx, types.NewEIP155Signer(vm.chainConfig.ChainID), vmtest.TestKeys[0].ToECDSA())
+		require.NoError(t, err)
+		blk, err := vmtest.IssueTxsAndSetPreference([]*types.Transaction{signedTx}, vm)
+		require.NoError(t, err)
+		require.NoError(t, blk.Accept(context.Background()))
+
+		vm.clock.Set(vm.clock.Time().Add(time.Second * 1))
+		// have another tx very quickly after the first one
+		tx = types.NewTransaction(uint64(1), vmtest.TestEthAddrs[1], big.NewInt(1), 21000, common.Big1, nil)
```
michaelkaplan13, line 2149: Confirming my understanding...the reason why this `tx` has 0 priority fee per gas is because it is a legacy tx that sets the minimum possible (1 wei) gas price, right? 

It's the same exact functionally, but it may be more clear to do:
```
tx = types.NewTx(&types.DynamicFeeTx{
			ChainID: vm.chainConfig.ChainID,
			Nonce:   1,
			To:      &vmtest.TestEthAddrs[1],
			Value:   big.NewInt(1),
			Gas:     21000,
			GasFeeCap: common.Big1,
			GasTipCap: common.Big0,
		})
```
so that it's very explicit that there is no priority fee.

(Looks like `types.NewTransaction` is deprecated also either way)

consensus/dummy/consensus.go
```diff
@@ -311,6 +313,14 @@ func (eng *DummyEngine) verifyBlockFee(
 	if requiredBlockGasCost == nil || !requiredBlockGasCost.IsUint64() {
 		return fmt.Errorf("invalid block gas cost (%d) in apricot phase 4", requiredBlockGasCost)
 	}
+	// In Granite we don't need to verify the block fee
```
michaelkaplan13, line 316: It seems our verification of `BlockGasCost` and the block fee for various upgrades is somewhat inconsistent. For instance in `Finalize`, we check if we're in apricot phase 4 prior to calling `verifyBlockFee`. Would it make more sense to instead modify that check to be `if rules.IsApricotPhase4 && !rules.IsGranite`? 

https://github.com/ava-labs/coreth/blob/master/consensus/dummy/consensus.go#L393

plugin/evm/wrapped_block.go
```diff
@@ -421,6 +429,12 @@ func (b *wrappedBlock) syntacticVerify() error {
 		}
 	}
 
+	if rulesExtra.IsGranite {
```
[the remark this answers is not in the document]
ceyonur: I was under the impression we were following chronological order in upgrades and their effects here.

Yes, we should be returning early before processing transactions.

plugin/evm/vm_test.go
ceyonur: yea it was just a copy pasta from other tests.

plugin/evm/wrapped_block.go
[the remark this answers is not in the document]
michaelkaplan13: > I was under the impression we were following chronological order in upgrades and their effects here.
Gotcha, that makes sense.

Separately though, is this check necessary? IIUC, we will already verify that the BlockGasCost is 0 prior to processing any transactions without this check here, because of the logic in `verifyHeaderGasFields`. If so, this seems like an unnecessary change that can also make it less clear where the "source of truth" check occurs.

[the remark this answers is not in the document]
powerslider: I consolidated BlockGasCost checks in `plugin/evm/wrapped_block.go` under `ApricotPhase4` and assert `Granite` requires 0. Check [here](https://github.com/ava-labs/coreth/pull/1127/commits/1633e7ce7d8074be6e997d0738fc8a2d482ec2e0).

consensus/dummy/consensus.go
```diff
@@ -328,18 +329,19 @@ func (eng *DummyEngine) Finalize(chain consensus.ChainHeaderReader, block *types
 		if blockExtDataGasUsed := customtypes.BlockExtDataGasUsed(block); blockExtDataGasUsed == nil || !blockExtDataGasUsed.IsUint64() || blockExtDataGasUsed.Cmp(extDataGasUsed) != 0 {
 			return fmt.Errorf("invalid extDataGasUsed: have %d, want %d", blockExtDataGasUsed, extDataGasUsed)
 		}
+	}
 
+	if !eng.consensusMode.ModeSkipBlockFee {
```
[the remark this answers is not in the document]
powerslider: Done. Check [here](https://github.com/ava-labs/coreth/pull/1127/commits/f6279e84e17282771ec8b8612c95ee66a799be95)

plugin/evm/customheader/block_gas_cost.go
```diff
@@ -15,31 +15,37 @@ import (
...
-	errBlockGasCostNil   = errors.New("block gas cost is nil")
-	errExtDataGasUsedNil = errors.New("extDataGasUsed is nil")
-	errNoGasUsed         = errors.New("no gas used")
-
-	ErrInsufficientBlockGas = errors.New("insufficient gas to cover the block cost")
+	errBaseFeeNil                 = errors.New("base fee is nil")
+	errBlockGasCostNil            = errors.New("block gas cost is nil")
+	errExtDataGasUsedNil          = errors.New("extDataGasUsed is nil")
+	errNoGasUsed                  = errors.New("no gas used")
+	errInvalidBlockGasCostGranite = errors.New("invalid block gas cost in Granite")
```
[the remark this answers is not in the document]
powerslider: Renamed. Check [here](https://github.com/ava-labs/coreth/pull/1127/commits/7c3442ebaed588924071d407d2b0fe1dbf29cae9).

alarso16 (approved):

plugin/evm/wrapped_block.go
```diff
@@ -387,6 +396,13 @@ func (b *wrappedBlock) syntacticVerify() error {
 		case !headerExtra.BlockGasCost.IsUint64():
 			return fmt.Errorf("too large blockGasCost: %d", headerExtra.BlockGasCost)
 		}
+
+		// In Granite, block gas cost must be exactly 0.
+		if rulesExtra.IsGranite {
+			if !utils.BigEqual(headerExtra.BlockGasCost, common.Big0) {
+				return fmt.Errorf("%w: have %d, expected 0", errNonZeroBlockGasCostGranite, headerExtra.BlockGasCost)
+			}
+		}
```
StephenButtolph, line 405: This seems to be duplicate verification from what we are already doing in `consensus.go` why is this added here?

plugin/evm/wrapped_block.go
```diff
@@ -40,6 +41,14 @@ var (
 )
 
 var (
+	errInvalidBlock                  = errors.New("invalid block")
+	errInvalidNonce                  = errors.New("invalid nonce")
+	errUnclesUnsupported             = errors.New("uncles unsupported")
+	errNilBaseFeeApricotPhase3       = errors.New("nil base fee is invalid after apricotPhase3")
+	errNilBlockGasCostApricotPhase4  = errors.New("nil blockGasCost is invalid after apricotPhase4")
+	errInvalidHeaderPredicateResults = errors.New("invalid header predicate results")
+	errNonZeroBlockGasCostGranite    = errors.New("non zero block gas cost in Granite - must be 0")
```
StephenButtolph, line 50: Seems like moving these errors aren't relevant to this PR

plugin/evm/vm_test.go
```diff
@@ -2110,3 +2088,149 @@ func TestCreateHandlers(t *testing.T) {
...
+
+// TestGraniteDeactivatesBlockGasCost verifies block gas cost semantics across
+// the Fortuna -> Granite fork boundary:
+//   - Under Fortuna, attempting to build a second block too quickly should fail
+//     with ErrInsufficientBlockGas.
+//   - Under Granite, block gas cost is disabled, building succeeds and the
+//     header extra's BlockGasCost is 0.
+//
+// This guards against regressions in block gas charging behavior after Granite.
+func TestGraniteDeactivatesBlockGasCost(t *testing.T) {
```
StephenButtolph, line 2100: (perhaps an unpopular opinion) I'm not really convinced that we need these integration tests for this.

These tests are really hard to setup and pretty annoying to maintain... What coverage are we realistically getting that we aren't getting from the `block_gas_cost_test.go` file?

plugin/evm/customheader/block_gas_cost_test.go
```diff
@@ -306,7 +316,9 @@ func TestVerifyBlockFee(t *testing.T) {
 		txs                    []*types.Transaction
 		receipts               []*types.Receipt
 		extraStateContribution *big.Int
-		shouldErr              bool
+		rules                  extras.AvalancheRules
+		overrideBlockGasCost   *big.Int
```
StephenButtolph, line 320: `overrideBlockGasCost` doesn't really make sense for testing `VerifyBlockFee`.

Perhaps the name of `VerifyBlockFee` is misleading, but the function is not supposed to be verifying the `BlockGasCost`. It is verifying that the block burns a sufficient tip to cover the block fee.

Verifying that the `BlockGasCost` is correct is done here: https://github.com/ava-labs/coreth/blob/c53165428f1901f90740670aa4796f551ba8b56f/consensus/dummy/consensus.go#L177-L179

consensus/dummy/consensus.go
StephenButtolph, line 1: I don't think this file's diff is needed for this PR

plugin/evm/wrapped_block.go
powerslider: My understanding on why this is done is maybe because these paths guard different entry points:

- VM-level (wrapped_block.Verify): rejects incoming/parsing blocks with non‑zero BlockGasCost under Granite, regardless of which consensus engine is used.
- Consensus-level (consensus/dummy.Finalize): re-validates during block finalization/assembly to catch violations in locally built blocks and ensure engine-driven correctness.

StephenButtolph: Block verification flows from:
- VM level through to consensus-level.

There is never a time that one is performed but the other isn't. (Even in build block, we actually fully verify the block before returning it to the consensus engine)

eth/gasprice/gasprice_test.go
```diff
@@ -382,13 +382,29 @@ func TestSuggestGasPricePreAP3(t *testing.T) {
...
+	}{
+		// TODO: remove Fortuna case when we activate Granite
+		{
+			chainConfig: params.TestFortunaChainConfig,
+			expectedTip: big.NewInt(3),
+		},
+		{
+			chainConfig: params.TestChainConfig,
+			expectedTip: big.NewInt(1),
+		},
```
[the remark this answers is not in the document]
powerslider: I added a comment with an explanation based on what I discovered.

PR #1133, opened 2025-08-19
activate all upgrades in TestChainConfig and fix the genesis tests

ceyonur:
## Why this should be merged

`TestChainConfig` did not activate upgrades at the genesis (and in some tests never). This PR activates all upgrades in the `TestChainConfig.  

## How this works

* Fixes the corner-case in Cancun genesis activation, see comment:
```
// When Cancun is active, the block gas cost is decoded to 0 if it's nil.
// This is because BlockGasCost comes before than other optional Cancun fields in RLP order.
// This only occurs with a serialized and written genesis block, and then reading it back.
// While this does not affect anything, because we don't use `ToBlock` to retrieve the genesis block,
// it's still confusing and breaking few tests. So we set it here to 0 to make it consistent.
```

* Changes all tests accordingly to comfort `Cancun` changes in the genesis
* There are some tests that were not testing under `Cancun` and other recent upgrades, changed them as well (especially in the test data)

## How this was tested

UTs changed 

## Need to be documented?

No

No


internal/ethapi/testdata/eth_getBlockByHash-hash-1.json
ceyonur, line 1: Actually if we want to we can just change the `TestChainConfig` to an older chain config (pre-Cancun) to reduce the changes here. However I feel like we should be using latest chain config as much as possible.

params/config_test.go
```diff
@@ -125,7 +125,7 @@ func TestCheckCompatible(t *testing.T) {
 			},
 		},
 		{
-			stored:        TestChainConfig,
+			stored:        TestApricotPhase5Config,
```
ceyonur, line 128: This gives a different error because now `TestChainConfig` activates `Shanghai` and `TestApricotPhase4Config` not; and since we first check the eth upgrade errors it returns the `Shanghai` error rather than `AP5` error. If we want we can just change the expected error here as well.

powerslider (approved):

core/genesis.go
```diff
@@ -275,15 +276,34 @@ func (g *Genesis) toBlock(db ethdb.Database, triedb *triedb.Database) *types.Blo
...
 			if g.BaseFee != nil {
 				head.BaseFee = g.BaseFee
 			} else {
 				head.BaseFee = big.NewInt(ap3.InitialBaseFee)
 			}
 		}
+		headerExtra := customtypes.GetHeaderExtra(head)
+
+		// When Etna/Cancun is active, `BlockGasCost` and `ExtDataGasUsed` are decoded to 0 if it's nil.
+		// This is because these fields come before than other optional Cancun fields in RLP order.
```
alarso16, line 295: There's a typo in this line. I think this is correct:
```suggestion
		// This is because these fields come before the other optional Cancun fields in RLP order.
```

alarso16 (approved):

michaelkaplan13 (approved):

Issue #1139, opened 2025-08-22
Register the code hash syncer as a separate concurrent syncer in the registry


powerslider:
- Unify interface for concurrent start with all of the other types of syncers.
- Benefit from parallel execution.
- Not have it coupled to the EVM state syncer.

PR #1140, opened 2025-08-22
feat(sync): decouple code syncer from state syncer and run concurrently

powerslider:
## Why this should be merged

Check ava-labs/coreth#1139 

### Introduce `CodeQueue` (producer of code hashes)
- Purpose: Accept code hashes, persist durable "to-fetch" markers, and enqueue hashes for the consumer. No in-memory dedupe or local-code checks - the consumer handles dedupe/skip/retry.
- Construction: Requires a `quit <-chan struct{}` and supports functional option `WithCapacity(int)`.
- Initialization: Auto-inits on construction - recovers persisted "to-fetch" markers and enqueues them immediately.

#### Supported Operations

- `AddCode([]common.Hash) error`:
  - Persists markers for all input hashes (idempotent per hash) and enqueues them as-is.
  - Blocks if the output buffer is full until drained or `quit` triggers.
  -  Returns `errFailedToAddCodeHashesToQueue` if `quit` has already fired or fires during send.
  - Does not enforce the "no AddCode after Finalize" rule thus callers are responsible for contract adherence.
  - NOTE: Ordering across separate `AddCode` calls is not guaranteed - consumer re-batches from the stream.

- `Finalize() error`:
  - `CompareAndSwap` to finalized - if successful, waits for in-flight enqueues and closes the output channel.
  - If already quit, returns `errFailedToFinalizeCodeQueue`. If already finalized, no-op.
  - Early quit behavior: A goroutine listens on `quit`, `CompareAndSwap`s to quit, waits for in-flight enqueues, and closes the output channel to unblock consumers.

- `CodeHashes() <-chan common.Hash` exposed for consumers.

### Modifications on `CodeSyncer` (consumer of code hashes)

- Config Options: `WithNumCodeFetchingWorkers(int)`, `WithCodeHashesPerRequest(int)`.
- Deduplicate using an in-flight map before checking `rawdb.HasCode`. If code present, delete marker and release in-flight. On successful fetch, write code and delete marker in a batch, then release in-flight only after the batch commit.

### VM client wiring (`plugin/evm/vmsync`)

- `createCodeQueue()` returns `*statesync.CodeQueue` via `NewCodeQueue(chainDB, StateSyncDone)`.
- `createCodeSyncer()` passes `codeQueue.CodeHashes()` to `statesync.NewCodeSyncer`.
- `createEVMSyncer()` accepts the concrete `*statesync.CodeQueue` producer.

## How this was tested

Modified existing unit tests and add new ones for `CodeQueue`.

## Need to be documented?

no

no

resolves ava-labs/coreth#1139

### DIFF
--- a/sync/statesync/code_queue.go
+++ b/sync/statesync/code_queue.go
@@ -109,6 +109,12 @@ func (q *CodeQueue) AddCode(codeHashes []common.Hash) error {
 	if len(codeHashes) == 0 {
 		return nil
 	}
+
+	// If the queue has quit, do not attempt to send to the closed channel.
+	if q.closed.didQuit() {
+		return errFailedToAddCodeHashesToQueue
+	}
+
 	// Mark this enqueue as in-flight immediately so shutdown paths wait for us
 	// before closing the output channel.
 	q.enqueueWG.Add(1)
--- a/sync/statesync/code_syncer.go
+++ b/sync/statesync/code_syncer.go
@@ -16,13 +16,13 @@ import (
 	"github.com/ava-labs/coreth/plugin/evm/customrawdb"
 	"github.com/ava-labs/coreth/plugin/evm/message"
 
-	synccommon "github.com/ava-labs/coreth/sync"
+	syncpkg "github.com/ava-labs/coreth/sync"
 	statesyncclient "github.com/ava-labs/coreth/sync/client"
 )
 
 const defaultNumCodeFetchingWorkers = 5
 
-var _ synccommon.Syncer = (*CodeSyncer)(nil)
+var _ syncpkg.Syncer = (*CodeSyncer)(nil)
 
 // CodeSyncer syncs code bytes from the network in a separate thread.
 // It consumes code hashes from a queue and persists code into the DB.


sync/statesync/code_syncer.go
```diff
@@ -47,8 +47,9 @@ type codeSyncer struct {
 	done                  <-chan struct{}
 }
 
-// newCodeSyncer returns a code syncer that will sync code bytes from the network in a separate thread.
-func newCodeSyncer(client statesyncclient.Client, db ethdb.Database, config Config) (*codeSyncer, error) {
+// NewCodeSyncer allows external packages (e.g., registry wiring) to create a code syncer as a separate task and
+// inject it as a [CodeHashSink] into the state syncer.
+func NewCodeSyncer(client statesyncclient.Client, db ethdb.Database, config Config) (*codeSyncer, error) {
```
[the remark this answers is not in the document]
powerslider: Nope -> 	cfg := config.WithUnsetDefaults()

sync/statesync/state_syncer.go
```diff
@@ -125,12 +141,12 @@ func NewSyncer(client syncclient.Client, db ethdb.Database, root common.Hash, co
...
-	var err error
-	ss.codeSyncer, err = newCodeSyncer(client, db, cfg)
-	if err != nil {
-		return nil, err
+	if sink == nil {
+		return nil, errors.New("CodeHashSink must be provided")
 	}
+	ss.codeSink = sink
 
+	var err error
```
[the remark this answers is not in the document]
powerslider: no, it is needed

scripts/lint.sh
```diff
@@ -148,9 +148,17 @@ function test_require_error_is_no_funcs_as_params {
 }
 
 function test_require_no_error_inline_func {
-  if grep -R -zo -P '\t+err :?= ((?!require|if).|\n)*require\.NoError\((t, )?err\)' "${DEFAULT_FILES[@]}"; then
+  # Flag only when a single variable whose name contains "err" or "Err"
```
powerslider, line 151: This modification was needed in this PR, because it was wrongly matching that functions with multiple return values should also be inlined which is not possible. Now the check skips those cases.

alarso16, line 151: This is out of scope I think

sync/statesync/code_syncer.go
```diff
@@ -217,18 +224,21 @@ func (c *codeSyncer) addCode(codeHashes []common.Hash) error {
...
 // Note: this allows the worker threads to exit and return a nil error.
-func (c *codeSyncer) notifyAccountTrieCompleted() {
-	<-c.open // The code syncer must queue the previous code from the db first
+func (c *CodeSyncer) Finalize() {
+	<-c.open // The code syncer must queue the previous code from the db first.
 	close(c.codeHashes)
 }
 
+// Ready returns a channel that is closed when the code syncer is ready to accept code hashes.
+func (c *CodeSyncer) Ready() <-chan struct{} { return c.open }
```
alarso16, line 237: Why is this necessary? `AddCode` will check this first anyway

sync/types.go
```diff
@@ -39,3 +48,32 @@ type Extender interface {
 	// OnFinishAfterCommit is called after committing the sync results.
 	OnFinishAfterCommit(summaryHeight uint64) error
 }
+
+// CodeFetcher is a minimal interface for accepting discovered code hashes
+// and signaling when no more code hashes will be produced from the account trie.
+type CodeFetcher interface {
```
alarso16, line 54: I don't think this interface is necessary, since there's only one implementation. I think it adds unnecessary complexity

sync/types.go
```diff
@@ -39,3 +48,32 @@ type Extender interface {
...
+
+	// Ready returns a channel that is closed once the fetcher is ready to
+	// accept code hashes via [CodeFetcher.AddCode]. Callers can wait on this signal to avoid
+	// enqueueing work before the fetcher has initialized.
+	Ready() <-chan struct{}
+}
+
+// CodeSyncer is implemented by the concrete code syncer and combines
+// the code fetcher and syncer behaviours so callers can use a single type.
+type CodeSyncer interface {
```
alarso16, line 76: This interface too. There's only the one code syncer. If anything, this one is less necessary

alarso16 (changes requested):

sync/statesync/code_fetcher.go
```diff
@@ -117,17 +122,21 @@ func (q *CodeFetcherQueue) Init() error {
...
 	}
-	<-q.open
 
 	return q.addCode(codeHashes)
 }
 
 // Finalize implements [synccommon.CodeFetcher] by signaling no further code hashes will be added.
 func (q *CodeFetcherQueue) Finalize() {
 	<-q.open
+	q.finalized.Store(true)
```
alarso16, line 139: I'm not sure this is sufficient. What if `q.finalized` is read as false, and the `Finalize` is called, and then it attempts to add the code?

sync/statesync/code_fetcher_test.go
```diff
@@ -49,125 +47,173 @@ func TestCodeFetcherQueue_DedupesEnqueues(t *testing.T) {
...
+	codeBytes := utils.RandomBytes(20)
+	want := crypto.Keccak256Hash(codeBytes)
+	customrawdb.AddCodeToFetch(db, want)
+
+	fetcher, err := NewCodeFetcherQueue(db, make(chan struct{}), WithAutoInit(false))
+	require.NoError(t, err)
+
+	// Before Init, nothing should be readable.
+	select {
+	case <-fetcher.CodeHashes():
```
alarso16, line 63: There's nothing to read yet, right? How could this fail?

sync/statesync/code_fetcher_test.go
```diff
@@ -49,125 +47,173 @@ func TestCodeFetcherQueue_DedupesEnqueues(t *testing.T) {
...
+	// Prepare a hash that will be added via AddCode (which should block pre-Init).
+	codeBytes := utils.RandomBytes(10)
+	want := crypto.Keccak256Hash(codeBytes)
+
+	fetcher, err := NewCodeFetcherQueue(db, make(chan struct{}), WithAutoInit(false))
+	require.NoError(t, err)
+
+	// Before Init, nothing should be readable.
+	select {
+	case <-fetcher.CodeHashes():
```
alarso16, line 91: Same here

sync/statesync/code_fetcher_test.go
```diff
@@ -49,125 +47,173 @@ func TestCodeFetcherQueue_DedupesEnqueues(t *testing.T) {
...
-			got := <-fetcher.CodeHashes()
-			require.Equal(t, want, got)
-
-			fetcher.Finalize()
-			_, ok := <-fetcher.CodeHashes()
-			require.False(t, ok)
-		})
+	// Ensure AddCode blocks pre-Init.
+	addedCh := make(chan error, 1)
+	go func() { addedCh <- fetcher.AddCode([]common.Hash{want}) }()
```
alarso16, line 98: this is racy. It will probably run after the `select` statement finishes anyway

sync/statesync/code_fetcher.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	return q.enqueue(selected)
+}
+
+func (q *CodeFetcherQueue) enqueue(codeHashes []common.Hash) error {
+	for _, codeHash := range codeHashes {
+		select {
+		case q.codeHashes <- codeHash:
+		case <-q.done:
```
alarso16, line 178: Should this be first? Because if it's done, we don't want to enqueue anything

alarso16: Actually, what is this even for? How is it used?

sync/statesync/code_fetcher_test.go
powerslider: I will add a signal channel to avoid the race.

sync/statesync/code_fetcher.go
powerslider: I will add a `WaitGroup` to track in-flight `AddCode` calls to coordinate clean shutdown.

sync/statesync/code_fetcher.go
```diff
@@ -0,0 +1,239 @@
...
+	return q.addCode(codeHashes)
+}
+
+// Finalize implements [synccommon.CodeFetcher] by signaling no further code hashes will be added.
+func (q *CodeFetcherQueue) Finalize() {
+	<-q.open
+	q.finalized.Store(true)
+	// Wait for any in-flight AddCode calls to complete before closing the channel.
+	q.addWG.Wait()
+	close(q.codeHashes)
```
ARR4N, line 148: I'm not convinced that this protects against a race condition that will result in a send-after-close on this channel. At first I thought that I should walk through different scenarios, to prove to myself one way or the other, but that doesn't really solve the problem of the code being "obviously correct".

The golden rule for channel closing is that it must be done by either the sole sender, or the coordinator of the group of senders. Here we're in the latter scenario, but it's the coordination that I'm unsure about. I'll DM you to discuss.

scripts/lint.sh
```diff
@@ -148,9 +148,17 @@ function test_require_error_is_no_funcs_as_params {
...
+  # (e.g., err, myErr, parseError) is assigned from a call (:= or =), and later
+  # that same variable is passed to require.NoError(...). We explicitly require
+  # no commas on the LHS to avoid flagging multi-return assignments like
+  # "val, err := f()" or "err, val := f()".
+  #
+  # Capture the variable name and enforce it matches in the subsequent require.NoError.
+  local -r pattern='^\s*([A-Za-z_][A-Za-z0-9_]*[eE]rr[A-Za-z0-9_]*)\s*:?=\s*[^,\n]*\([^)]*\).*\n(?:(?!^\s*(?:if|require)).*\n)*^\s*require\.NoError\((?:t,\s*)?\1\)'
```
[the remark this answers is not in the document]
powerslider: I completely agree with this one. The reason I had to fix it in this PR was because it matched a wrong case. I also don't like that we have that in a script, so it is worth changing this approach in other specialized PRs.

sync/statesync/code_fetcher.go
powerslider: I'll give it a shot at explaining my understanding of the flow:

- `Finalize` is the coordinator - it sets `finalized=true`, then waits for all in-flight senders via `addWG.Wait()`, then closes `codeHashes`.

- Each `AddCode` "joins" the sender group with `addWG.Add(1)` before it can enqueue, and only "leaves" with the deferred `addWG.Done()` after its enqueues finish. That makes all sends happen-before the `Wait` completes.

- New `AddCode` calls after `finalized=true` are rejected (`errAddCodeAfterFinalize`) and still counted in `addWG` to avoid the narrow race where `Wait` begins just as `AddCode` arrives.

- Channel close happens only after all possible senders are either rejected or finished sending meaning there is no send-after-close path.

sync/statesync/code_fetcher.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+func (q *CodeFetcherQueue) addCode(codeHashes []common.Hash) error {
+	if len(codeHashes) == 0 {
+		return nil
+	}
+	batch := q.db.NewBatch()
+
+	q.lock.Lock()
```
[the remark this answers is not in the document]
powerslider: Actually this is a great idea of an improvement.

sync/statesync/code_fetcher.go
```diff
@@ -0,0 +1,239 @@
...
+		autoInit:       defaultAutoInit,
+		maxOutstanding: defaultMaxOutstandingCodeHashes,
+	}
+	for _, opt := range opts {
+		opt(&o)
+	}
+
+	dbCodeHashes, err := getCodeToFetchFromDB(db)
+	if err != nil {
+		return nil, fmt.Errorf("failed to add code hashes to queue: %w", err)
```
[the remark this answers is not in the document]
powerslider: I agree, but we have this linter configured that disallows this change:
```
/go/src/github.com/ava-labs/coreth/sync/statesync/code_fetcher.go:102:68: non-wrapping format verb for fmt.Errorf. Use `%w` to format errors (errorlint)
		return nil, fmt.Errorf("failed to add code hashes to queue: %v", err)
		                                                                                                           ^
```

sync/statesync/code_fetcher.go
```diff
@@ -0,0 +1,239 @@
...
+	// Any persisted code markers found on startup to re-enqueue when initialized.
+	dbCodeHashes []common.Hash
+}
+
+type codeFetcherOptions struct {
+	autoInit       bool
+	maxOutstanding int
+}
+
+type CodeFetcherOption func(*codeFetcherOptions)
```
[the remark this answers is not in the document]
powerslider: Let me know if you want me to change this once this PR gets merged https://github.com/ava-labs/libevm/pull/225

sync/statesync/code_syncer_test.go
```diff
@@ -54,25 +54,34 @@ func testCodeSyncer(t *testing.T, test codeSyncerTest) {
 		clientDB = rawdb.NewMemoryDatabase()
 	}
 
-	codeSyncer, err := newCodeSyncer(
+	fetcher, err := NewCodeFetcherQueue(
+		clientDB,
+		make(chan struct{}),
+		WithAutoInit(true),
```
ARR4N, line 60: When would anyone not want to auto-init? Why is there an additional method?

sync/statesync/code_fetcher.go
ARR4N, line 1: This file should be called `code_queue` or `code_fetcher_queue` (similarly for the test file) because it currently suggests that it defines the type responsible for fetching.

Additionally why is it a `CodeFetcherQueue` and not a `CodeSyncerQueue`?

sync/types.go
ARR4N: I agree with @alarso16. Whatever is accepting the `CodeFetcher` can just accept the concrete type, while the interface can be added i.f.f. necessary. This currently feels like premature abstraction, which results in duplication of maintenance.

sync/statesync/code_queue.go
```diff
@@ -0,0 +1,260 @@
...
+			return nil, fmt.Errorf("failed to write batch removing old code markers: %w", err)
+		}
+	}
+
+	return codeHashes, nil
+}
+
+// atomicCloseState provides a tiny typed wrapper around an atomic uint32.
+// It exposes enum-like operations to make intent explicit at call sites.
+type atomicCloseState struct {
```
ARR4N, line 239: (info) The fact that this is necessary points to significant problems in the consuming side of the queue, inherited from earlier work (i.e. not part of this PR). Fixing them is beyond the scope of the PR but I'm noting it nonetheless so we know to address it later.

(task) Please add an issue with a tech-debt label to fix the rest of this package.

sync/statesync/code_queue.go
```diff
@@ -0,0 +1,260 @@
...
+// Persists idempotent "to-fetch" markers for all inputs and enqueues them as-is.
+// Returns errAddCodeAfterFinalize after a clean finalize and errFailedToAddCodeHashesToQueue on early quit.
+func (q *CodeQueue) AddCode(codeHashes []common.Hash) error {
+	// If the queue has been closed, reject new work.
+	switch q.closed.get() {
+	case closeStateFinalized:
+		return errAddCodeAfterFinalize
+	case closeStateQuit:
+		return errFailedToAddCodeHashesToQueue
+	}
```
ARR4N, line 124:
```suggestion
```

It's not valid to call `AddCode()` after a call to `Finalized()` so the only way this can happen is if the `quit` channel is closed externally. In this case, `enqueue()` will return an error anyway and this `switch` won't guarantee avoiding it because it could race.

sync/statesync/code_queue.go
```diff
@@ -0,0 +1,260 @@
...
+	// Persist all input hashes as to-fetch markers. Consumer will dedupe and skip
+	// already-present code. Persisting all enables consumer-side retry.
+	// Note: markers are keyed by code hash, so repeated persists overwrite the same
+	// key rather than growing DB usage. The consumer deletes the marker after
+	// fulfilling the request (or when it detects code is already present).
+	selected := make([]common.Hash, 0, len(codeHashes))
+	for _, codeHash := range codeHashes {
+		customrawdb.AddCodeToFetch(batch, codeHash)
+		selected = append(selected, codeHash)
+	}
```
ARR4N, line 162:
```suggestion
	for _, codeHash := range codeHashes {
		customrawdb.AddCodeToFetch(batch, codeHash)
	}
```

With consumer-side deduplication, `selected` is an unnecessary clone of `codeHashes`. You'll also need to change the `range` below.

sync/statesync/code_queue.go
```diff
@@ -0,0 +1,260 @@
...
+// If the queue was already closed due to early quit, returns errFailedToFinalizeCodeQueue.
+func (q *CodeQueue) Finalize() error {
+	q.enqueueWG.Wait()
+	// If already closed due to quit, report early-exit error.
+	if q.closed.get() == closeStateQuit {
+		return errFailedToFinalizeCodeQueue
+	}
+	// Mark as finalized and close the channel (no-op if already finalized).
+	q.closed.markFinalizedAndClose(q.out)
+	return nil
```
ARR4N, line 191: This can race. `closed.markFinalizedAndClose()` needs to propagate the boolean from `transition()` and, if false, then you can return the error here.

plugin/evm/vmsync/client.go
```diff
@@ -169,52 +163,82 @@ func (client *client) stateSync(ctx context.Context) error {
...
+		client.summary.GetBlockRoot(),
+		fetcher,
+		statesync.NewDefaultConfig(client.RequestSize),
+	)
+}
+
+func (client *client) createCodeQueue() (syncpkg.CodeRequestQueue, error) {
+	return statesync.NewCodeQueue(
+		client.ChainDB,
+		client.StateSyncDone,
```
ARR4N, line 233: (rant) This is badly named since it's actually a quit channel. "Done" implies that everything has finished cleanly, not that things should quit early. It's the root of many of the problems we've had to work around here.

plugin/evm/vmsync/registry_test.go
```diff
@@ -67,12 +80,12 @@ func TestSyncerRegistry_Register(t *testing.T) {
...
 		},
 		{
-			name: "duplicate name registration",
+			name: "duplicate id registration",
 			registrations: []*mockSyncer{
 				newMockSyncer("Syncer1", nil),
 				newMockSyncer("Syncer1", nil),
 			},
-			expectedError: "syncer with name 'Syncer1' is already registered",
+			expectedError: "syncer with id 'Syncer1' is already registered",
```
ARR4N, line 88: (fyi) This is exactly what [this post is about](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html).

sync/statesync/code_syncer.go
```diff
@@ -165,76 +159,22 @@ func (c *codeSyncer) work(ctx context.Context) error {
...
 
-	// Hold the lock while modifying outstandingCodeHashes.
-	c.lock.Lock()
 	batch := c.db.NewBatch()
 	for i, codeHash := range codeHashes {
 		customrawdb.DeleteCodeToFetch(batch, codeHash)
-		c.outstandingCodeHashes.Remove(codeHash)
 		rawdb.WriteCode(batch, codeHash, codeByteSlices[i])
+		// Mark not in-flight after persisting.
+		c.inFlight.Delete(codeHash)
```
ARR4N, line 173: This needs to come after `batch.Write()` as if it returns error then another worker could see that the hash isn't in-flight, and then its database check races with the `Write()` so it duplicates the work.

sync/statesync/code_syncer.go
```diff
@@ -145,10 +123,26 @@ func (c *codeSyncer) work(ctx context.Context) error {
...
+				if err := batch.Write(); err != nil {
+					return fmt.Errorf("failed to write batch for stale code marker: %w", err)
+				}
+				continue
+			}
+
+			// Deduplicate in-flight code hashes across workers.
+			if _, loaded := c.inFlight.LoadOrStore(codeHash, struct{}{}); loaded {
+				continue
+			}
```
ARR4N, line 141: This needs to be before the `rawdb.HasCode()` check otherwise it could race. Effectively, the `inFlight` map is an atomic way of taking responsibility for a single hash by first checking if it's already in the database and then fetching+writing if it's not. By placing the `LoadOrStore()` here, multiple workers can all hammer `HasCode()` with the same hash.

sync/statesync/code_syncer_test.go
```diff
@@ -55,25 +54,33 @@ func testCodeSyncer(t *testing.T, test codeSyncerTest) {
...
 	go func() {
 		for _, codeHashes := range test.codeRequestHashes {
-			if err := codeSyncer.AddCode(codeHashes); err != nil {
-				assert.ErrorIs(t, err, test.err)
+			if err := codeQueue.AddCode(codeHashes); err != nil {
+				require.ErrorIs(t, err, test.err)
 			}
 		}
-		codeSyncer.notifyAccountTrieCompleted()
+		codeQueue.Finalize()
```
ARR4N, line 76: Check the error.

ARR4N (commented):
There are a few races and some unnecessary code. I'm finishing off some recommended changes to simplify the tests and will send them to you shortly.

sync/types.go
```diff
@@ -39,3 +48,25 @@ type Extender interface {
...
+	// produced (e.g., after Finalize and draining any outstanding items).
+	CodeHashes() <-chan common.Hash
+
+	// Finalize signals that no more code hashes will be produced by the
+	// producer (e.g., after the account trie has been fully scanned). After
+	// this call, the queue should complete any outstanding work and then
+	// close the CodeHashes channel. Returns a non-nil error if the queue is
+	// shutting down.
+	Finalize() error
+}
```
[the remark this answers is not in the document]
powerslider: I need the interface, because I want to move in a follow up PR the code syncer and code queue in `plugin/evm` and I don't want the top level `sync` package to depend on `plugin/evm` given that we would move a big chunk of that code to avalanche-go soon.

sync/statesync/code_queue.go
```diff
@@ -180,13 +181,14 @@ func (q *CodeQueue) enqueue(codeHashes []common.Hash) error {
 // Waits for in-flight enqueues to complete, then closes the output channel.
 // If the queue was already closed due to early quit, returns errFailedToFinalizeCodeQueue.
 func (q *CodeQueue) Finalize() error {
-	q.enqueueWG.Wait()
-	// If already closed due to quit, report early-exit error.
-	if q.closed.get() == closeStateQuit {
+	// Attempt to transition to finalized. If already quit, report early-exit error.
+	if q.closed.get() == closeStateQuit || !q.closed.canTransitionToFinalized() {
```
ARR4N, line 185:
```suggestion
	if !q.closed.canTransitionToFinalized() {
```

The check is redundant because the CaS guarantees that it's in open state.

sync/types.go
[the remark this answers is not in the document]
powerslider: Here is the issue: https://github.com/ava-labs/coreth/issues/1258

sync/statesync/code_queue.go
ARR4N: It's unnecessary though because it's trying to manage improper use of the API contract, which is that it's invalid to call `AddCode()` after `Finalize()`. Trying to account for all the ways that consumers can do wrong only over-complicates the code here.

```go
switch q.closed.get() {
case closeStateFinalized:
    // The consumer has broken the API contract; it's not on this code to protect against the send-after-close panic.
case closeStateQuit:
    // enqueue() will return the same error anyway.
}
```

ARR4N (approved):
Approving pending deferral of `interface` introduction. At the moment it's not needed, and coupled interface+implementation code is a major problem in our codebases.

PR #1142, opened 2025-08-22
`plugin/evm/config`: updated docs for allow-unfinalized-queries

navillanueva:
## Why this should be merged

Some clients have used this flag in production environments and has caused user complaints


plugin/evm/config/config.md
```diff
@@ -395,8 +395,12 @@ Specifies the maximum amount of CPU time that can be stored for a single WS conn
 
 ### `allow-unfinalized-queries`
 
+_Boolean_
+
 Allows queries for unfinalized (not yet accepted) blocks/transactions. Defaults to `false`.
 
+**⚠️ WARNING: Always keep this set to `false` in production.** Setting to `true` causes severe performance degradation as the node must track and serve all unfinalized blocks. Only enable for testing/debugging when specifically needed.
```
yulin-dong, line 402: I would suggest to incorporate the following from Jovica into the warning

"Querying unfinalized blocks makes sense on chains that by design do have reorgs and forks and block confirmations are a necessity so you don't have to wait 10 minutes (or however long) for actual finalization before finding out about a transaction. On Avalanche, block confirmations do not exist, you get a 100% finalized block every 2s, and you do not not need any block to be built on top of a that block to be sure that the block is part of the canonical chain.

So in effect, they are asking you to worsen performance on your end for no actual benefit anywhere. That is not a good approach, and you should push back on that."

plugin/evm/config/config.md
```diff
@@ -395,7 +395,13 @@ Specifies the maximum amount of CPU time that can be stored for a single WS conn
...
 ### `allow-unfinalized-queries`
 
-Allows queries for unfinalized (not yet accepted) blocks/transactions. Defaults to `false`.
+_Boolean_
+
+Allows queries for unfinalized (not yet accepted) blocks/transactions. Defaults to `false`. 
+
+**⚠️ WARNING: Always keep this set to `false` in production.** Enabling this flag can result in a confusing/unreliable user experience. Enabling this flag should only be done when users are expected to have knowledge of how Snow* consensus finalizes blocks.
+
+Unlike chains with reorgs and forks that require block confirmations, Avalanche provides 100% finalized blocks every ~2 seconds. There is no need to wait for additional blocks to confirm finalization. Enabling this flag provides no benefit while causing confusion about block finality.
```
StephenButtolph, line 404:
```suggestion
**⚠️ WARNING: This should likely be set to `false` in production.** Enabling this flag can result in a confusing/unreliable user experience. Enabling this flag should only be done when users are expected to have knowledge of how Snow* consensus finalizes blocks.

Unlike chains with reorgs and forks that require block confirmations, Avalanche **does not** increase the confidence that a block will be finalized based on the depth of the chain. Waiting for additional blocks **does not** confirm finalization. Enabling this flag removes the guarantee that the node only exposes finalized blocks; requiring users to guess if a block is finalized.
```

StephenButtolph (commented):
I think this description is a bit more clear.

Specifically, this flag _does_ expose blocks sooner. So there _are_ benefits... but if the user doesn't understand what the benefits are (and the tradeoffs) then it is useless.

PR #1145, opened 2025-08-25
Remove EIP-4844 package in favor of libevm version

JonathanOppenheimer:
## Why this should be merged
The EIP-4844 package here is identical to the version in `libevm`. We should reduce code duplication by using the `libevm` package. This is in-line with the longer-term goals slimming down `coreth` and `subnet-evm` and merging them.


alarso16 (approved):
Seems great to me

PR #1146, opened 2025-08-25
Fix `TestBuildEthTxBlock` flake 

JonathanOppenheimer:
## Why this should be merged

**Prior behavior:** The test was flaky on the firewood scheme because it checked the state of a restarted VM while the original VM was still running. The original VM's shutdown (and associated firewood database close/compression) was deferred until the end of the test, creating a race condition where the database state was inconsistent.

**New behavior:** The test now explicitly shuts down the original VM before creating the restarted VM, ensuring the firewood database is properly closed and compressed before checking its state. Then at the end, the restarted VM is closed so both VMs are properly cleaned up. 

## How this works
I moved the `vm.Shutdown()` from a defer statement to an explicit call before creating `restartedVM`, and added cleanup for `restartedVM` at the end.

## How this was tested
Existing test

## Need to be documented?
No

No

### DIFF
--- a/plugin/evm/vm_test.go
+++ b/plugin/evm/vm_test.go
@@ -200,12 +200,6 @@ func testBuildEthTxBlock(t *testing.T, scheme string) {
 		Scheme: scheme,
 	})
 
-	defer func() {
-		if err := vm.Shutdown(context.Background()); err != nil {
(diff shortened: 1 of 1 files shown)


JonathanOppenheimer:
Superseded by #1151

PR #1148, opened 2025-08-26
Add time check for bootstrapping 

JonathanOppenheimer:
## Why this should be merged
This PR establishes a time a check in `makePrecompile` to fix historical execution bootstrapping errors. 

```
// invalidateDelegateTime is the Unix timestamp for August 2nd, 2025, midnight Eastern Time
// (August 2nd, 2025, 04:00 UTC)
const invalidateDelegateTime = 1754107200
```

## How this was tested
Existing CI

## Need to be documented?
No

## Need to update RELEASES.md?
No

PR #1149, opened 2025-08-26
Add Granite 

JonathanOppenheimer:
## Why this should be merged
This PR builds on https://github.com/ava-labs/coreth/pull/1148 and adds explicit reversion logic upon Granite activation while honoring historical execution.

## How this was tested
I added two new unit tests, `TestMakePrecompileInvalidate_Fortuna` and `TestMakePrecompileRevert_Granite`. 

## Need to be documented?
No

No

### DIFF
--- a/params/hooks_libevm.go
+++ b/params/hooks_libevm.go
@@ -143,10 +143,13 @@ func makePrecompile(contract contract.StatefulPrecompiledContract) libevm.Precom
 
 		callType := env.IncomingCallType()
 		isDissallowedCallType := callType == vm.DelegateCall || callType == vm.CallCode
-		if env.BlockTime() >= invalidateDelegateUnix {
-			if isDissallowedCallType {
-				env.InvalidateExecution(fmt.Errorf("precompile cannot be called with %s", callType))
+		if env.BlockTime() >= invalidateDelegateUnix && isDissallowedCallType {
+			if rules.IsGranite {
+				// After Granite activation, revert the transaction (include with failed receipt)
+				return nil, 0, vm.ErrExecutionReverted
 			}
+			// Before Granite activation, invalidate execution to exclude the tx from the block
+			env.InvalidateExecution(fmt.Errorf("precompile cannot be called with %s", callType))
 		}
 
 		return contract.Run(accessibleState, env.Addresses().Caller, env.Addresses().Self, input, suppliedGas, env.ReadOnly())
--- a/plugin/evm/vm_test.go
+++ b/plugin/evm/vm_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/ava-labs/avalanchego/vms/components/chain"
 	"github.com/ava-labs/libevm/common"
 	"github.com/ava-labs/libevm/core/types"
+	"github.com/ava-labs/libevm/crypto"
 	"github.com/ava-labs/libevm/log"
 	"github.com/ava-labs/libevm/trie"
 	"github.com/holiman/uint256"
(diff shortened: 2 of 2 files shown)


michaelkaplan13 (approved):

params/hooks_libevm.go
```diff
@@ -142,11 +142,14 @@ func makePrecompile(contract contract.StatefulPrecompiledContract) libevm.Precom
...
-			if isDissallowedCallType {
-				env.InvalidateExecution(fmt.Errorf("precompile cannot be called with %s", callType))
+		isDisallowedCallType := callType == vm.DelegateCall || callType == vm.CallCode
+		if env.BlockTime() >= invalidateDelegateUnix && isDisallowedCallType {
+			if rules.IsGranite {
+				// After Granite activation, revert the transaction (include with failed receipt)
+				return nil, 0, vm.ErrExecutionReverted
 			}
+			// Before Granite activation, invalidate execution to exclude the tx from the block
+			env.InvalidateExecution(fmt.Errorf("precompile cannot be called with %s", callType))
```
ceyonur, line 152: we should not be needing checking both rule activations
```suggestion
		if isDisallowedCallType {
			switch {
			case rules.IsGranite:
				// After Granite activation, revert the transaction (include with failed receipt)
				return nil, 0, vm.ErrExecutionReverted
			case env.BlockTime() >= invalidateDelegateUnix:
				// After expiration, invalidate execution to exclude the tx from the block
				env.InvalidateExecution(fmt.Errorf("precompile cannot be called with %s", callType))
			}
		}
```

plugin/evm/vm_test.go
```diff
@@ -2110,3 +2113,116 @@ func TestCreateHandlers(t *testing.T) {
...
+			require.Equal(t, uint64(1), receipts[i].Status) // 1 = success
+			break
+		}
+	}
+	require.True(t, found)
+
+	return crypto.CreateAddress(callerAddr, nonce)
+}
+
+func TestMakePrecompileRevert_Granite(t *testing.T) {
```
ceyonur, line 2155: Can we merge this test with Fortuna and have test-table like cases for:
1- granite = should revert 
2- pre-granite (fortuna) = should invalidate and return err
3- pre-granite + pre-invalidateDelegateUnix = should do be sucessfull

ARR4N (changes requested):
I can't review the test right now but wanted to give you early feedback in case it's useful. Feel free to wait for the full review before addressing comments, if you prefer.

params/hooks_libevm.go
```diff
@@ -142,9 +142,14 @@ func makePrecompile(contract contract.StatefulPrecompiledContract) libevm.Precom
 		}
 
 		callType := env.IncomingCallType()
-		isDissallowedCallType := callType == vm.DelegateCall || callType == vm.CallCode
-		if env.BlockTime() >= invalidateDelegateUnix {
-			if isDissallowedCallType {
+		isDisallowedCallType := callType == vm.DelegateCall || callType == vm.CallCode
+		if isDisallowedCallType {
+			switch {
+			case rules.IsGranite:
```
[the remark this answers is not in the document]
JonathanOppenheimer: That rule of thumb is very useful! I haven't heard something along those lines before! I'll make these changes.

plugin/evm/vm_test.go
```diff
@@ -2110,3 +2113,157 @@ func TestCreateHandlers(t *testing.T) {
...
+		wantIncluded          bool
+		wantReceiptStatus     uint64
+	}{
+		{
+			name:           "granite_should_revert",
+			fork:           upgradetest.Granite,
+			deployGasPrice: vmtest.InitialBaseFee,
+			txGasPrice:     vmtest.InitialBaseFee,
+			setTime: func(vm *VM) {
+				vm.clock.Set(time.Unix(2000000000, 0))
```
ARR4N, line 2177: What is this value?

plugin/evm/vm_test.go
```diff
@@ -2110,3 +2113,157 @@ func TestCreateHandlers(t *testing.T) {
...
+	}
+	require.True(t, found)
+
+	return crypto.CreateAddress(callerAddr, nonce)
+}
+
+func TestDelegatePrecompile_BehaviorAcrossUpgrades(t *testing.T) {
+	// Mirror params.invalidateDelegateUnix here - if exported and
+	// imported we get a circular dependency
+	const invalidateDelegateUnix = 1754107200
```
ARR4N, line 2158: There should, instead, be a common package that both import. This value is too important to risk it getting out of sync through copy and paste.

plugin/evm/vm_test.go
```diff
@@ -2110,3 +2113,157 @@ func TestCreateHandlers(t *testing.T) {
...
+			fork:           upgradetest.Fortuna,
+			deployGasPrice: big.NewInt(ap0.MinGasPrice),
+			txGasPrice:     big.NewInt(ap0.MinGasPrice),
+			setTime: func(vm *VM) {
+				// Ensure timestamp is >= cutoff by adding the necessary delta
+				now := vm.clock.Time().Unix()
+				if now < int64(invalidateDelegateUnix+1) {
+					delta := (int64(invalidateDelegateUnix+1) - now)
+					vm.clock.Set(vm.clock.Time().Add(time.Duration(delta) * time.Second))
+				}
```
ARR4N, line 2194:
```suggestion
				if now := vm.clock.Time().Unix(); now <= invalidateDelegateUnix {
					delta := time.Duration(invalidateDelegateUnix - now + 1) * time.Second
					vm.clock.Set(vm.clock.Time().Add(delta))
				}
```

Rationale:
* The `int64` casting was redundant
* `x < y + 1` is clearer as `x <= y`
* Scope variables (`now`) to signal to readers how long they'll need to remember them

plugin/evm/vm_test.go
```diff
@@ -2110,3 +2113,157 @@ func TestCreateHandlers(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			fork := tt.fork
+			vm := newDefaultTestVM()
+			vmtest.SetupTestVM(t, vm, vmtest.TestVMConfig{
+				Fork: &fork,
+			})
```
ARR4N, line 2221:
```suggestion
			vm := newDefaultTestVM()
			vmtest.SetupTestVM(t, vm, vmtest.TestVMConfig{
				Fork: &tt.fork,
			})
```

params/hooks_libevm.go
```diff
@@ -141,12 +141,12 @@ func makePrecompile(contract contract.StatefulPrecompiledContract) libevm.Precom
...
-			if isDissallowedCallType {
-				env.InvalidateExecution(fmt.Errorf("precompile cannot be called with %s", callType))
-			}
+		switch call := env.IncomingCallType(); {
+		case call != vm.DelegateCall && call != vm.CallCode: // Others always allowed
+		case rules.IsGranite:
+			return nil, 0, vm.ErrExecutionReverted
+		case env.BlockTime() >= InvalidateDelegateUnix:
+			env.InvalidateExecution(fmt.Errorf("precompile cannot be called with %s", call))
 		}
```
alarso16, line 150: This is silently handling the pre InvalidateDelegateUnix case, maybe say that in a comment

alarso16 (approved):

params/hooks_libevm.go
ARR4N: I think he means that the `default` is implicit and that you could add it but have it only include a comment without any code.

plugin/evm/vm_test.go
```diff
@@ -1749,12 +1752,14 @@ func TestBuildBlockWithInsufficientCapacity(t *testing.T) {
 		err error
 	)
 	for i := uint64(0); i < 2; i++ {
-		tx := types.NewContractCreation(
-			i,
-			big.NewInt(0),
-			acp176.MinMaxCapacity,
-			big.NewInt(ap0.MinGasPrice),
-			[]byte{0xfe}, // invalid opcode consumes all gas
+		tx := types.NewTx(
```
ARR4N, line 1755: Sorry, I didn't mean across the entire file, but we shouldn't be adding new code that relies on deprecated functions. I think you can leave the old usage as is (or change in another PR) but this is way too large a change to arbitrarily put into this PR.

plugin/evm/vm_test.go
```diff
@@ -2110,3 +2125,158 @@ func TestCreateHandlers(t *testing.T) {
...
+// deployContract deploys the provided EVM bytecode using a prefunded test account
+// and returns the created contract address. It is reusable for any contract code.
+func deployContract(t *testing.T, vm *VM, gasPrice *big.Int, code []byte) common.Address {
+	callerAddr := vmtest.TestEthAddrs[0]
+	callerKey := vmtest.TestKeys[0]
+
+	nonce := vm.txPool.Nonce(callerAddr)
+	tx := types.NewTx(
+		&types.LegacyTx{
+			Nonce:    nonce,
```
ARR4N, line 2138:
```suggestion
			Nonce:    nonce,
			To:       nil, // contract creation
```

Makes it explicit when reading the code.

PR #1150, opened 2025-08-26
fix: add a default batch limit

alarso16:
## Why this should be merged

The batch limit fields are never used from the config

## How this works

Adds to config and docs

## How this was tested

Added UT to confirm default behavior

## Need to be documented?

Done

Done


plugin/evm/config/config.md
```diff
@@ -409,6 +411,18 @@ _Integer_
...
+
+_Integer_
+
+Maximum number of requests that can be batched in an RPC call. For no limit, set to 0. If set to 0, `batch-response-max-size` must also be set to 0. Defaults to `1000`.
+
+### `batch-response-max-size`
+
+_Integer_
+
+Maximum size of response that can be returned from a batched RPC call. For no limit, set to 0. If set to 0, `batch-request-limit` must also be set to 0. Defaults to `25 MB`.
```
StephenButtolph, line 424: I feel like this doc isn't quite accurate

> `If set to 0, `batch-XXX` must also be set to 0.`

To me, that reads like the node will error if that isn't the case... In reality, we only enable the limits if both of them are non-0.

RELEASES.md
```diff
@@ -2,7 +2,11 @@
 
 ## Pending Release
 
+<<<<<<< Updated upstream
```
michaelkaplan13, line 5: The auto merge conflict messages need to be removed here

plugin/evm/config/config.md
alarso16: I think this is more clear

plugin/evm/config/config.md
```diff
@@ -409,6 +411,18 @@ _Integer_
...
+
+_Integer_
+
+Maximum number of requests that can be batched in an RPC call. For no limit, set both this parameter and `batch-response-max-size` to 0. Defaults to `1000`.
+
+### `batch-response-max-size`
+
+_Integer_
+
+Maximum size (in bytes) of response that can be returned from a batched RPC call. For no limit, set both this parameter and `batch-request-limit` to 0. Defaults to `25 MB`.
```
StephenButtolph, line 424:
```suggestion
Maximum number of requests that can be batched in an RPC call. For no limit, set either this or `batch-response-max-size` to 0. Defaults to `1000`.

### `batch-response-max-size`

_Integer_

Maximum size (in bytes) of response that can be returned from a batched RPC call. For no limit, set either this or `batch-request-limit` to 0. Defaults to `25 MB`.
```

michaelkaplan13 (approved):

ceyonur (approved):

StephenButtolph (approved):
lgtm

PR #1151, opened 2025-08-26
chore: update firewood and fix test flakes

alarso16:
## Why this should be merged

Updates firewood to fix a memory leak

## How this works

Addresses some issues with testing firewood causing flakes

## How this was tested

Existing UT

## Need to be documented?

No

No


triedb/firewood/database.go
```diff
@@ -96,7 +96,7 @@ type Database struct {
 // Any error during creation will cause the program to exit.
 func New(config *Config) *Database {
```
StephenButtolph, line 97: Why does this function `log.Crit` rather than returning an error?

triedb/firewood/database.go
```diff
@@ -96,7 +96,7 @@ type Database struct {
...
 	if config == nil {
-		config = Defaults
+		log.Crit("firewood: config must be provided")
 	}
 
 	fwConfig, err := validatePath(config)
```
StephenButtolph, line 102: Why does `validatePath` return a config?

triedb/firewood/database.go
```diff
@@ -381,6 +381,7 @@ func (db *Database) Close() error {
 	// We don't need to explicitly dereference the proposals, since they will be cleaned up
 	// within the firewood close method.
 	db.proposalMap = nil
+	db.dereference(db.proposalTree)
```
StephenButtolph, line 384: This seems to go against the documented comment a couple lines up, is this line needed?

StephenButtolph:
May have gotten ahead of myself, only noticed now that this is in draft

triedb/firewood/database.go
alarso16: libevm interfaces...

plugin/evm/vm_test.go
```diff
@@ -281,6 +275,11 @@ func testBuildEthTxBlock(t *testing.T, scheme string) {
 		t.Fatalf("Found unexpected blkID for parent of blk2")
 	}
 
+	// Close the vm and all databases
+	if err := vm.Shutdown(context.Background()); err != nil {
+		t.Fatal(err)
+	}
```
StephenButtolph, line 281: This close seems to be needed because of racy behavior with the async acceptor and firewood.

triedb/firewood/database.go
alarso16: Removed for now

michaelkaplan13 (approved):

PR #1159, opened 2025-08-27
chore: bump libevm version

JonathanOppenheimer:
## Why this should be merged
Bumps libevm version - note that this modifys the runner context for `makePrecompile`.

### DIFF
--- a/params/hooks_libevm.go
+++ b/params/hooks_libevm.go
@@ -149,7 +149,7 @@ func makePrecompile(contract contract.StatefulPrecompiledContract) libevm.Precom
 			}
 		}
 
-		return contract.Run(accessibleState, env.Addresses().Caller, env.Addresses().Self, input, suppliedGas, env.ReadOnly())
+		return contract.Run(accessibleState, env.Addresses().Raw.Caller, env.Addresses().Raw.Self, input, suppliedGas, env.ReadOnly())
 	}
 	return vm.NewStatefulPrecompile(legacy.PrecompiledStatefulContract(run).Upgrade())
 }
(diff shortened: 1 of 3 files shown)


params/hooks_libevm.go
```diff
@@ -149,7 +149,7 @@ func makePrecompile(contract contract.StatefulPrecompiledContract) libevm.Precom
 			}
 		}
 
-		return contract.Run(accessibleState, env.Addresses().Caller, env.Addresses().Self, input, suppliedGas, env.ReadOnly())
+		return contract.Run(accessibleState, env.Addresses().Raw.Caller, env.Addresses().Raw.Self, input, suppliedGas, env.ReadOnly())
```
michaelkaplan13, line 152: Moving over [this convo from another PR](https://github.com/ava-labs/coreth/pull/1044#discussion_r2304931311).

> My understanding is that:
Currently (i.e. in v1.13.4), the Caller and Self addresses passed to precompiles are the EVM semantic versions.
In the Warp precompile, the Self address is unused, and except for delegatecall or callcode call types (which are disallowed), the Raw vs EVMSemantic Caller addresses are equivalent.
Given the above, would it make more sense to pass the EVMSemantic addresses here? It shouldn't make a difference for existing precompiles, but future ones could choose to support more functionality. I realize this likely relates to the precompile interface changes, so totally fine if it needs to be deferred for now, but want to make sure what we set here is correct for the time being of course.

@ceyonur do you agree with that, or why do you think it should be `Raw`. IMO, using `EVMSemantic` makes more sense in this case, since we decided to not change it back to the old behavior and instead have delegate calls revert.

PR #1166, opened 2025-08-28
Refactor predication verification logic

StephenButtolph:
## Why this should be merged

SAE also needs to perform predicate verification, so I needed to copy+paste the code. Rather than doing that, I decided to make a helper.

Additionally, we should parallelize the verification here (since it will be on the critical path of SAE), so I added a TODO for that.

Finally there are a small number of code cleanups.

## How this works

Moves code into a helper.

## How this was tested

The current tests were pretty odd imo. They tested `CheckPredicates` with two different test vectors and the first test required setting up mocks for each iteration.

This PR replaced the prior tests for `CheckPredicates` with a single test and added a new test for `CheckBlockPredicates`.

## Need to be documented?

No.

No.


core/predicate_check.go
```diff
@@ -18,50 +18,90 @@ import (
...
+			intrinsicGas,
+		)
 	}
 
 	rulesExtra := params.GetRulesExtra(rules)
-	predicateResults := make(predicate.PrecompileResults)
 	// Short circuit early if there are no precompile predicates to verify
 	if !rulesExtra.PredicatersExist() {
-		return predicateResults, nil
+		return nil, nil
```
StephenButtolph, line 69: `predicate.PrecompileResults` is useful as a zero value. There is no need to allocate anything here.

core/predicate_check.go
```diff
@@ -18,50 +18,90 @@ import (
...
 	}
 
 	if predicateContext == nil || predicateContext.ProposerVMBlockCtx == nil {
 		return nil, ErrMissingPredicateContext
 	}
 
+	var (
+		txHash           = tx.Hash()
+		predicateResults = make(predicate.PrecompileResults, len(predicateArguments))
```
StephenButtolph, line 87: Since we deferred the allocation of the map, now we can allocate the correct size for it 🚀

core/predicate_check.go
```diff
@@ -18,50 +18,90 @@ import (
...
+// returns the results.
+//
+// Returning an error invalidates the block.
+func CheckBlockPredicates(
+	rules params.Rules,
+	predicateContext *precompileconfig.PredicateContext,
+	txs []*types.Transaction,
+) (predicate.BlockResults, error) {
+	var results predicate.BlockResults
+	// TODO: Calculate tx predicates concurrently.
```
StephenButtolph, line 31: Because the checks here are stateless, each predicate can be trivially verified concurrently.

I didn't make that change in this PR, but will in a follow-up.

core/predicate_check.go
```diff
@@ -18,50 +18,90 @@ import (
...
+// CheckBlockPredicates verifies the predicates of a block of transactions and
+// returns the results.
+//
+// Returning an error invalidates the block.
+func CheckBlockPredicates(
+	rules params.Rules,
+	predicateContext *precompileconfig.PredicateContext,
+	txs []*types.Transaction,
+) (predicate.BlockResults, error) {
+	var results predicate.BlockResults
```
StephenButtolph, line 30: Also, we could have pre-allocated this map to the number of txs. Which would ensure we never need to resize the block results... But because most txs probably don't have predicates, I felt like that might not actually be a good optimization.

core/predicate_check.go
```diff
@@ -18,50 +18,90 @@ import (
...
 		for i, predicate := range predicates {
 			if err := predicaterContract.VerifyPredicate(predicateContext, predicate); err != nil {
 				bitset.Add(i)
 			}
 		}
-		log.Debug("predicate verify", "tx", tx.Hash(), "address", address, "res", bitset.Bytes())
+		log.Debug("predicate verify",
+			"tx", txHash,
+			"address", address,
+			"res", bitset.String(),
```
StephenButtolph, line 103: I think `String()` makes more sense than `Bytes()` since this is going into the logger.

core/predicate_check.go
```diff
@@ -18,50 +18,90 @@ import (
...
 	}
 
+	var (
+		txHash           = tx.Hash()
+		predicateResults = make(predicate.PrecompileResults, len(predicateArguments))
+	)
 	for address, predicates := range predicateArguments {
-		// Since [address] is only added to [predicateArguments] when there's a valid predicate in the ruleset
-		// there's no need to check if the predicate exists here.
-		rules := params.GetRulesExtra(rules)
```
StephenButtolph, line 56: We already fetched the rules extra, we don't need to fetch it again.

core/predicate_check.go
```diff
@@ -18,50 +18,90 @@ import (
...
 	}
-	if tx.Gas() < intrinsicGas {
-		return nil, fmt.Errorf("%w for predicate verification (%d) < intrinsic gas (%d)", ErrIntrinsicGas, tx.Gas(), intrinsicGas)
+	if gas := tx.Gas(); gas < intrinsicGas {
+		return nil, fmt.Errorf("%w for predicate verification (%d) < intrinsic gas (%d)",
+			ErrIntrinsicGas,
+			gas,
+			intrinsicGas,
+		)
 	}
```
StephenButtolph, line 64: As a followup I'm going to look into if this check is actually needed. It seems dubious at best that this check is required, and it makes this function much more gross.

core/predicate_check_test.go
```diff
@@ -19,446 +21,325 @@ import (
...
-					StorageKeys: []common.Hash{
-						{1},
-					},
-				},
-			}),
-			expectedErr: ErrIntrinsicGas,
 		},
-	} {
-		t.Run(name, func(t *testing.T) {
+	}
```
JonathanOppenheimer, line 105: Would a test case for a tx with no predicates make sense? (key present with nil vs omitted)

JonathanOppenheimer (approved):
I think the biggest change in this PR is the introduction of the nil zero value. An idea for an additional is a test  that runs `CheckBlockPredicates` on a block with:
- tx with no predicates (nil results)
- tx with predicates but all valid (empty bitset) 

then call `predicateResults.Bytes()` to examine, but perhaps this is already tested adequately

core/predicate_check.go
JonathanOppenheimer: I misunderstood the change to Set that changed it to get deleted entirely - disregard.

plugin/evm/wrapped_block.go
```diff
@@ -441,15 +440,14 @@ func (b *wrappedBlock) verifyPredicates(predicateContext *precompileconfig.Predi
...
-		predicateResults.Set(tx.Hash(), results)
+	predicateResults, err := core.CheckBlockPredicates(
+		rules,
+		predicateContext,
+		b.ethBlock.Transactions(),
+	)
+	if err != nil {
+		return err
 	}
-	// TODO: document required gas constraints to ensure marshalling predicate results does not error
```
michaelkaplan13, line 452: Why is this TODO no longer relevant here?

StephenButtolph: It was documented: https://github.com/ava-labs/avalanchego/blob/master/vms/evm/predicate/results.go#L20-L32

PR #1167, opened 2025-08-28
remove pathdb

JonathanOppenheimer:


JonathanOppenheimer:
I am intentionally leaving these lines intact: 

https://github.com/ava-labs/coreth/blob/0b9a5f8e28c820879af120c2d72a516e43b59e53/plugin/evm/vm.go#L412-L415

core/state/snapshot/generate_test.go
```diff
@@ -81,7 +79,7 @@ func testGeneration(t *testing.T, scheme string) {
 
 	root, snap := helper.CommitAndGenerate() // two of which also has the same 3-slot storage trie attached.
 
-	if have, want := root, common.HexToHash("0xa819054cfef894169a5b56ccc4e5e06f14829d4a57498e8b9fb13ff21491828d"); have != want {
+	if have, want := root, common.HexToHash("0xe3712f1a226f3782caca78ca770ccc19ee000552813a9f59d479f8611db9b1fd"); have != want {
```
JonathanOppenheimer, line 82: the hashes have changed after removing the testing for the path scheme.

JonathanOppenheimer: > **What changed**
With pathdb removed, rawdb.DeleteTrieNode is called with scheme=rawdb.HashScheme. In this mode it ignores owner/path and deletes by exact node hash only.
The test hard-codes targetHash (e.g., 0xf731… for acc-2). Those constants no longer match the actual leaf hashes produced by your current trie/state encoding. Result: delete is a no-op, the trie remains intact, and snapshot generation completes (closing genPending), triggering “Snapshot generated against corrupt account trie”.
**Why this didn’t happen before**
When using PathScheme, DeleteTrieNode deletes by owner+path instead of hash. So even if a hash constant drifted, the path-based delete would still hit the intended node. Removing pathdb removed that safety net and made the test dependent on exact node hashes.

core/genesis_test.go
```diff
@@ -314,56 +309,3 @@ func newDbConfig(t *testing.T, scheme string) *triedb.Config {
...
-	}
-
-	expected := common.Hex2Bytes("22678ccc2daa04e91013ce47799973bd6c1824f37989d7cea4cbdcd79b39137f")
-	got := genesis.ToBlock().Root().Bytes()
-	if !bytes.Equal(got, expected) {
-		t.Fatalf("invalid genesis state root, expected %x, got %x", expected, got)
-	}
-
-	db := rawdb.NewMemoryDatabase()
-	triedb := triedb.NewDatabase(db, &triedb.Config{IsVerkle: true, DBOverride: pathdb.Defaults.BackendConstructor})
```
JonathanOppenheimer, line 355: no longer exists!

core/state/snapshot/generate_test.go
alarso16: I'm really confused how this hash could be different, but that's interesting. Could this be related to the flakiness we've seen on this test in the past?

core/state/snapshot/generate_test.go
```diff
@@ -437,11 +429,8 @@ func testGenerateCorruptAccountTrie(t *testing.T, scheme string) {
 
 	root := helper.Commit() // Root: 0xfa04f652e8bd3938971bf7d71c3c688574af334ca8bc20e64b01ba610ae93cad
 
-	// Delete an account trie node and ensure the generator chokes
-	targetPath := []byte{0xc}
-	targetHash := common.HexToHash("0xf73118e0254ce091588d66038744a0afae5f65a194de67cff310c683ae43329e")
-
-	rawdb.DeleteTrieNode(helper.diskdb, common.Hash{}, targetPath, targetHash, scheme)
+	// Force corruption under hash scheme by deleting the root node.
+	rawdb.DeleteTrieNode(helper.diskdb, common.Hash{}, nil, root, scheme)
```
alarso16, line 433: This is different than the original test, fwiw. We probably shouldn't change this

core/blockchain_snapshot_test.go
```diff
@@ -452,7 +452,7 @@ func TestNoCommitCrashWithNewSnapshot(t *testing.T) {
 	//
 	// Expected head block     : C4
 	// Expected snapshot disk  : C4
-	for _, scheme := range []string{rawdb.HashScheme, rawdb.PathScheme} {
+	for _, scheme := range []string{rawdb.HashScheme} {
```
alarso16, line 455: For these tests, since you're looping through one thing, would you just remove the loop?

triedb/hashdb/database.go
```diff
@@ -717,7 +717,7 @@ func (db *Database) update(root common.Hash, parent common.Hash, nodes *trienode
 // persistent database layer.
 //
 // The first return will always be 0, representing the memory stored in unbounded
-// diff layers above the dirty cache. This is only available in pathdb.
```
alarso16, line 720: You can leave this line. This is from geth

alarso16 (changes requested):
The testing changes you point out seem dubious, I'll look more into them as well.

core/state/snapshot/generate_test.go
JonathanOppenheimer: Yeah it absolutely could be - saw some very unexpected test behavior after removing pathdb.

JonathanOppenheimer:
@aaronbuchwald has expressed a preference to keep `pathdb`, and to work instead to make it functional for re-execution benchmarks. See https://github.com/ava-labs/avalanchego/issues/4233.

I will freeze this PR for now, with the hope to return to it in the future.

PR #1168, opened 2025-08-29
move block fee verification to customheader

ceyonur:
## Why this should be merged

moves block fee verification from consensus package to customheader package

## How this works

moves function and tests

## How this was tested

UT

## Need to be documented?

no

 no


core/blockchain_ext_test.go
```diff
@@ -1488,7 +1489,7 @@ func InsertChainInvalidBlockFeeTest(t *testing.T, create createFunc) {
 	})
 	require.NoError(err)
 	_, err = blockchain.InsertChain(chain)
-	require.ErrorIs(err, dummy.ErrInsufficientBlockGas)
+	require.ErrorIs(err, customheader.ErrInsufficientBlockGas)
```
michaelkaplan13, line 1492: For my own understanding, how is it that `ModeSkipBlockFee: true` above in this test, but then we are still doing performing the block fee verification? I must be missing something...

michaelkaplan13 (approved):

PR #1173, opened 2025-08-29
fix: free memory used by proposals in firewood

demosdemon:
## Why this should be merged

Previously, when the database is closing, coreth would not free the memory used by the proposals in firewood. This is relying on an internal implementation detail of firewood but it is a violation of the firewood API contract. In the [future], this will lead to memory leaks.

The Go garbage collector will be unable to free this memory because it is not owned by Go but instead the C/Rust layer.

[future]: https://github.com/ava-labs/firewood/pull/1221

## How this works

This change reuses the dereference method and calls it for all cached proposals.

## How this was tested

Existing integration tests cover this path well.

## Need to be documented?

No


alarso16 (approved):

triedb/firewood/database.go
```diff
@@ -375,10 +375,15 @@ func (db *Database) Close() error {
 	db.proposalLock.Lock()
 	defer db.proposalLock.Unlock()
 
-	// We don't need to explicitly dereference the proposals, since they will be cleaned up
-	// within the firewood close method.
+	// clear the map before calling `dereference` to prevent extra work on close
 	db.proposalMap = nil
```
JonathanOppenheimer, line 379: is `clear()` not preferred here for the map? The new comment is unclear to me

demosdemon: on reflection, setting this to nil is preferred for _this case_ because I'm specifically checking for nil instead of len == 0. Checking for len==0 would work but feels less correct since nil indicates an invalid state (we always `make` in the New function).

JonathanOppenheimer (approved):

triedb/firewood/database.go
```diff
@@ -452,8 +457,11 @@ func (db *Database) dereference(pCtx *ProposalContext) {
 	}
 	pCtx.Children = nil
 
-	// Remove the proposal from the map.
-	db.removeProposalFromMap(pCtx)
+	// the proposalMap will be nil if the database is closing so we can skip doing
+	// unnecessary work removing the proposals from the map one-by-one.
+	if db.proposalMap != nil {
+		db.removeProposalFromMap(pCtx)
+	}
```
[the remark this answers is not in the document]
demosdemon: Yeah, that's fair. This does penalize the common case by adding the additional null check.

PR #1177, opened 2025-09-01
refactor(plugin/evm): move syncer client/registry to plugin/evm/vmsync

### LINKED ISSUE #1176: Move syncer registry and sync client to `plugin/evm`
The goal here is to fix the direction of dependency inversion. Given that `plugin/evm` contains all the custom code produced by us, it makes sense for that code to be the consumer of all other packages in the project without leaking its own internals to other top level packages. 

The suggestion is to move `sync/vm` to `plugin/evm/vmsync` and state, code and block syncers which live in the `sync` package will be imported into `plugin/evm/vmsync/client.go`. The atomic syncer already lives under `plugin/evm/atomic/sync`.

This should cleanup the dependency graph and enable easier uplifting to avalanche-go.

powerslider:
## Why this should be merged

Check #1176 

## How this works

- Move `sync/vm` -> `plugin/evm/vmsync` (client, registry, tests).
- Update imports across `plugin/evm` to consume the new `vmsync` package.

## How this was tested

Same tests that test the `SyncerRegistry`.

## Need to be documented?

no

no

resolves #1176

### DIFF
--- a/consensus/dummy/consensus.go
+++ b/consensus/dummy/consensus.go
@@ -7,9 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
-	"time"
 
-	"github.com/ava-labs/avalanchego/utils/timer/mockable"
 	"github.com/ava-labs/avalanchego/vms/components/gas"
 	"github.com/ava-labs/libevm/common"
 	"github.com/ava-labs/libevm/consensus/misc/eip4844"
@@ -26,9 +24,6 @@ import (
 )
 
 var (
-	allowedFutureBlockTime = 10 * time.Second // Max time from current time allowed for blocks, before they're considered future blocks
-
-	errInvalidBlockTime                    = errors.New("timestamp less than parent's")
 	errUnclesUnsupported                   = errors.New("uncles unsupported")
 	errExtDataGasUsedNil                   = errors.New("extDataGasUsed is nil")
 	errExtDataGasUsedTooLarge              = errors.New("extDataGasUsed is not uint64")
@@ -78,7 +73,6 @@ type (
 
 	DummyEngine struct {
 		cb                  ConsensusCallbacks
-		clock               *mockable.Clock
 		consensusMode       Mode
 		desiredTargetExcess *gas.Gas
 	}
@@ -87,12 +81,10 @@ type (
 func NewDummyEngine(
 	cb ConsensusCallbacks,
 	mode Mode,
-	clock *mockable.Clock,
 	desiredTargetExcess *gas.Gas,
 ) *DummyEngine {
 	return &DummyEngine{
 		cb:                  cb,
-		clock:               clock,
 		consensusMode:       mode,
 		desiredTargetExcess: desiredTargetExcess,
 	}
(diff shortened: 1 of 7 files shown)


plugin/evm/vmsync/registry.go
```diff
@@ -46,21 +47,24 @@ func (r *SyncerRegistry) Register(name string, syncer synccommon.Syncer) error {
...
 
 // RunSyncerTasks executes all registered syncers.
-func (r *SyncerRegistry) RunSyncerTasks(ctx context.Context, client *client) error {
+// The provided summary is used only for logging to decouple from concrete client types.
+func (r *SyncerRegistry) RunSyncerTasks(ctx context.Context, summary common.Hash) error {
 	if len(r.syncers) == 0 {
 		return nil
 	}
 
+	summaryStr := summary.Hex()
```
alarso16, line 56: nit: this is really vague. I believe this corresponds with a block hash

plugin/evm/vmsync/registry_test.go
```diff
@@ -162,7 +229,10 @@ func TestSyncerRegistry_RunSyncerTasks(t *testing.T) {
 				require.NoError(t, registry.Register(syncerConfig.name, mockSyncer))
 			}
 
-			err := registry.RunSyncerTasks(context.Background(), &client{})
+			ctx, cancel := utilstest.NewTestContext(t)
```
ARR4N, line 232: `utilstest.NewTestContext` is really opaque, adding nothing more than a 30-second deadline (or the test's own one) while forcing the reader to check that the new `Context` doesn't do anything that will interfere with the test itself. I would just use `context.Background()` or specify the deadline yourself (also just one line, but completely explicit).

plugin/evm/vmsync/registry_test.go
```diff
@@ -294,5 +364,13 @@ func TestSyncerRegistry_NoSyncersRegistered(t *testing.T) {
 	ctx, cancel := utilstest.NewTestContext(t)
 	t.Cleanup(cancel)
 
-	require.NoError(t, registry.RunSyncerTasks(ctx, &client{}))
+	require.NoError(t, registry.RunSyncerTasks(ctx, getTestClientSummary(t)))
+}
+
+func getTestClientSummary(t *testing.T) message.Syncable {
```
ARR4N, line 370: (nit, readability) How about `newDummyClientSummary`? Rationale:

The `get` prefix doesn't add anything informative while the `new` prefix suggests a constructor. In other scenarios "compute", "fetch", etc. are valuable to signal to the reader what the function is doing if more expensive or remote, respectively.

The `Test` element is also implied by the fact that we're in a test so could either be left out entirely or replaced with something informative like `Dummy` to indicate that it's a stand-in.

ARR4N (changes requested):

plugin/evm/vmsync/registry.go
```diff
@@ -46,21 +48,25 @@ func (r *SyncerRegistry) Register(name string, syncer synccommon.Syncer) error {
 }
 
 // RunSyncerTasks executes all registered syncers.
-func (r *SyncerRegistry) RunSyncerTasks(ctx context.Context, client *client) error {
+// The provided summary is used only for logging to decouple from concrete client types.
+func (r *SyncerRegistry) RunSyncerTasks(ctx context.Context, summary message.Syncable) error {
```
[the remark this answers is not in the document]
powerslider: No, it's because I did not want to depend on the entire client, given that I only only need the summary.

ARR4N (approved):

Issue #1144, opened 2025-08-25
AvalancheGo Linter integration


alarso16:
The avalanchego linter is applied on all Ava Labs owned files in this repo, but not all linters can be applied. Because it previously adhered only to a weaker linter, adding all linters would cause it to fail. Here are all the linters currently unused from` .avalanche-golangci.yml`:

- [x] depguard
- [x] errcheck
- [x] forbidigo (must do depguard first, or add handling for `assert` cases)
   - [x] `t.Fatal`
   - [x] `require.Error` and `require.ErrorContains`
   - [ ] Remaining
- [x] gocritic
- [x] gosec
- [x] nilerr
- [x] prealloc
- [x] predeclared
- [x] revive
- [x] staticcheck
- [x] testifylint

Each of these changes should be reflected in coreth prior to subnet-evm to keep the repos synced in properly.

JonathanOppenheimer:
nilerr: https://github.com/ava-labs/coreth/pull/1182

JonathanOppenheimer:
predeclared: https://github.com/ava-labs/coreth/pull/1183

JonathanOppenheimer:
gosec: ## Why this should be merged: https://github.com/ava-labs/coreth/issues/1144.

PR #1182, opened 2025-09-03
style: enable nilerr linter 

JonathanOppenheimer:
## Why this should be merged

### DIFF
--- a/sync/statesync/trie_sync_tasks.go
+++ b/sync/statesync/trie_sync_tasks.go
@@ -118,7 +118,6 @@ func (s *storageTrieTask) OnStart() (bool, error) {
 	storageTrie, err := trie.New(trie.StorageTrieID(s.sync.root, s.root, firstAccount), s.sync.trieDB)
 	if err != nil {
 		return false, nil //nolint:nilerr // the storage trie does not exist, so it should be rerequested
-		return false, nil //nolint:nilerr
 	}
 
 	// If the storage trie is already on disk, we only need to populate the storage snapshot for [accountHash]
--- a/sync/statesync/trie_sync_tasks.go
+++ b/sync/statesync/trie_sync_tasks.go
@@ -116,7 +116,8 @@ func (s *storageTrieTask) OnStart() (bool, error) {
 		firstAccount = s.accounts[0]
 	}
 	storageTrie, err := trie.New(trie.StorageTrieID(s.sync.root, s.root, firstAccount), s.sync.trieDB)
-	if err != nil { // the storage trie does not exist, so it should be rerequested
+	if err != nil {
+		return false, nil //nolint:nilerr // the storage trie does not exist, so it should be rerequested
 		return false, nil //nolint:nilerr
 	}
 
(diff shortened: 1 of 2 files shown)


sync/statesync/trie_sync_tasks.go
```diff
@@ -117,7 +117,7 @@ func (s *storageTrieTask) OnStart() (bool, error) {
 	}
 	storageTrie, err := trie.New(trie.StorageTrieID(s.sync.root, s.root, firstAccount), s.sync.trieDB)
 	if err != nil {
-		return false, nil
+		return false, err
```
alarso16, line 120: This is unintuitive - this actually means that the storage trie does not exist, so it should be rerequested i believe. So probably just want to silence the linter for this one

alarso16 (dismissed):

sync/statesync/trie_sync_tasks.go
JonathanOppenheimer: Thanks!

sync/statesync/trie_sync_tasks.go
```diff
@@ -116,8 +116,8 @@ func (s *storageTrieTask) OnStart() (bool, error) {
 		firstAccount = s.accounts[0]
 	}
 	storageTrie, err := trie.New(trie.StorageTrieID(s.sync.root, s.root, firstAccount), s.sync.trieDB)
-	if err != nil {
-		return false, nil
+	if err != nil { // the storage trie does not exist, so it should be rerequested
+		return false, nil //nolint:nilerr
```
ARR4N, line 120:
```suggestion
	if err != nil {
		return false, nil //nolint:nilerr // the storage trie does not exist, so it should be rerequested
```

This keeps the rationale coupled to the `nolint` directive, and is the [recommended pattern](https://golangci-lint.run/docs/linters/false-positives/#nolint-directive:~:text=You%20may%20add%20a%20comment%20explaining%20or%20justifying%20why%20//nolint%20is%20being%20used%20on%20the%20same%20line%20as%20the%20flag%20itself%3A).

sync/statesync/trie_sync_tasks.go
```diff
@@ -116,8 +116,8 @@ func (s *storageTrieTask) OnStart() (bool, error) {
 		firstAccount = s.accounts[0]
 	}
 	storageTrie, err := trie.New(trie.StorageTrieID(s.sync.root, s.root, firstAccount), s.sync.trieDB)
-	if err != nil {
-		return false, nil
+	if err != nil { // the storage trie does not exist, so it should be rerequested
```
ARR4N, line 119:
```suggestion
	if err != nil {
		return false, nil //nolint:nilerr // the storage trie does not exist, so it should be rerequested
```

This keeps the rationale coupled to the `nolint` directive, and is the [recommended pattern](https://golangci-lint.run/docs/linters/false-positives/#nolint-directive:~:text=You%20may%20add%20a%20comment%20explaining%20or%20justifying%20why%20//nolint%20is%20being%20used%20on%20the%20same%20line%20as%20the%20flag%20itself%3A).

ARR4N (approved):
> [!NOTE]
> Approving with a requested change that doesn't require re-review. Please action it before merging.

alarso16 (approved):

PR #1184, opened 2025-09-03
style: enable gocritic linter

JonathanOppenheimer:
## Why this should be merged

### DIFF
--- a/params/config_extra.go
+++ b/params/config_extra.go
@@ -51,11 +51,11 @@ func SetEthUpgrades(c *ChainConfig) error {
 	// Because Fuji and Mainnet have already accepted the Berlin and London
 	// blocks, it is assumed that they are scheduled for activation.
 
-	switch c.ChainID != nil {
-	case AvalancheFujiChainID.Cmp(c.ChainID) == 0:
+	switch {
+	case c.ChainID != nil && AvalancheFujiChainID.Cmp(c.ChainID) == 0:
 		c.BerlinBlock = big.NewInt(184985) // https://testnet.snowtrace.io/block/184985?chainid=43113, AP2 activation block
 		c.LondonBlock = big.NewInt(805078) // https://testnet.snowtrace.io/block/805078?chainid=43113, AP3 activation block
-	case AvalancheMainnetChainID.Cmp(c.ChainID) == 0:
+	case c.ChainID != nil && AvalancheMainnetChainID.Cmp(c.ChainID) == 0:
 		c.BerlinBlock = big.NewInt(1640340) // https://snowtrace.io/block/1640340?chainid=43114, AP2 activation block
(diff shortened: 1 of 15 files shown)


plugin/evm/vm.go
```diff
@@ -1131,7 +1130,7 @@ func (vm *VM) startContinuousProfiler() {
 		return
 	}
 	vm.profiler = profiler.NewContinuous(
-		filepath.Join(vm.config.ContinuousProfilerDir),
+		vm.config.ContinuousProfilerDir,
```
alarso16, line 1133: Dos this clean the path or anything? Seems suspicious

alarso16: The correct thing to do here (I believe) is to use `filepath.Clean`, since that will result in no behavior change:
https://pkg.go.dev/path/filepath#Join
Try out an example in go.dev/play

alarso16 (approved):

triedb/firewood/database.go
```diff
@@ -584,7 +584,7 @@ func arrangeKeyValuePairs(nodes *trienode.MergedNodeSet) ([][]byte, [][]byte) {
 	}
 
 	// We need to do all storage operations first, so prefix-deletion works for accounts.
-	keys := append(storageKeys, acctKeys...)
-	values := append(storageValues, acctValues...)
-	return keys, values
+	storageKeys = append(storageKeys, acctKeys...)
+	storageValues = append(storageValues, acctValues...)
+	return storageKeys, storageValues
```
StephenButtolph, line 589: I don't really like having the variables referencing something incorrect.. I think we should either nolint this or directly return the results of the append.
```suggestion
	return append(storageKeys, acctKeys...), append(storageValues, acctValues...)
```

Technically we could use `slices.Concat` here... But that would introduce another copy of the slices that we probably don't want to do in prod code.

params/config_extra.go
```diff
@@ -50,13 +50,15 @@ func SetEthUpgrades(c *ChainConfig) error {
 	extra := GetExtra(c)
 	// Because Fuji and Mainnet have already accepted the Berlin and London
 	// blocks, it is assumed that they are scheduled for activation.
-	if c.ChainID != nil && AvalancheFujiChainID.Cmp(c.ChainID) == 0 {
+
```
StephenButtolph, line 53:
```suggestion
```

plugin/evm/vm.go
```diff
@@ -1131,7 +1131,7 @@ func (vm *VM) startContinuousProfiler() {
 		return
 	}
 	vm.profiler = profiler.NewContinuous(
-		filepath.Join(vm.config.ContinuousProfilerDir),
+		filepath.Clean(vm.config.ContinuousProfilerDir),
```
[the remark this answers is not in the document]
StephenButtolph: Sorry, I was referring to originally when it was: [originally added](https://github.com/ava-labs/coreth-internal/pull/197)

StephenButtolph (approved):

PR #1188, opened 2025-09-05
Fix test aggregation goroutine leak

michaelkaplan13:
## Why this should be merged
Fixes small potential goroutine leak in tests, as identified by Trail of Bits.

## How this works
The signature fetching routine potentially would try to push to an unbuffered channel even after the receiver loop returned early (after getting enough signatures), which would cause them to block indefinitely. The buffered channel allows the routines to still be able to push even after the receiver exits, and the select allow for skipping pushing to the channel all together if the context has already been cancelled by the receiver.

## How this was tested
N/A

## Need to be documented?
N/A

No

### DIFF
--- a/tests/warp/aggregator/aggregator.go
+++ b/tests/warp/aggregator/aggregator.go
@@ -63,7 +63,7 @@ func (a *Aggregator) AggregateSignatures(ctx context.Context, unsignedMessage *a
 	defer signatureFetchCancel()
 
 	// Fetch signatures from validators concurrently.
-	signatureFetchResultChan := make(chan *signatureFetchResult)
+	signatureFetchResultChan := make(chan *signatureFetchResult, len(a.validators))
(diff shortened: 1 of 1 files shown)


tests/warp/aggregator/aggregator.go
```diff
@@ -86,7 +86,10 @@ func (a *Aggregator) AggregateSignatures(ctx context.Context, unsignedMessage *a
 					"err", err,
 					"msgID", unsignedMessage.ID(),
 				)
-				signatureFetchResultChan <- nil
+				select {
+				case <-signatureFetchCtx.Done():
+				case signatureFetchResultChan <- nil:
+				}
```
StephenButtolph, line 92: Why are these select cases modified? Now that the length of the channel is sufficient, these are guaranteed never to block right?

PR #1200, opened 2025-09-10
style: enable prealloc linter 

JonathanOppenheimer:
## Why this should be merged

### DIFF
--- a/plugin/evm/atomic/txpool/mempool.go
+++ b/plugin/evm/atomic/txpool/mempool.go
@@ -131,10 +131,11 @@ func (m *Mempool) ForceAddTx(tx *atomic.Tx) error {
 func (m *Mempool) checkConflictTx(tx *atomic.Tx) (uint256.Int, ids.ID, []*atomic.Tx, error) {
 	utxoSet := tx.InputUTXOs()
 
-	var (
+	var ( //nolint:prealloc
 		highestGasPrice             uint256.Int
 		highestGasPriceConflictTxID ids.ID
-		conflictingTxs              []*atomic.Tx //nolint:prealloc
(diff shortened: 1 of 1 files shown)


precompile/contracts/warp/contract.go
```diff
@@ -321,7 +321,7 @@ func UnpackSendWarpEventDataToMessage(data []byte) (*warp.UnsignedMessage, error
 
 // createWarpPrecompile returns a StatefulPrecompiledContract with getters and setters for the precompile.
 func createWarpPrecompile() contract.StatefulPrecompiledContract {
-	var functions []*contract.StatefulPrecompileFunction
+	functions := make([]*contract.StatefulPrecompileFunction, 0, len(WarpABI.Methods))
```
alarso16, line 324: Maybe use `abiFunctionMap` instead

plugin/evm/atomic/txpool/mempool.go
```diff
@@ -134,7 +134,7 @@ func (m *Mempool) checkConflictTx(tx *atomic.Tx) (uint256.Int, ids.ID, []*atomic
 	var (
 		highestGasPrice             uint256.Int
 		highestGasPriceConflictTxID ids.ID
-		conflictingTxs              []*atomic.Tx
+		conflictingTxs              = make([]*atomic.Tx, 0, len(utxoSet))
```
alarso16, line 137: Wait, is the linter right this time? If it's not present in `utxoSpenders` (which is apparently expected sometimes if it doesn't return an error), then should we really allocate the maximum?

JonathanOppenheimer: Yup we can ignore this.

PR #1217, opened 2025-09-15
Have `checkPredicates` account for maximum available gas capacity

JonathanOppenheimer:
## Why this should be merged
This resolves an outstanding issue- we do all the intrinsic gas checks up front rather than waiting until later. 

## How this was tested
I added a new test `TestBlockGasValidation` that hits both of the two new errors. 

## Need to be documented?
No

No


plugin/evm/wrapped_block.go
```diff
@@ -260,6 +266,10 @@ func (b *wrappedBlock) verify(predicateContext *precompileconfig.PredicateContex
 	// bootstrapping only verifies blocks that have been canonically accepted by
 	// the network, these checks would be guaranteed to pass on a synced node.
 	if b.vm.bootstrapped.Get() {
+		if err := b.verifyIntrinsicGas(); err != nil {
+			return fmt.Errorf("failed to verify intrinsic gas: %w", err)
+		}
```
ceyonur, line 271: (not directly relevant to this PR) We should move these to `semanticVerify` in another PR

PR #1218, opened 2025-09-15
refactor: move header time verification

### LINKED ISSUE #1211: Time verification scattered
Since we now require additional timestamp verification for ACP-226, it makes sense to gather all verification (some done in `plugin/evm/wrapped_block.go`, some in `consensus/dummy/consensus.go`, into a single function added in #1208. Additionally, the time verification will become more complicated after #1157.

alarso16:
## Why this should be merged

It's more sensible to do all time verification for block headers in a single location for ACP-226, since it is spread currently between `semanticVerify` and `eng.VerifyHeader`, which is unnecessary. One fun side effect is that we no longer need to provide a clock to the consensus callback engine.

Closes #1211

## How this works

Moves a check from consensus callbacks

## How this was tested

Existing UT, added one

## Need to be documented?

No.

No.

### DIFF
--- a/plugin/evm/customheader/time_test.go
+++ b/plugin/evm/customheader/time_test.go
@@ -15,82 +15,141 @@ import (
 	"github.com/ava-labs/coreth/utils"
 )
 
+func generateHeader(timeSeconds uint64, timeMilliseconds *uint64) *types.Header {
+	return customtypes.WithHeaderExtra(
+		&types.Header{
+			Time: timeSeconds,
+		},
+		&customtypes.HeaderExtra{
+			TimeMilliseconds: timeMilliseconds,
+		},
+	)
+}
+
 func TestVerifyTime(t *testing.T) {
 	var (
-		time        = time.Unix(1714339200, 123_456_789)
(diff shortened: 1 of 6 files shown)


ceyonur (approved):

plugin/evm/customheader/time_test.go
```diff
@@ -15,81 +15,139 @@ import (
...
-			header := customtypes.WithHeaderExtra(
-				&types.Header{
-					Time: test.timeSeconds,
-				},
-				&customtypes.HeaderExtra{
-					TimeMilliseconds: test.timeMilliseconds,
-				},
-			)
-			err := VerifyTime(test.extraConfig, header, time)
+			// if parentHeader is nil, copy the current header
```
JonathanOppenheimer, line 145: If I was Arran reviewing, I'd say this comment just describes what the code does and isn't useful - what's the "why," rather than the "what"?

JonathanOppenheimer (approved):
I agree with Cey, this is a very good clean up!

Issue #1196, opened 2025-08-27
Epic: libevm upstreaming


JonathanOppenheimer:
There is an incredibly large amount of duplicated `geth` code, present in both `coreth` and `subnet-evm.` The correct amount of duplicated `geth` code that should be present is 0! 😂

For all `geth` code, two basic things need to happen: 

1. Code movement 
   - If our modifications to the `geth` code can be parsed out, they should be moved into a separate file in the directory, named `*_extra.go`. 
   - If our modifications to the `geth` code can not be parsed out, libevm must provide the proper hooks for consumers of it to be able to register themselves to alter the EVM behavior how they would like. This process is "libevm-ifying" packages. In short, any remaining packages that have modifications from geth will need to be libevm-ified.
2. Code deletion
   - Once the duplicated code is move or separated, the duplicate code package present in an evm repository should be deleted entirely, and imports adjusted to consume from libevm. 

The following list of packages must be addressed: 

**coreth packages remaining**:

- [ ] `accounts/abi` 
- [x] `cmd/abigen` 
- [x] `cmd/utils` 
- [ ] `consensus` 
- [ ] `core` 😨😨😨⏰
<img width="877" height="50" alt="Image" src="https://github.com/user-attachments/assets/5440f2ee-881b-452e-ac04-dc1d30c7fbbe" />

- [ ] `eth` 😨😨
- [ ] `internal` 
    - This will most likely need to be one of the last packages upstreamed 
- [ ] `miner` 
- [ ] `node` 
- [ ] `params` 
- [ ] `plugin/evm/customtypes` 
- [ ] `rpc` 
- [x] `signer/core/apitypes` 
- [ ] `tests` 
- [ ] `triedb/hashdb`
- [ ] `triedb/pathdb`

**subnet-evm packages remaining**:
_coming soon_

PR #1224, opened 2025-09-16
Remove libevm `cmd/*` packages. 

JonathanOppenheimer:
## Why this should be merged
These `cmd/*` packags are identical to the versions in `libevm`. We should reduce code duplication by using the `libevm` package. This is in-line with the longer-term goals slimming down `coreth` and `subnet-evm` and merging them. 

Works towards https://github.com/ava-labs/coreth/issues/1196


cmd/abigen/main.go
```diff
@@ -1,252 +0,0 @@
...
-
-import (
-	"encoding/json"
-	"fmt"
-	"io"
-	"os"
-	"regexp"
-	"strings"
-
-	"github.com/ava-labs/coreth/accounts/abi/bind"
```
alarso16, line 38: This package is different, does that matter?

JonathanOppenheimer: We never use these packages anywhere in `coreth` so it doesn't matter.

ceyonur (approved):

PR #1235, opened 2025-09-18
Remove unused types from codec registration

StephenButtolph:
## Why this should be merged

I noticed that we currently register some types that should never be used on the C-chain.

Removing them from the codec means we never need to handle them later on (fail fast rather than fail late).

## How this works

Skips registering `secp256k1fx.MintOutput` and `secp256k1fx.MintOperation`

## How this was tested

CI

## Need to be documented?

no

### DIFF
--- a/plugin/evm/atomic/codec.go
+++ b/plugin/evm/atomic/codec.go
@@ -36,9 +36,13 @@ func init() {
 	lc.SkipRegistrations(3)
 	errs.Add(
 		lc.RegisterType(&secp256k1fx.TransferInput{}),
-		lc.RegisterType(&secp256k1fx.MintOutput{}),
+	)
+	lc.SkipRegistrations(1)
+	errs.Add(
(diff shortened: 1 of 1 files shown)


JonathanOppenheimer (approved):

plugin/evm/atomic/codec.go
```diff
@@ -36,9 +36,13 @@ func init() {
 	lc.SkipRegistrations(3)
 	errs.Add(
 		lc.RegisterType(&secp256k1fx.TransferInput{}),
-		lc.RegisterType(&secp256k1fx.MintOutput{}),
+	)
+	lc.SkipRegistrations(1)
+	errs.Add(
 		lc.RegisterType(&secp256k1fx.TransferOutput{}),
```
alarso16, line 42: Why is this add multiple times?

powerslider (approved):

alarso16 (approved):

PR #1244, opened 2025-09-19
feat: Enable PathDB config

### LINKED ISSUE #1222: PathDB config option
PathDB should be added as a config option on startup, with a big warning saying never to use it in production (since it's completely untested). This will unblock performance testing comparing to Firewood

alarso16:
## Why this should be merged

Firewood needs a decent performance comparison, and HashDB is not it. It is extremely slow, but PathDB is a more reasonable alternative to compare against.

Note: PathDB largely does not work, but unblocks performance testing

Closes #1222

## How this works

Enables PathDB as a config option.

## How this was tested

Ensures that PathDB is run when the config is provided. WARNING: this is not tested sufficiently to run a validator - it will likely result in an invalid state.

## Need to be documented?

Yes.

No - no need to advertise a database that won't work.


JonathanOppenheimer (approved):
As an addendum to your comment - I played around with `pathdb` when removing it entirely here: https://github.com/ava-labs/coreth/pull/1167.

It seemingly does not work / is not compatible. If this doesn't matter, this PR LGTM.

Issue #1137, opened 2025-08-21
bug: inspect databases


RodrigoVillar:
Currently, enabling the `inspect-database` option in Coreth results in VM initialization erroring. I first noticed this bug here: https://github.com/ava-labs/avalanchego/actions/runs/17109691592/job/48527748332?pr=4190#step:4:1158

Code Path/Steps To Reproduce:

1. Prior to starting up the VM, make sure that the `inspect-database` option in the EVM config is set to `true` [[ref](https://github.com/ava-labs/coreth/blob/ca842473e5861db83d7d282fb628203466c6b079/plugin/evm/config/config.go#L136)]
2. In `Initialize()` [[ref](https://github.com/ava-labs/coreth/blob/ca842473e5861db83d7d282fb628203466c6b079/plugin/evm/vm.go#L264)], we create the VM databases including `chainDB` which is instantiated as follows: `vm.chaindb = rawdb.NewDatabase(database.WrapDatabase(prefixdb.NewNested(ethDBPrefix, db)))` [[ref](https://github.com/ava-labs/coreth/blob/ca842473e5861db83d7d282fb628203466c6b079/plugin/evm/vm_database.go#L25)]. Note that `vm.chainDB` is wrapped in the `nofreezedb` type
3. Since we specified `inspect-database` to `true`, we call `inspectDatabases()`[[ref](https://github.com/ava-labs/coreth/blob/ca842473e5861db83d7d282fb628203466c6b079/plugin/evm/vm.go#L320)]
4. In `inspectDatabases()`, we call `rawdb.InspectDatabases()` on `vm.chainDB` which eventually inspects the DB freezers [[ref](https://github.com/ava-labs/libevm/blob/9e4c147bf2b7df308efe49d340cf443c6a157d2f/core/rawdb/database.go#L607)].
5. The freezer inspection code eventually calls one of the methods of `chainDB`, which results in the error `errNotSupported` being returned [[ref](https://github.com/ava-labs/libevm/blob/9e4c147bf2b7df308efe49d340cf443c6a157d2f/core/rawdb/database.go#L93-L167)].

This error occurs because in `Coreth`, we do not utilize a freezer database while `geth` does. A proposed solution for this would be to inject a flag into `libevm` (since `InspectDatabases` lives there) such that we can skip the freezer inspection; this could look like the following:

```golang
  // Inspect all registered append-only file store then.
ancients, err := inspectFreezers(db)
if err != nil {
[...]

PR #1245, opened 2025-09-19
fix: database inspection

RodrigoVillar:
## Why this should be merged

As mentioned in #1137, starting an instance of Coreth with database inspection enabled will fail as Coreth doesn't use freezers but upstream code (Libevm) previously assumed so. With https://github.com/ava-labs/libevm/pull/220 now available, we can now skip freezer inspection.

## How this works

Adds an option to skip freezer inspection when inspecting the database.

## How this was tested

Upstream: https://github.com/ava-labs/libevm/blob/ba9ab538b90236c480c412c54816717f3eb501e9/core/rawdb/database.libevm_test.go#L25-L49

## Need to be documented?

Not sure

Not sure

### DIFF
--- a/plugin/evm/vm_database.go
+++ b/plugin/evm/vm_database.go
@@ -35,7 +35,7 @@ func (vm *VM) initializeDBs(db avalanchedatabase.Database) {
 func (vm *VM) inspectDatabases() error {
 	start := time.Now()
 	log.Info("Starting database inspection")
-	if err := rawdb.InspectDatabase(vm.chaindb, nil, nil); err != nil {
+	if err := rawdb.InspectDatabase(vm.chaindb, nil, nil, rawdb.WithSkipFreezers()); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


plugin/evm/vm_test.go
```diff
@@ -2421,3 +2422,15 @@ func TestBlockGasValidation(t *testing.T) {
...
+func TestInspectDatabases(t *testing.T) {
+	require := require.New(t)
+
+	var (
+		vm = newDefaultTestVM()
+		db = memdb.New()
+	)
+
+	vm.initializeDBs(db)
+	require.NoError(vm.inspectDatabases())
```
ceyonur, line 2435: (nit) since there is only 1 require usage 
```suggestion
	var (
		vm = newDefaultTestVM()
		db = memdb.New()
	)

	vm.initializeDBs(db)
	require.NoError(t, vm.inspectDatabases())
}
```

ceyonur (approved):
just an optional nit

ARR4N (approved):
Approving pending the suggestion by @ceyonur

Issue #1243, opened 2025-09-19
ACP-226 Configureable Genesis Value


ceyonur:
We want this to be configurable especially for subnet-evm

Additionally, for coreth should we set this such that the initial delay is 2s?

I know it doesn't matter for mainnet/fuji, but I wonder if we shouldn't be starting the local network with 1ms block delays.

_Originally posted by @StephenButtolph in https://github.com/ava-labs/coreth/pull/1234#discussion_r2361164573_

PR #1246, opened 2025-09-19
ACP-226: Make MinDelayExcess Configurable

ceyonur:
## Why this should be merged

* We should use default from `acp226` package
* `MinDelayExcess` should be configurable especially for L1s 

Closes: #1243 

## How this works

Adds `MinDelayExcess` to genesis spec so it can be marshaled and set in the genesis header.

## How this was tested

added tests

## Need to be documented?

no (at least not for C-chain)

updated


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.23.9
 
 require (
 	github.com/VictoriaMetrics/fastcache v1.12.1
-	github.com/ava-labs/avalanchego v1.13.5-rc.4
+	github.com/ava-labs/avalanchego v1.13.6-0.20250919181458-24aa8901931e
```
alarso16, line 7: Why does this require updating avalanchego?

ceyonur:
We have discussed this and concluded that adding a new field to the `Genesis` struct will increase the diff and complexity between upstream code. It would also not worth that much since this can only affect new chains to be created in Granite rules, existing chains activating Granite won't be able to use that. All chains can/will be able to use node configs to tune the excess, so use-case for new chains is very minimal and redundant.

Issue #1170, opened 2025-08-29
ACP-226: Enforce minimumBlockDelay post-Granite


michaelkaplan13:
- Add minimumBlockDelay excess in block header
- Update block verification to enforce minimumBlockDelay

Note: The `minimumBlockDelay` in milliseconds itself will **not** be included in the block header.

PR #1263, opened 2025-09-23
ACP-226: add min delay excess to block and verify

ceyonur:
## Why this should be merged

It adds calculated min delay excess to the header and verifies it

Closes: #1170 

## How this works

* calculations in in `customheader` pkg
* add to the header in `consensus.go/FinalizeAndAssemble`
* verify in `wrapped_block.go/semanticVerify`

## How this was tested

added uts

## Need to be documented?

no

no


plugin/evm/customheader/min_delay_excess.go
```diff
@@ -0,0 +1,119 @@
...
+	minDelayExcess := acp226.DelayExcess(acp226.InitialDelayExcess)
+	if config.IsGranite(parent.Time) {
+		// If the parent block was running with ACP-226, we start with the
+		// resulting min delay excess from the parent block.
+		parentMinDelayExcess := customtypes.GetHeaderExtra(parent).MinDelayExcess
+		if parentMinDelayExcess == nil {
+			return 0, fmt.Errorf("%w: %s", errParentMinDelayExcessNil, parent.Hash())
+		}
+		minDelayExcess = acp226.DelayExcess(*parentMinDelayExcess)
+	}
```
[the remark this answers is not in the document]
ceyonur: that is being checked during the parent verification.

plugin/evm/customheader/min_delay_excess_test.go
```diff
@@ -0,0 +1,235 @@
...
+			header:      &types.Header{Time: 1001},
+			expectedErr: nil,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			if test.expectedErr != nil && test.expectedErr.Error() == "panic expected" {
+				// Test panic cases
+				require.Panics(t, func() {
```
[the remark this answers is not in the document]
ceyonur: ahh nice catch, it was from a previously possible case which is not relevant anymore.

plugin/evm/customheader/min_delay_excess_test.go
```diff
@@ -0,0 +1,220 @@
...
+	"github.com/ava-labs/avalanchego/vms/evm/acp226"
+	"github.com/ava-labs/libevm/core/types"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/coreth/params/extras"
+	"github.com/ava-labs/coreth/plugin/evm/customtypes"
+	"github.com/ava-labs/coreth/utils"
+)
+
+func generateHeaderWithMinDelayExcess(timeSeconds uint64, minDelayExcess *uint64) *types.Header {
```
powerslider, line 18: nit: Move this function to the end of the file given that it is unexported.

plugin/evm/customheader/min_delay_excess.go
```diff
@@ -0,0 +1,127 @@
...
+var (
+	errRemoteMinDelayExcessSet = errors.New("remote min delay excess should be nil")
+	errRemoteMinDelayExcessNil = errors.New("remote min delay excess should not be nil")
+	errIncorrectMinDelayExcess = errors.New("incorrect min delay excess")
+	errParentMinDelayExcessNil = errors.New("parent min delay excess should not be nil")
+)
+
+// MinDelay
+//
+// If the `desiredMinDelayExcess` is nil, the parent's delay excess is used.
```
powerslider, line 26: nit: improvement on the godoc of this function:
```suggestion
// MinDelayExcess computes the min delay excess to include in the header.
// If the `desiredMinDelayExcess` is nil, the parent's delay excess (or the
// initial delay excess at activation) is used.
```

consensus/dummy/consensus.go
```diff
@@ -334,6 +337,18 @@ func (eng *DummyEngine) FinalizeAndAssemble(chain consensus.ChainHeaderReader, h
 	}
 	header.Extra = append(extraPrefix, header.Extra...)
 
+	// Set the min delay excess
+	minDelayExcess, err := customheader.MinDelayExcess(
```
michaelkaplan13, line 341: For my own understanding, is there a reason why we don't use `customheader.ExtraPrefix` for handling setting the min delay excess as we already do for the desired gas target excess?

plugin/evm/customheader/min_delay_excess.go
```diff
@@ -0,0 +1,127 @@
...
+		if customtypes.GetHeaderExtra(header).MinDelayExcess != nil {
+			return fmt.Errorf("%w: %s", errRemoteMinDelayExcessSet, header.Hash())
+		}
+	}
+	return nil
+}
+
+// minDelayExcessBeforeBlock takes the previous header and the timestamp of its child
+// block and calculates the minimum delay before the child block is executed.
+// Should only be called with timestamp in Granite.
```
michaelkaplan13, line 90: This comment seems inaccurate for ACP-226 because it doesn't take the child timestamp.

michaelkaplan13: IIUC, this function is really just returning the effective min delay excess of the given header, right?

plugin/evm/config/config.go
```diff
@@ -27,6 +27,11 @@ type Config struct {
 	// default to use the parent block's target gas per second.
 	GasTarget *gas.Gas `json:"gas-target,omitempty"`
 
+	// MinDelayTarget is the minimum delay per second that this node will attempt to use
+	// when creating blocks. If this config is not specified, the node will
+	// default to use the parent block's target delay per second.
```
StephenButtolph, line 32: I don't think this comment is quite accurate. This seems like the `minimum time between blocks denominated in milliseconds` rather than the `minimum delay per second`

plugin/evm/customheader/min_delay_excess.go
```diff
@@ -0,0 +1,127 @@
...
+		if err != nil {
+			return fmt.Errorf("calculating expected min delay excess: %w", err)
+		}
+
+		if *remoteDelayExcess != uint64(expectedDelayExcess) {
+			return fmt.Errorf("%w: expected %+v, found %+v",
+				errIncorrectMinDelayExcess,
+				expectedDelayExcess,
+				remoteDelayExcess,
+			)
```
StephenButtolph, line 77:
```suggestion
			return fmt.Errorf("%w: expected %d, found %d",
				errIncorrectMinDelayExcess,
				expectedDelayExcess,
				*remoteDelayExcess,
			)
```

plugin/evm/customheader/min_delay_excess.go
```diff
@@ -0,0 +1,127 @@
...
+		if parentMinDelayExcess == nil {
+			return 0, fmt.Errorf("%w: %s", errParentMinDelayExcessNil, parent.Hash())
+		}
+		minDelayExcess = acp226.DelayExcess(*parentMinDelayExcess)
+	}
+
+	return minDelayExcess, nil
+}
+
+// minDelayExcessAfterBlock takes the previous header and returns the min delay excess.
```
StephenButtolph: Can we document the behavior around nil/non-nil `desiredMinDelayExcess` values?

plugin/evm/customheader/min_delay_excess.go
ceyonur: I think we probably don't need this function as in acp-176, so just merged two functions.

plugin/evm/wrapped_block.go
StephenButtolph, line 345: Can we assign `b.ethBlock.Header()` to a var? It's actually pretty expensive to call it multiple times because it does a full copy of the header each time.

plugin/evm/vm_test.go
```diff
@@ -2052,3 +2054,77 @@ func TestBlockGasValidation(t *testing.T) {
...
+			require.NoError(err)
+
+			// Check the min delay excess in the header
+			ethBlock := blk.(*chain.BlockWrapper).Block.(*wrappedBlock).ethBlock
+			headerExtra := customtypes.GetHeaderExtra(ethBlock.Header())
+
+			require.Equal(test.expectedMinDelayExcess, headerExtra.MinDelayExcess, "expected %s, got %s",
+				utils.SafeDerefUint64String(utils.Uint64PtrFrom(test.expectedMinDelayExcess)),
+				utils.SafeDerefUint64String(utils.Uint64PtrFrom(headerExtra.MinDelayExcess)))
+		})
```
StephenButtolph, line 2128: imo, I feel like just:
```suggestion
			require.Equal(test.expectedMinDelayExcess, headerExtra.MinDelayExcess, "expected %d, got %d",
				test.expectedMinDelayExcess,
				headerExtra.MinDelayExcess)
		})
```
is enough to allow the output to be readable (without introducing the `utils.SafeDerefUint64String` helper)... But if we really want to keep it I feel like we could make it a test helper.

plugin/evm/customheader/customheadertest/test_min_delay_excess.go
```diff
@@ -0,0 +1,11 @@
...
+// See the file LICENSE for licensing terms.
+
+package customheadertest
+
+import "github.com/ava-labs/avalanchego/vms/evm/acp226"
+
+func NewDelayExcessPtr(value uint64) *acp226.DelayExcess {
+	delayExcess := acp226.DelayExcess(value)
+	return &delayExcess
+}
```
StephenButtolph, line 11: Can we unify all of these implementations?
- https://github.com/ava-labs/coreth/blob/0a6ce77563f1865b0a05f8d507bb6a4ba222aa26/plugin/evm/customtypes/header_ext_test.go#L192
- https://github.com/ava-labs/coreth/blob/0a6ce77563f1865b0a05f8d507bb6a4ba222aa26/utils/numbers.go#L11

I think a single:
```golang
func PointerTo[T any](x T) *T {
	return &x
}
```
should work for all our usages.

plugin/evm/customheader/min_delay_excess.go
```diff
@@ -0,0 +1,118 @@
...
+	if desiredMinDelayExcess != nil {
+		minDelayExcess.UpdateDelayExcess(*desiredMinDelayExcess)
+	}
+	return minDelayExcess, nil
+}
+
+func NewDelayExcessPtr(value uint64) *acp226.DelayExcess {
+	delayExcess := acp226.DelayExcess(value)
+	return &delayExcess
+}
```
StephenButtolph, line 118: This seems to have been accidentally included (was moved to a different file)
```suggestion
```

plugin/evm/vm_test.go
ceyonur: it's indeed annoying to debug...

StephenButtolph (approved):

PR #1287, opened 2025-09-26
refactor out the time calculations to customtypes package

ceyonur:
## Why this should be merged

After adding time milliseconds and the delay, we mostly own the header time calculations. This calculation can be moved to `customheader` package.

## How this works

Moves the calculation to `customheader` package

## How this was tested

added UTs

## Need to be documented?

No

No


plugin/evm/customheader/time.go
```diff
@@ -26,6 +26,30 @@ var (
 	ErrTimeMillisecondsBeforeGranite = errors.New("TimeMilliseconds should be nil before Granite activation")
 )
 
+// GetTimestamp calculates the timestamp (in seconds and milliseconds) for the header based on the parent's timestamp and the current time.
+// First return value is the timestamp in seconds, second return value is the timestamp in milliseconds.
+func GetTimestamp(parent *types.Header, now time.Time) (uint64, uint64) {
```
alarso16, line 31: I do like the idea of a `GetTimestamp` which returns `parent.Time` and `parent.TimeMilliseconds` if set, otherwise `1000*parentTime`, but this isn't really what this does. This is more like `GetNextTimestamp`

plugin/evm/customheader/time.go
```diff
@@ -26,6 +26,30 @@ var (
...
+	// per second.
+	parentExtra := customtypes.GetHeaderExtra(parent)
+	if parent.Time >= timestamp ||
+		(parentExtra.TimeMilliseconds != nil && *parentExtra.TimeMilliseconds >= timestampMS) {
+		timestamp = parent.Time
+		// If the parent has a TimeMilliseconds, use it. Otherwise, use the parent time * 1000.
+		if parentExtra.TimeMilliseconds != nil {
+			timestampMS = *parentExtra.TimeMilliseconds
+		} else {
+			timestampMS = parent.Time * 1000 // TODO: establish minimum time
```
michaelkaplan13, line 47: Is this still a TODO? It seems correct as is to me, so I don't understand what will need to change for it if so.

michaelkaplan13 (approved):
Makes sense to me, just have 1 question if the TODO is still needed.

PR #1288, opened 2025-09-26
ACP-226: adjust ACP-176 to use milliseconds

alarso16:
## Why this should be merged

Starting with ACP-226, the capacity must be refilled at a millisecond interval. This passes milliseconds instead of seconds down to the ACP-176 implementation. The additional method was added to avalanchego, so this was upstreamed.

## How this works

Uses milliseconds instead of seconds for all gas-related calculations.

## How this was tested

Existing UT

## Need to be documented?

No

Yes

### DIFF
--- a/plugin/evm/customheader/base_fee.go
+++ b/plugin/evm/customheader/base_fee.go
@@ -50,7 +50,7 @@ func BaseFee(
 // Warning: This function should only be used in estimation and should not be
 // used when calculating the canonical base fee for a block.
 //
-// TODO(#1260): Adapt for Granite upgrade.
+// TODO(#1290): Adapt for Granite upgrade.
 func EstimateNextBaseFee(
 	config *extras.ChainConfig,
 	parent *types.Header,


.github/workflows/ci.yml
```diff
@@ -100,11 +100,15 @@ jobs:
         with:
           go-version-file: "go.mod"
       - name: Run e2e tests
-        uses: ava-labs/avalanchego/.github/actions/run-monitored-tmpnet-cmd@cc3242f48a6262c52057e98a85d6cf8184dfaca3
+        uses: ava-labs/avalanchego/.github/actions/run-monitored-tmpnet-cmd@36baa0509d61679085d35bffa590d63732b0a7bd
         with:
           run: ./scripts/run_task.sh test-e2e-ci
+          prometheus_url: ${{ secrets.PROMETHEUS_URL || '' }}
```
michaelkaplan13, line 106: Are these proetheus/loki changes required because of the AvalancheGo update in this PR? Otherwise seem like they could be made separately

plugin/evm/customheader/dynamic_fee_state.go
```diff
@@ -43,7 +44,17 @@ func feeStateBeforeBlock(
...
-	state.AdvanceTime(timestamp - parent.Time)
+	switch {
+	case config.IsGranite(timestamp):
+		parentMS := parent.Time * 1000
+		if config.IsGranite(parent.Time) {
+			parentMS = *customtypes.GetHeaderExtra(parent).TimeMilliseconds
+		}
+		state.AdvanceMilliseconds(timeMS - parentMS)
+	case config.IsFortuna(timestamp):
+		// If the parent block was not running with ACP-176, we start with the
```
michaelkaplan13, line 55: Start with the what? :)

plugin/evm/customheader/dynamic_fee_state.go
```diff
@@ -43,7 +44,17 @@ func feeStateBeforeBlock(
 		}
 	}
 
-	state.AdvanceTime(timestamp - parent.Time)
+	switch {
+	case config.IsGranite(timestamp):
+		parentMS := parent.Time * 1000
+		if config.IsGranite(parent.Time) {
+			parentMS = *customtypes.GetHeaderExtra(parent).TimeMilliseconds
+		}
```
michaelkaplan13, line 52: This logic appears repeated in a few places..is there a clean/easy way to move it into a `header.GetTimeMS()` type of helper?

plugin/evm/customheader/dynamic_fee_state.go
michaelkaplan13: Seems like this comment may have been an accidental copy and should be removed actually

plugin/evm/upgrade/acp176/acp176.go
michaelkaplan13, line 1: Not 100% sure how easy this would be, but it may be best to split this into 2 PRs for ease of review:
1. Using the upstreamed `acp176` package (should be a trivial review)
2. Updating to using millisecond based time post-Granite (will have a smaller diff than currently)

Thoughts?

plugin/evm/customheader/dynamic_fee_state.go
```diff
@@ -56,7 +67,11 @@ func feeStateAfterBlock(
 	desiredTargetExcess *gas.Gas,
 ) (acp176.State, error) {
 	// Calculate the gas state after the parent block
-	state, err := feeStateBeforeBlock(config, parent, header.Time)
+	timeMS := header.Time * 1000
+	if config.IsGranite(header.Time) {
+		timeMS = *customtypes.GetHeaderExtra(header).TimeMilliseconds
+	}
+	state, err := feeStateBeforeBlock(config, parent, timeMS)
```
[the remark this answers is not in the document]
powerslider: I would advocate in encapsulating this logic in a function somewhere and re-using in at all those places.

consensus/dummy/consensus.go
```diff
@@ -129,16 +129,19 @@ func verifyHeaderGasFields(config *extras.ChainConfig, header *types.Header, par
 	}
 
 	// Verify header.BaseFee matches the expected value.
-	expectedBaseFee, err := customheader.BaseFee(config, parent, header.Time)
+	headerExtra := customtypes.GetHeaderExtra(header)
+	timeMS := header.Time * 1000
+	if config.IsGranite(header.Time) {
+		timeMS = *headerExtra.TimeMilliseconds
+	}
```
powerslider, line 136: Check https://github.com/ava-labs/coreth/pull/1288#issuecomment-3346446701
```suggestion
    headerExtra := customtypes.GetHeaderExtra(header)
    timeMS := customtypes.HeaderTimeMilliseconds(header)
```

plugin/evm/customheader/gas_limit.go
```diff
@@ -101,8 +107,22 @@ func VerifyGasLimit(
...
+		if header.GasLimit != uint64(maxCapacity) {
+			return fmt.Errorf("%w: have %d, want %d",
+				errInvalidGasLimit,
+				header.GasLimit,
+				maxCapacity,
+			)
+		}
 	case config.IsFortuna(header.Time):
-		state, err := feeStateBeforeBlock(config, parent, header.Time)
+		state, err := feeStateBeforeBlock(config, parent, header.Time*1000)
```
powerslider, line 125: Check https://github.com/ava-labs/coreth/pull/1288#issuecomment-3346446701

```suggestion
    case config.IsFortuna(header.Time):
        state, err := feeStateBeforeBlock(config, parent, customtypes.HeaderTimeMilliseconds(header))
```

powerslider, line 125: Check https://github.com/ava-labs/coreth/pull/1288#issuecomment-3346446701

```suggestion
    case config.IsFortuna(header.Time):
        state, err := feeStateBeforeBlock(config, parent, customtypes.HeaderTimeMilliseconds(header))
```

plugin/evm/customheader/gas_limit.go
```diff
@@ -188,7 +209,12 @@ func RemainingAtomicGasCapacity(
 		return ap5.AtomicGasLimit, nil
 	}
 
-	state, err := feeStateBeforeBlock(config, parent, header.Time)
+	timeMS := header.Time * 1000
+	if config.IsGranite(header.Time) {
+		timeMS = *customtypes.GetHeaderExtra(header).TimeMilliseconds
+	}
```
powerslider, line 215: Check #1288 (comment)

```suggestion
    timeMS := customtypes.HeaderTimeMilliseconds(header)
```

plugin/evm/customheader/gas_limit_test.go
```diff
@@ -9,27 +9,27 @@ import (
...
 	"github.com/ava-labs/coreth/plugin/evm/customtypes"
-	"github.com/ava-labs/coreth/plugin/evm/upgrade/acp176"
 	"github.com/ava-labs/coreth/plugin/evm/upgrade/ap0"
 	"github.com/ava-labs/coreth/plugin/evm/upgrade/ap1"
 	"github.com/ava-labs/coreth/plugin/evm/upgrade/ap5"
 	"github.com/ava-labs/coreth/plugin/evm/upgrade/cortina"
+	"github.com/ava-labs/coreth/utils"
 )
 
 func TestGasLimit(t *testing.T) {
```
powerslider, line 26: I see some good opportunities here to add the following test cases in `TestGasLimit`:

- `granite_initial_max_capacity` - Granite GasLimit at t=0 ms: Ensures GasLimit returns MaxCapacity on Granite genesis, validating the ms-based path and that GasLimit uses capacity ceiling, not current fill.

- `granite_after_1_5s_max_capacity` - Granite GasLimit at t=1500 ms: Confirms GasLimit remains MaxCapacity even mid-second, guarding against accidental dependence on fractional elapsed time.

Of course you would need to parameterize the `timestamp` and and set it this way when calling GasLimit:
```
got, err := GasLimit(config, test.parent, test.timestamp)
```

core/txpool/legacypool/legacypool_test.go
```diff
@@ -91,10 +92,15 @@ func (bc *testBlockChain) CurrentBlock() *types.Header {
...
-	return &types.Header{
+	header := &types.Header{
 		Number:   new(big.Int),
 		GasLimit: bc.gasLimit.Load(),
 	}
+	if params.GetExtra(bc.config).IsGranite(0) {
+		var timeMS uint64
+		customtypes.GetHeaderExtra(header).TimeMilliseconds = &timeMS
+	}
+	return header
```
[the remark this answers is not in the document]
alarso16: Because it panics! Not setting the seconds really sets it to the zero value, so it still sets it to something - we need it to be zero

plugin/evm/customheader/dynamic_fee_state.go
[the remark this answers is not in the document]
alarso16: Not all callers have a header

plugin/evm/customheader/gas_limit.go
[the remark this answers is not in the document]
alarso16: Not all callers have a header, but hopefully the helper fixes this

core/chain_makers.go
```diff
@@ -414,8 +415,7 @@ func (cm *chainMaker) makeHeader(parent *types.Block, gap uint64, state *state.S
 
 	if config.IsGranite(header.Time) {
 		headerExtra := customtypes.GetHeaderExtra(header)
-		timeMilliseconds := header.Time * 1000 // convert to milliseconds
-		headerExtra.TimeMilliseconds = &timeMilliseconds
+		headerExtra.TimeMilliseconds = &timeMS
```
michaelkaplan13, line 418: Does this mean that even post-granite activation, the `TimeMilliseconds` values for blocks will be round seconds?

miner/worker.go
```diff
@@ -273,7 +273,8 @@ func (w *worker) createCurrentEnvironment(predicateContext *precompileconfig.Pre
 	}
 
 	chainConfigExtra := params.GetExtra(w.chainConfig)
-	capacity, err := customheader.GasCapacity(chainConfigExtra, parent, header.Time)
+	timeMS := customtypes.HeaderTimeMilliseconds(header)
```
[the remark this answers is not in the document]
alarso16: I prefer `timeMS`, but I just added the `MS` suffix to whatever the local variable already was (weird example: json encoding uses `timestampMilliseconds` even though that's not the field name)

plugin/evm/customheader/base_fee.go
```diff
@@ -58,5 +60,7 @@ func EstimateNextBaseFee(
 	}
 
 	timestamp = max(timestamp, parent.Time, *config.ApricotPhase3BlockTimestamp)
-	return BaseFee(config, parent, timestamp)
+	// Use a fake header with this timestamp.
+	header := &types.Header{Time: timestamp}
+	return BaseFee(config, parent, header)
```
ceyonur, line 65: this is probably much worse than all these complicated stuff with handling ms and sec. as `BaseFee` can unintentionally require stuff from header in the future.

ceyonur: and I guess it would not even work since we don't set `ms` timestamp in the here.

ceyonur (changes requested):
I think that `EstimateBaseFee` would not work and we should not create fake headers. (marking as request changes to prevent an accidental merge)

plugin/evm/customheader/dynamic_fee_state.go
ceyonur, line 40: part of me wants to say we don't need to check them here as we check/fix that in block_verification and in worker.

ceyonur: additionaly I think it would be enough to check 
```
parentMS := customtypes.HeaderTimeMilliseconds(parent)
if timeMS < parentMS {
```

right? (because parentMS will be guaranteed to at least `parent.Time * 1000`)

plugin/evm/customheader/base_fee.go
```diff
@@ -48,6 +49,8 @@ func BaseFee(
 //
 // Warning: This function should only be used in estimation and should not be
 // used when calculating the canonical base fee for a block.
+//
+// TODO(#1260): Adapt for Granite upgrade.
```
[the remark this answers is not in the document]
ceyonur: I think we can actually change this here since we already  changed relevant/affecting parts.

plugin/evm/customtypes/header_ext.go
```diff
@@ -43,6 +43,17 @@ type HeaderExtra struct {
...
+// HeaderTimeMilliseconds returns the header timestamp in milliseconds.
+// If the header has the Granite field TimeMilliseconds set in extras, it is used.
+// Otherwise, it falls back to seconds-based Time multiplied by 1000.
+func HeaderTimeMilliseconds(h *ethtypes.Header) uint64 {
+	extra := GetHeaderExtra(h)
+	if extra != nil && extra.TimeMilliseconds != nil {
+		return *extra.TimeMilliseconds
+	}
+	return h.Time * 1000
+}
```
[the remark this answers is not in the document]
alarso16: It is, any functions you're referring to?

core/txpool/blobpool/blobpool_test.go
```diff
@@ -120,9 +120,6 @@ func (bc *testBlockChain) CurrentBlock() *types.Header {
 		}
 		config := params.GetExtra(bc.config)
 		timeMS := blockTime * 1000
-		if config.IsGranite(blockTime) {
-			customtypes.GetHeaderExtra(parent).TimeMilliseconds = &timeMS
-		}
```
ceyonur, line 125: I think this is panicking now

alarso16: This probably should stay, you're right. But it's panicking in legacy pool - turns out we needed to register customtypes.

plugin/evm/customheader/base_fee.go
michaelkaplan13, line 53: This should be #1290, right?

michaelkaplan13, line 53:
```suggestion
// TODO(#1290): Adapt for Granite upgrade.
```

#1290

michaelkaplan13 (approved):

ceyonur (approved):
LGTM

PR #1292, opened 2025-09-26
ACP-226: min block delay verify

ceyonur:
## Why this should be merged

This PR adds the minimum delay verification and is the final ACP-226 protocol changes.

## How this works

* Adds the `verifyMinDelay` which verifies the header based on parent's min excess delay and milliseconds timestamp.
* Handles the edge case of wall clock lagging behind the parent's time and returns an error.

## How this was tested

Added UTs

## Need to be documented?

No

yes

### DIFF
--- a/plugin/evm/customheader/time_test.go
+++ b/plugin/evm/customheader/time_test.go
@@ -146,7 +146,7 @@ func TestVerifyTime(t *testing.T) {
 			extraConfig:  extras.TestGraniteChainConfig,
 		},
 		{
-			name: "granite_min_delay_met",
+			name: "granite_initial_delay_met",
 			header: generateHeaderWithMinDelayExcessAndTime(
 				timeSeconds,
 				utils.NewUint64(timeMillis),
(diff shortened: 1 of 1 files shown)


plugin/evm/customheader/time.go
```diff
@@ -118,5 +129,46 @@ func VerifyTime(extraConfig *extras.ChainConfig, parent *types.Header, header *t
...
+	parentMinDelayExcess := customtypes.GetHeaderExtra(parent).MinDelayExcess
+	// Parent might not have a min delay excess if this is the first Granite block
+	// in this case we cannot verify the min delay,
+	// Otherwise parent should have been verified in VerifyMinDelayExcess
+	if parentMinDelayExcess == nil {
+		return nil
+	}
+
+	// if parent has no TimeMilliseconds, no min delay is required
+	// This should not happen as it should have been verified for Granite in VerifyTime.
```
alarso16, line 154: Maybe move this to the doc comment for the function, since this is a requirement for the input

PR #1295, opened 2025-09-29
Enforce block building delay only for retries

ceyonur:
## Why this should be merged

We only need to understand if we're trying block building or is it the first time we block a build with height. So we can immediately try to build a block without waiting for any delay, but only wait for the delay in retry.

## How this works

Adds `isRetry` and `lastBuiltHeight`

## How this was tested

existing UTs?

## Need to be documented?

no

no


plugin/evm/block_builder.go
```diff
@@ -58,10 +59,11 @@ func (vm *VM) NewBlockBuilder(extraMempool extension.BuilderMempool) *blockBuild
 }
 
 // handleGenerateBlock is called from the VM immediately after BuildBlock.
-func (b *blockBuilder) handleGenerateBlock() {
+func (b *blockBuilder) handleGenerateBlock(currentHeight uint64) {
```
StephenButtolph, line 62: I think we should be flagging on the parent hash rather than the height. There are some weird potential reorgs where we successfully build a block, but then get reorged and then attempt to build a new block at the same height (with no incorrect behavior).

plugin/evm/block_builder.go
```diff
@@ -116,38 +118,34 @@ func (b *blockBuilder) awaitSubmittedTxs() {
...
-	timeUntilNextBuild := MinBlockBuildingRetryDelay - timeSinceLastBuildTime
 	b.ctx.Log.Debug("Last time we built a block was too recent, waiting",
-		zap.Duration("timeUntilNextBuild", timeUntilNextBuild),
+		zap.Duration("timeUntilNextBuild", MinBlockBuildingRetryDelay),
 	)
 	select {
 	case <-ctx.Done():
 		return 0, ctx.Err()
-	case <-time.After(timeUntilNextBuild):
+	case <-time.After(MinBlockBuildingRetryDelay):
```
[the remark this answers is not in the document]
StephenButtolph: While I don't think we _need_ to base this delay off the time the last build attempt happened at. The `WaitForEvent` context is commonly cancelled by the consensus engine... So I don't think we should be resetting the delay timer every time we call it.

plugin/evm/block_builder.go
```diff
@@ -58,10 +59,11 @@ func (vm *VM) NewBlockBuilder(extraMempool extension.BuilderMempool) *blockBuild
 }
 
 // handleGenerateBlock is called from the VM immediately after BuildBlock.
-func (b *blockBuilder) handleGenerateBlock() {
+func (b *blockBuilder) handleGenerateBlock(currentHeight uint64) {
 	b.buildBlockLock.Lock()
 	defer b.buildBlockLock.Unlock()
-	b.lastBuildTime = time.Now()
+	b.isRetry = b.lastBuiltHeight == currentHeight
```
StephenButtolph, line 65: So, `isRetry` will be set to `true` after we have attempted to build a block at the same height 2 times in a row right?

- `waitForEvent` -> `PendingTxs`
- `buildBlock` -> `isRetry = false ; lastBuiltHeight = x`
- `waitForEvent` -> `PendingTxs`
- `buildBlock` -> `isRetry = true ; lastBuiltHeight = x`
- `waitForEvent` -> delay for 500ms

Based on the naming, I would have expected us to be delaying between the first call to `buildBlock` and the second call (rather than after the second call). Specifically:

- `waitForEvent` at `x` -> `PendingTxs`
- `buildBlock` -> `lastBuiltHeight = x ; lastBuiltTime = now()`
- `waitForEvent` at `x` -> delay for 500ms after `lastBuiltTime`

ceyonur: updated

StephenButtolph: I think this comment still applies.

I think the only way to delay the first call is if we fetch the currently preferred block in `WaitForEvent` and compare it to the last build block parent hash.

Specifically, I think we should:
- set `lastBuildParentHash` (unconditionally) in `handleGenerateBlock`
- set `lastBuildTime` (unconditionally) in `handleGenerateBlock`
- compare the preferred block hash to the `lastBuildParentHash` in `WaitForEvent`, if they are equal perform the delay based on `lastBuildTime`. Otherwise do not introduce any artificial delay.

StephenButtolph (commented):
code lgtm, should we add a test for this?

plugin/evm/vm_test.go
```diff
@@ -1616,64 +1601,58 @@ func TestWaitForEvent(t *testing.T) {
...
 				wg.Add(1)
 
 				go func() {
 					defer wg.Done()
+					timeStart := time.Now()
 					msg, err := vm.WaitForEvent(context.Background())
+					timeSinceStart := time.Since(timeStart)
 					assert.NoError(t, err)
 					assert.Equal(t, commonEng.PendingTxs, msg)
+					assert.Less(t, timeSinceStart, 50*time.Millisecond)
```
alarso16, line 1629: What is this checking? Also, it's unclear to me why this can't be done synchronously

plugin/evm/vm_test.go
```diff
@@ -1584,23 +1584,8 @@ func TestWaitForEvent(t *testing.T) {
...
-				}
-
-				blk, err := vm.BuildBlock(context.Background())
-				require.NoError(t, err)
-
-				require.NoError(t, blk.Verify(context.Background()))
-
-				require.NoError(t, vm.SetPreference(context.Background(), blk.ID()))
-
-				require.NoError(t, blk.Accept(context.Background()))
```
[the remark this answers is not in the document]
yacovm: > the test actually ensuring mempool is emptied after accepting a block first, and then checking the builder behaviour. However it's flaky because mempool is emptied async.

Wait, so after `Accept()` is called, the mempool may still contain the transactions that exist in the block that was just accepted?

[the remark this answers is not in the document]
StephenButtolph: Yes it seems so.

StephenButtolph (approved):

PR #1303, opened 2025-09-30
style: Use require.ErrorIs whenever possible

alarso16:
## Why this should be merged

Works to follow avalanchego's linter whenever possible. Some of the ErrorContains really tested geth code, making it difficult to remove.

## How this works

Uses sentinel errors and nolint clauses.

## How this was tested

Existing UT

## Need to be documented?

No

No


plugin/evm/atomic/vm/import_tx_test.go
```diff
@@ -234,7 +232,7 @@ func TestImportTxVerify(t *testing.T) {
 			},
 			ctx:         ctx,
 			rules:       vmtest.ForkToRules(upgradetest.NoUpgrades),
-			expectedErr: "atomic input failed verification",
+			expectedErr: avax.ErrNilTransferableInput,
```
alarso16, line 235: This is the specific error we were checking, but I'm not sure it's good practice to rely on an error "so far away"

plugin/evm/wrapped_block.go
```diff
@@ -482,12 +482,12 @@ func (b *wrappedBlock) syntacticVerify() error {
 		}
 	} else {
 		switch {
+		case ethHeader.ParentBeaconRoot != nil:
```
alarso16, line 485: This is only moved because the test fails for a different error than what it was intending to check.

JonathanOppenheimer:
> > Just a question (and I'll do an actual review) - I remember talking about the `vmerrors`in terms of the uplift - do we want to use that package for the sentinel errors declared? Do we want to get rid of the package entirely?
> 
> That package is used to share errors between packages to avoid circular dependencies - for the errors I added, I didn't actually need to move them, but I definitely could. I think the package is necessary for the existing errors though

I see - I didn't realize it was just for errors that would create circular dependencies!

plugin/evm/vm_test.go
```diff
@@ -507,9 +507,9 @@ func testReorgProtection(t *testing.T, scheme string) {
 	// should NEVER happen. However, the VM defends against this
 	// just in case.
 	err = vm1.SetPreference(context.Background(), vm1BlkC.ID())
-	require.ErrorContains(err, "cannot orphan finalized block")
+	require.ErrorContains(err, "cannot orphan finalized block") //nolint:forbidigo // uses upstream code
 	err = vm1BlkC.Accept(context.Background())
-	require.ErrorContains(err, "expected accepted block to have parent")
+	require.ErrorContains(err, "expected accepted block to have parent") //nolint:forbidigo // uses upstream code
```
JonathanOppenheimer, line 512: saw this a couple times - wdym it uses upstream code?

JonathanOppenheimer (approved):

plugin/evm/vm_test.go
alarso16: geth code. I don't want to change the errors returned from geth code (e.g. exporting an unexported error), so I just nolint-ed

PR #1306, opened 2025-09-30
Gstuart/warp gas costs granite

geoff-vball:
## Why this should be merged
Updates warp gas fees for granite upgrade. (Final numbers still TBD).

### DIFF
--- a/params/extras/network_upgrades.go
+++ b/params/extras/network_upgrades.go
@@ -264,6 +264,22 @@ type AvalancheRules struct {
 	IsGranite                                                                           bool
 }
 
+func (a *AvalancheRules) IsDurangoActivated() bool {
+	return a.IsDurango
+}
+
+func (a *AvalancheRules) IsEtnaActivated() bool {
+	return a.IsEtna
+}
+
+func (a *AvalancheRules) IsFortunaActivated() bool {
+	return a.IsFortuna
+}
+
+func (a *AvalancheRules) IsGraniteActivated() bool {
+	return a.IsGranite
+}
+
 func (n *NetworkUpgrades) GetAvalancheRules(timestamp uint64) AvalancheRules {
 	return AvalancheRules{
 		IsApricotPhase1:     n.IsApricotPhase1(timestamp),
--- a/precompile/contracts/warp/contract.go
+++ b/precompile/contracts/warp/contract.go
@@ -31,11 +31,16 @@ const (
 	SendWarpMessageGasCost uint64 = contract.LogGas + 3*contract.LogTopicGas + AddWarpMessageGasCost + contract.WriteGasCostPerSlot
 	// SendWarpMessageGasCostPerByte cost accounts for producing a signed message of a given size
 	SendWarpMessageGasCostPerByte uint64 = contract.LogDataGas
-
-	GasCostPerWarpSigner            uint64 = 500
-	GasCostPerWarpMessageChunk      uint64 = 3_200
-	GasCostPerSignatureVerification uint64 = 200_000
 )
+
+type WarpGasConfig struct {
+	// Gas cost per warp signer in the validator set
+	PerWarpSigner uint64
+	// Gas cost per chunk of the warp message (each chunk is 128 bytes)
+	PerWarpMessageChunk uint64
+	// Gas cost to verify a warp message signature
+	PerSignatureVerification uint64
+}
 
 var (
 	errInvalidSendInput  = errors.New("invalid sendWarpMessage input")
@@ -342,4 +347,19 @@ func createWarpPrecompile() contract.StatefulPrecompiledContract {
 		panic(err)
 	}
 	return statefulContract
+}
+
+func WarpPredicateGasConfig(isGraniteActivated bool) WarpGasConfig {
+	if isGraniteActivated {
+		return WarpGasConfig{
+			PerWarpSigner:            500,
+			PerWarpMessageChunk:      3_200,
+			PerSignatureVerification: 200_000,
+		}
+	}
+	return WarpGasConfig{
+		PerWarpSigner:            250,
+		PerWarpMessageChunk:      3_200,
+		PerSignatureVerification: 100_000,
+	}
 }
--- a/precompile/contracts/warp/predicate_test.go
+++ b/precompile/contracts/warp/predicate_test.go
@@ -202,6 +202,7 @@ func createSnowCtx(tb testing.TB, validatorRanges []validatorRange) *snow.Contex
 }
 
 func createValidPredicateTest(snowCtx *snow.Context, numKeys uint64, predicate predicate.Predicate) precompiletest.PredicateTest {
+	gasConfig := WarpPredicateGasConfig(false)
 	return precompiletest.PredicateTest{
 		Config: NewDefaultConfig(utils.NewUint64(0)),
 		PredicateContext: &precompileconfig.PredicateContext{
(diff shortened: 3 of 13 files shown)


precompile/precompileconfig/config.go
```diff
@@ -77,4 +77,14 @@ type Accepter interface {
 type ChainConfig interface {
 	// IsDurango returns true if the time is after Durango.
 	IsDurango(time uint64) bool
+	// IsGranite returns true if the time is after Granite.
+	IsGranite(time uint64) bool
```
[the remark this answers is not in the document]
geoff-vball: It's used here https://github.com/ava-labs/coreth/blob/gstuart/warp-gas-costs-granite/precompile/contract/utils.go#L63

precompile/contracts/warp/contract.go
```diff
@@ -343,3 +348,18 @@ func createWarpPrecompile() contract.StatefulPrecompiledContract {
...
 	return statefulContract
 }
+
+func GetWarpPredicateGasConfig(isGraniteActivated bool) GasConfig {
+	if isGraniteActivated {
+		return GasConfig{
+			PerWarpSigner:            250,
+			PerWarpMessageChunk:      3_200,
+			PerSignatureVerification: 100_000,
+		}
```
ceyonur, line 358: nit: we can just define them as constants just for a better visibility

precompile/contracts/warp/contract.go
```diff
@@ -343,3 +348,18 @@ func createWarpPrecompile() contract.StatefulPrecompiledContract {
 	}
 	return statefulContract
 }
+
+func GetWarpPredicateGasConfig(isGraniteActivated bool) GasConfig {
```
ARR4N, line 352:
```suggestion
func CurrentGasConfig(rules precompileconfig.Rules) GasConfig {
```

Rationale re readability / maintainability:
- The "get" prefix doesn't communicate anything of value; "current" tells the reader that the function changes what it returns over time.
- We're already in the `warp` package so `warp.GetWarp...` is jarring when called from outside, and unnecessary when called from inside the package.
- "Predicate" could possibly remain, but is unnecessary and may not always be where this is used (e.g. under SAE the gas charge is the same but it's done in a different place).
- Raw booleans lend themselves to hard-to-read call sites, often requiring comments like `config := CurrentGasConfig(false /*isGranite*/)` (e.g. your first change to `TestGetVerifiedWarpMessage` is opaque). Across geth/libevm, `Rules` are the point-in-time carriers of these booleans, so accepting them as the argument makes code self-explanatory and less brittle if changed (e.g. in a future fork where there would be two booleans).

ARR4N (commented):
Small point on readability, and I see it's pretty similar to another comment that Stephen left. Otherwise LGTM

precompile/contracts/warp/contract.go
```diff
@@ -31,12 +32,17 @@ const (
...
-	GasCostPerWarpMessageChunk      uint64 = 3_200
-	GasCostPerSignatureVerification uint64 = 200_000
 )
 
+type GasConfig struct {
+	// Gas cost per warp signer in the validator set
+	PerWarpSigner uint64
+	// Gas cost per chunk of the warp message (each chunk is 128 bytes)
+	PerWarpMessageChunk uint64
+	// Gas cost to verify a warp message signature
```
michaelkaplan13, line 42:
```suggestion
	// Gas cost to verify a BLS signature
```

I think of the "warp message signature" as the BLS aggregate signature and bit vector together, as opposed to this just being the fixed cost to verify a BLS signature given a public key.

params/hooks_libevm.go
```diff
@@ -195,6 +195,11 @@ func (a accessibleState) GetChainConfig() precompileconfig.ChainConfig {
 	return GetExtra(a.env.ChainConfig())
 }
 
+func (a accessibleState) GetRules() precompileconfig.Rules {
+	extra := GetExtra(a.GetPrecompileEnv().ChainConfig())
```
ceyonur, line 199: nit: can we rename this to `chainConfigExtra` as there are actual `extras` package which can be confusing

params/hooks_libevm.go
```diff
@@ -195,6 +195,11 @@ func (a accessibleState) GetChainConfig() precompileconfig.ChainConfig {
 	return GetExtra(a.env.ChainConfig())
 }
 
+func (a accessibleState) GetRules() precompileconfig.Rules {
```
ceyonur, line 198: Thanks for refactoring this.

params/extras/network_upgrades.go
```diff
@@ -264,6 +264,10 @@ type AvalancheRules struct {
 	IsGranite                                                                           bool
 }
 
+func (a AvalancheRules) IsGraniteActivated() bool {
```
ceyonur, line 267: can we add a comment why this is required (to make it clear this is called from precompile environment etc)

ceyonur (commented):
Overall lgtm, just few nits for comments and renaming etc

JonathanOppenheimer:
I think all of the comments except the two I did not resolve, have been addressed.

precompile/precompiletest/test_precompile.go
```diff
@@ -91,6 +95,10 @@ func (test PrecompileTest) setup(t testing.TB, module modules.Module, state *tes
 	t.Helper()
 	contractAddress := module.Address
 
+	if test.Rules == nil {
+		test.Rules = extras.AvalancheRules{IsGranite: true}
```
ceyonur, line 99: why we set this here? If this is an absolute requirement we should provide the correct `AvalancheRules` like everything is activated up to `IsGranite` (i.e `IsFortuna, IsDurango: true`). 

I just put up a quick [PR](https://github.com/ava-labs/coreth/pull/1327), you can use the helper from `extrastest` to get the rules with the fork.

precompile/contracts/warp/predicate_test.go
```diff
@@ -92,6 +96,9 @@ func init() {
 		panic(err)
 	}
 
+	graniteRules = extras.AvalancheRules{IsGranite: true}
```
ceyonur, line 99: same comment to use the helper for full valid rule set

precompile/contracts/warp/predicate_test.go
```diff
@@ -92,6 +96,9 @@ func init() {
...
+	graniteGasConfig = CurrentGasConfig(graniteRules)
```
ceyonur, line 100: can we set this in var?

precompile/contracts/warp/predicate_test.go
ceyonur: also we should probably set this in var block if possible

ARR4N: > also we should probably set this in var block if possible

+1 there's no need to delay its definition

precompile/contracts/warp/predicate_test.go
```diff
@@ -212,7 +233,8 @@ func createValidPredicateTest(snowCtx *snow.Context, numKeys uint64, predicate p
 			},
 		},
 		Predicate:   predicate,
-		Gas:         gasConfig.PerSignatureVerification + uint64(len(predicate))*gasConfig.PerWarpMessageChunk + numKeys*gasConfig.PerWarpSigner,
+		Rules:       graniteRules,
+		Gas:         graniteGasConfig.PerSignatureVerification + uint64(len(predicate))*graniteGasConfig.PerWarpMessageChunk + numKeys*graniteGasConfig.PerWarpSigner,
```
ARR4N, line 237: Please abstract a helper function to calculate this. Something like:

```go
func predicateGasCost(c *GasConfig, chunks, signers int) uint64

// or

func (c *GasConfig) predicateGasCost(chunks, signers int) uint64
```

I recommend accepting `int` instead of `uint64` so you can pass `len(x)` at the call sites, which is where this is otherwise messy. The function name may not be appropriate as I'm not intimately familiar with the rest of the code.

ARR4N (commented):
LGTM pending requested changes that I don't need to re-review if you agree with them. If not then please let me know so we can discuss.

I'm not giving an official approval as I'm not familiar enough with this part of the code to have it merged without another set of eyes.

PR #1311, opened 2025-10-02
Add warp test to verify equivalent gas consumed

StephenButtolph:
## Why this should be merged

As part of ACP-181, we will be caching all validator sets at the current epoch height.

As part of this we will be changing the canonical representation of a validator set with a total weight that overflows uint64 to be the empty set rather than erroring during construction.

This adds an equivalence test to ensure this a not a breaking change.

## How this works

Adds a test which verifies that the gas consumed is equal for both cases (and that both cases error).

## How this was tested

It is a test

## Need to be documented?

no

no

### DIFF
--- a/precompile/contracts/warp/predicate_test.go
+++ b/precompile/contracts/warp/predicate_test.go
@@ -619,7 +619,7 @@ func TestWarpNoValidatorsAndOverflowUseSameGas(t *testing.T) {
 	noValidators := precompiletest.PredicateTest{
 		Config: NewConfig(utils.NewUint64(0), 0, false),
 		PredicateContext: &precompileconfig.PredicateContext{
-			SnowCtx: createSnowCtx(t, nil),
+			SnowCtx: createSnowCtx(t, nil /*=validators*/), // No validators in state
 			ProposerVMBlockCtx: &block.Context{
 				PChainHeight: 1,
 			},
@@ -635,7 +635,7 @@ func TestWarpNoValidatorsAndOverflowUseSameGas(t *testing.T) {
 			SnowCtx: createSnowCtx(t, []validatorRange{
 				{
 					start:     0,
-					end:       100,
+					end:       2, // Generate two validators each with max weight to force overflow
 					weight:    math.MaxUint64,
 					publicKey: true,
 				},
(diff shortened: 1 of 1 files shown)


precompile/contracts/warp/predicate_test.go
```diff
@@ -136,18 +138,34 @@ func newTestValidator() *testValidator {
 // createWarpMessage constructs a signed warp message using the global variable [unsignedMsg]
 // and the first [numKeys] signatures from [blsSignatures]
 func createWarpMessage(numKeys int) *avalancheWarp.Message {
```
alarso16, line 140: Probably out of scope, but instead of a panic, why not take `*testing.T`?

precompile/contracts/warp/predicate_test.go
```diff
@@ -594,6 +612,49 @@ func TestWarpSignatureWeightsNonDefaultQuorumNumerator(t *testing.T) {
...
+					weight:    math.MaxUint64,
+					publicKey: true,
+				},
+			}),
+			ProposerVMBlockCtx: &block.Context{
+				PChainHeight: 1,
+			},
+		},
+		Predicate:   pred,
+		Gas:         GasCostPerSignatureVerification + uint64(len(pred))*GasCostPerWarpMessageChunk, // Gas cost should be equal to [TestWarpNoValidators]
```
alarso16, line 648:
```suggestion
		Gas:       expectedGas
```

precompile/contracts/warp/predicate_test.go
StephenButtolph: The whole testing package needs to be re-written imo

precompile/contracts/warp/predicate_test.go
```diff
@@ -136,18 +138,34 @@ func newTestValidator() *testValidator {
...
+
+	var sig *bls.Signature
+	if numKeys > 0 {
+		aggregateSignature, err := bls.AggregateSignatures(blsSignatures[0:numKeys])
+		if err != nil {
+			panic(err)
+		}
+		sig = aggregateSignature
+	} else {
+		// populate with a random signature
```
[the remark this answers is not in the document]
StephenButtolph: Changed the comment to explain why it exists rather than what it is doing.

PR #1313, opened 2025-10-02
Remove secondary validator override

StephenButtolph:
## Why this should be merged

Currently `warp_getMessageAggregateSignature` and `warp_getBlockAggregateSignature` incorrectly override the `subnetID` twice.

Specifically, the intended behavior is:
- If the `subnetID` is not overridden, use this chain's validator set
  - This is the "normal" case as this chain is sending the message.
- If the `subnetID` is overridden, use the validator set requested by the caller
  - This allows the C-chain's APIs to be used to generate a signature for a subnet which is using their own validator set to verify the message.

This is almost what the logic does currently, but there is a weird edge case. If the `subnetID` is overridden to use the primary network's validator set and this chain has enabled the optimization to use the local validator set for the primary network, then the `subnetID` is overridden for a second time and the chain's validator set will be used.

We should not override the `subnetID` based on this chain's configuration. If the caller overrides the `subnetID`, that override should be honored.

This edge case has likely never been hit, because no one would reasonably override the validator set to the primary network, as that would just make the message more expensive to issue.

## How this works

Removes the second override along with the resulting dead code.

## How this was tested

Existing CI still passes. I think this API is planned on being removed at some point in favor of the standalone relayer implementation, so I didn't add a regression test here.

## Need to be documented?

No.

Probably not? I doubt anyone has hit this.

### DIFF
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -1077,7 +1077,7 @@ func (vm *VM) CreateHandlers(context.Context) (map[string]http.Handler, error) {
 		warpSDKClient := vm.Network.NewClient(p2p.SignatureRequestHandlerID)
 		signatureAggregator := acp118.NewSignatureAggregator(vm.ctx.Log, warpSDKClient)
 
-		if err := handler.RegisterName("warp", warp.NewAPI(vm.ctx, vm.warpBackend, signatureAggregator, vm.requirePrimaryNetworkSigners)); err != nil {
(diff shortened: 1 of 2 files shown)


ARR4N (commented):
~You're gonna hate me for asking, but why doesn't this change break any existing tests nor require any new ones to cover it? It would also really help understand said edge case because I'm confused by the description (granted I'm not familiar with this code and it's late here).~

EDIT: I missed the description about why you're not adding a regression test.

PR #1315, opened 2025-10-02
chore: remove deprecated genesis types

JonathanOppenheimer:
## Why this should be merged
These types are explicitly deprecated and just serve as a type alias. 

## How this was tested
Existing UTs

## Need to be documented?
No

No


alarso16 (approved):

ARR4N:
This is just upstream `geth` code. Although it's unlikely that `core` will undergo another `geth` sync given SAE, I still don't think we should be adding further diffs.

PR #1317, opened 2025-10-02
chore: Use go 1.24's 'go tool' support for CLI tooling

maru-ava:
## Why this should be merged

This change mirrors a change already made to avalanchego and avoids the problem of a cached version of golangci-lint being incompatible with a new version of golang when upgrading the golang version between minor versions.

## How this was tested

CI

## Need to be documented?

Yes - updated `CONTRIBUTING.md`

No


.github/CONTRIBUTING.md
```diff
@@ -40,13 +40,13 @@ Mocks are auto-generated using [mockgen](https://pkg.go.dev/go.uber.org/mock/moc
 * To **re-generate all mocks**, use the command below from the root of the project:
 
     ```sh
-    go generate -run "go.uber.org/mock/mockgen" ./...
+    go generate -run mockgen ./...
```
ARR4N, line 43: You couldn't just remove it entirely? 😛

ARR4N (approved):

tools/legacy-golangci-lint.mod
```diff
@@ -0,0 +1,197 @@
+module github.com/ava-labs/coreth/tools/legacy-golangci-lint
```
JonathanOppenheimer, line 1: What are the legacy files for?

JonathanOppenheimer (approved):

tools/legacy-golangci-lint.mod
JonathanOppenheimer: Ah I see - sorry I missed that!

StephenButtolph (approved):

PR #1340, opened 2025-10-14
feat: Add block database

DracoLi:
## Why this should be merged

Introduces a new `ethdb.Database` wrapper that stores block headers, bodies, and receipts in separate height-indexed databases, moving away from storing all block data in the main key-value database.

This should reduce database compactions in the key-value store, leading to improved performance for all existing read/write operations.

There's a migrator that will migrate block data from the key-value database to the block database in the background.

See https://github.com/ava-labs/avalanchego/blob/master/x/blockdb/README.md on how the height-indexed database works.

## How this works

The `BlockDatabase` wrapper intercepts database operations and routes block-related keys (headers, bodies, receipts) to dedicated height-indexed databases while passing all other operations through to the underlying key-value database.

Block headers and bodies are stored together in a single entry, requiring the system to wait for both components before writing to the block database.

Genesis block (block 0) is not stored in the block database to simplify implementation.

During the migration period, the wrapper falls back to the key-value database for any data not yet migrated, ensuring zero downtime and data availability throughout the transition.

## How this was tested

Unit tests

### Benchmark

Benchmarked using the Reexecute Benchmark across multiple block ranges:
- Block 100 - 1m
- Block 1.3m - 2.3m
- Block 10m - 10.5m
- Block ~70m - 70.1m (node state synced at height 70m)

### Mainnet Nodes

Default Mainnet node running blockdb for +4 weeks without issues

### Migration Testing

The following scenarios are tested on Mainnet nodes:
- Migrated mainnet c-chain leveldb state at 1M and 10M block heights
- Migrated an archive node with +60M blocks starting from genesis
- Migrated a state synced node at height 70,271,997.

### Block Data Verification

Created a script that took two evm rpc endpoints and periodically check the block and receipts data returned by those two endpoints for blocks from the last checked height to the latest block height to ensure they are the same.
This script is running 24/7 and is monitoring a blockdb enabled node and a node running `v1.13.5`.

### DIFF
--- a/plugin/evm/database/blockdb/database.go
+++ b/plugin/evm/database/blockdb/database.go
@@ -1,0 +1,464 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package blockdb
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"path/filepath"
+
+	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/database/heightindexdb/meterdb"
+	"github.com/ava-labs/avalanchego/database/prefixdb"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/avalanchego/x/blockdb"
+	"github.com/ava-labs/libevm/common"
+	"github.com/ava-labs/libevm/ethdb"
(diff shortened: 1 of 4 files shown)


JonathanOppenheimer:
@DracoLi as this is completely contained code, I think it make more sense to open this PR in AvalancheGo under `/vms/evm` -- we should try to avoid introducing new code to the EVM repositories while working towards a monorepo unless necessary.

PR #1342, opened 2025-10-15
use average for block tips after Granite

ceyonur:
## Why this should be merged

We have removed block gas cost with ACP-226 so we cannot return any estimation based on block gas costs. 

## How this works

* Changed default percentile from 60 to 40
* Before each block had a estimated tip which then would form a dataset of tips. The `SuggesTipCap` then takes the percentile of those estimated tips. Now we're forming a dataset from all txs and their tips and then taking the percentile. 

## How this was tested

Added UTs, simulated mainnet tips. 

The current mainnet tx medians are converging to the 1000 wei e.g:

```
{"summaryOverBlocks":{"p60MedianTipGwei":"0.000001","p40MedianTipGwei":"0.000001","p80MedianTipGwei":"0.01","p90MedianTipGwei":"0.2","eth_maxPriorityFeePerGas_Gwei":"0.107162306"}}
```

## Need to be documented?

Changed documentation slightly

yes


eth/gasprice/fee_info_provider.go
```diff
@@ -89,33 +85,23 @@ func newFeeInfoProvider(backend OracleBackend, minGasUsed uint64, size int) (*fe
...
 
-	totalGasUsed := new(big.Int).SetUint64(header.GasUsed)
-	if used := customtypes.GetHeaderExtra(header).ExtDataGasUsed; used != nil {
-		totalGasUsed.Add(totalGasUsed, used)
-	}
-	minGasUsed := new(big.Int).SetUint64(f.minGasUsed)
-
-	// Don't bias the estimate with blocks containing a limited number of transactions paying to
-	// expedite block production.
+	var tips []*big.Int
```
alarso16: nit: combine var block

eth/gasprice/gasprice_test.go
```diff
@@ -257,223 +245,135 @@ func testGenBlock(t *testing.T, tip int64, numTx int) func(int, *core.BlockGen)
...
+			numBlocks:   3,
+			genBlock:    testGenBlockWithTips(t, []int64{10, 20, 30}),
+			expectedTip: big.NewInt(20 * params.GWei),
+		},
+		{
+			name:        "four_transactions_with_even_count_tips",
+			numBlocks:   3,
+			genBlock:    testGenBlockWithTips(t, []int64{10, 20, 30, 40}),
+			expectedTip: big.NewInt(30 * params.GWei),
+		},
```
StephenButtolph: It feels a bit weird that we aren't testing against the default config here... but I think that's ok.

PR #1347, opened 2025-10-16
chore: align `PointerTo` functions

JonathanOppenheimer:
## Why this should be merged
There were a number of various pointer functions dispersed throughout the codebase. This PR aligns them all to a single implementation. 

This PR closes https://github.com/ava-labs/avalanchego/issues/4567.

### DIFF
--- a/rpc/websocket_test.go
+++ b/rpc/websocket_test.go
@@ -37,7 +37,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/ava-labs/coreth/utils"
 	"github.com/gorilla/websocket"
 )
 


eth/tracers/api_test.go
```diff
@@ -299,8 +300,6 @@ func testTraceCall(t *testing.T, scheme string) {
 		}
 	})
 
-	uintPtr := func(i int) *hexutil.Uint { x := hexutil.Uint(i); return &x }
```
JonathanOppenheimer: I can revert these changes if desired to reduce the `geth` difference.

utils/numbers.go
```diff
@@ -8,11 +8,12 @@ import (
 	"time"
 )
 
-func NewUint64(val uint64) *uint64 { return &val }
+// PointerTo returns a pointer to the given value.
+func PointerTo[T any](x T) *T { return &x }
```
JonathanOppenheimer: The function should be moved here, rather than `utilstest` as it not only used in tests.

eth/tracers/api_test.go
JonathanOppenheimer: reverted!

JonathanOppenheimer:
I guess maybe one benefit of doing it this way, is the linter now catches needless typing errors.

PR #812, opened 2025-02-19
Merge master (2e4e6bd) into libevm (2)

darioush:
Step 2. merge master [2e4e6bd](https://github.com/ava-labs/coreth/commit/2e4e6bd85132e7c655775e1b62a6397e0270cdb5)
Step 3. Make coreth & avalanchego mutually depend on each other via replace...
Step 4. Push one of them (eg, coreth)
Step 5. now go get coreth in avalanchego
Step 5. now go get avalanchego in coreth & set versions.sh.

PR #1348, opened 2025-10-17
Convert import script to a go test 

### LINKED ISSUE #813: Implement `lint_allowed_eth_imports.sh` in Go
The shell script `./scripts/lint_allowed_eth_imports.sh` is getting rather [complex and hard to understand](https://github.com/ava-labs/coreth/pull/812/commits/0f5f211acc17d26cf2b09c875e31b1ff031117e9), we should migrate it to use Go as a test, especially now that we have [`getDependencies` code](https://github.com/ava-labs/coreth/blob/0103ad89b6e6323f7b7a9b847e5c635bb832a4f3/plugin/evm/imports_test.go#L16) which can easily be adapted for this use case.

JonathanOppenheimer:
## Why this should be merged
This closes #813: 
> The shell script ./scripts/lint_allowed_eth_imports.sh is getting rather [complex and hard to understand](https://github.com/ava-labs/coreth/pull/812/commits/0f5f211acc17d26cf2b09c875e31b1ff031117e9), we should migrate it to use Go as a test, especially now that we have [getDependencies code](https://github.com/ava-labs/coreth/blob/0103ad89b6e6323f7b7a9b847e5c635bb832a4f3/plugin/evm/imports_test.go#L16) which can easily be adapted for this use case.

I was unable to adapt the `getDependencies()` function to work properly with ignoring `_` imports and imports that start with `eth`, but I did massively improve the test, and this now catches some imports that the shell script did not. 

I will remove `getDependencies()` in the next PR anyway. 

Sample: 

Screenshot 2025-10-20 at 11 58 32 AM

## How this was tested
The new `TestLibevmImportsAreAllowed` test 

## Need to be documented?
No

No

### DIFF
--- a/plugin/evm/imports_test.go
+++ b/plugin/evm/imports_test.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/ava-labs/avalanchego/utils/set"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/tools/go/packages"
 )
@@ -96,11 +97,10 @@ func TestLibevmImportsAreAllowed(t *testing.T) {
 	foundImports, err := findFilteredLibevmImportsWithFiles("../..")
 	require.NoError(t, err, "Failed to find libevm imports")
 
-	// Check for any imports not in the allowed list and build detailed error message
-	var disallowedImports []string
+	var disallowedImports set.Set[string]
 	for importPath := range foundImports {
-		if _, allowed := allowedPackages[importPath]; !allowed {
-			disallowedImports = append(disallowedImports, importPath)
+		if !allowedPackages.Contains(importPath) {
+			disallowedImports.Add(importPath)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


JonathanOppenheimer:
This is blocked by #1349 and #1350

plugin/evm/imports_test.go
```diff
@@ -76,3 +85,120 @@ func TestMustNotImport(t *testing.T) {
...
+	var extraImports []string
+	for importPath := range foundImports {
+		if _, allowed := allowedPackages[importPath]; !allowed {
+			extraImports = append(extraImports, importPath)
+		}
+	}
+
+	slices.Sort(extraImports)
+
+	if len(extraImports) > 0 {
```
alarso16: You should exit early

plugin/evm/imports_test.go
```diff
@@ -76,3 +85,120 @@ func TestMustNotImport(t *testing.T) {
...
+		}
+
+		// Skip generated files, main_test.go, and tempextrastest directory
+		filename := filepath.Base(path)
+		if strings.HasPrefix(filename, "gen_") || strings.Contains(path, "core/main_test.go") || strings.Contains(path, "tempextrastest/") {
+			return nil
+		}
+
+		node, err := parser.ParseFile(token.NewFileSet(), path, nil, parser.ParseComments)
+		if err != nil {
```
alarso16: This seems bad... why not return an error?

JonathanOppenheimer: that would be more concerning if there were.

plugin/evm/imports_test.go
```diff
@@ -76,3 +85,124 @@ func TestMustNotImport(t *testing.T) {
...
+
+	// Check for any imports not in the allowed list and build detailed error message
+	var disallowedImports []string
+	for importPath := range foundImports {
+		if _, allowed := allowedPackages[importPath]; !allowed {
+			disallowedImports = append(disallowedImports, importPath)
+		}
+	}
+
+	slices.Sort(disallowedImports)
```
alarso16: nit: no point in sorting an empty list. Can you move this after the early return?

alarso16 (commented):
A couple nits

JonathanOppenheimer:
Thanks for the review @alarso16 !

alarso16 (commented):
Looks good to me, but might want to think about a good location for this test in a monorepo

PR #1349, opened 2025-10-17
Remove `libevm/psuedo` import

JonathanOppenheimer:
## Why this should be merged
This libevm import is not allowed and should not be used. 

## How this was tested
Existing UT

## Need to be documented?
No

No

### DIFF
--- a/plugin/evm/customtypes/state_account_ext.go
+++ b/plugin/evm/customtypes/state_account_ext.go
@@ -1,8 +1,9 @@
+// Copyright (C) 2019-2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
 package customtypes
 
-import (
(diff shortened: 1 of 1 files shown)


plugin/evm/customtypes/state_account_ext.go
```diff
@@ -3,15 +3,11 @@
...
-	"github.com/ava-labs/libevm/libevm/pseudo"
-
-	ethtypes "github.com/ava-labs/libevm/core/types"
-)
+import ethtypes "github.com/ava-labs/libevm/core/types"
 
 type isMultiCoin bool
 
-var IsMultiCoinPayloads pseudo.Accessor[ethtypes.StateOrSlimAccount, isMultiCoin]
+var IsMultiCoinPayloads = extras.StateAccount
```
ARR4N: This suggests to a reader that the value is ready for use, which it's not.

PR #1360, opened 2025-10-23
test: Fix mempool flakes

### LINKED ISSUE #1356: Fix mempool test flakes
There are two tests on the mempool that are flaky:
```
TestMempoolAtmTxsAppGossipHandlingDiscardedTx
TestMempoolEthTxsAppGossipHandling
```

alarso16:
## Why this should be merged

Closes #1356 as pre-req for monorepo. These tests were hazy in scope anyway and didn't accurately test what it intended to.

## How this works

It removes one of the tests, and combines two of the tests utilizing the `AppGossip` function instead.

## How this was tested

UT

## Need to be documented?

No

No

### DIFF
--- a/scripts/known_flakes.txt
+++ b/scripts/known_flakes.txt
@@ -1,4 +1,3 @@
 TestResumeSyncAccountsTrieInterrupted
 TestResyncNewRootAfterDeletes
 TestTransactionSkipIndexing
-


plugin/evm/gossiper_eth_gossiping_test.go
```diff
@@ -113,26 +117,9 @@ func TestMempoolEthTxsAppGossipHandling(t *testing.T) {
 	// Txs must be submitted over the API to be included in push gossip.
 	// (i.e., txs received via p2p are not included in push gossip)
 	require.NoError(vm.eth.APIBackend.SendTx(context.Background(), tx))
+	require.NoError(vm.ethTxPushGossiper.Get().Gossip(context.Background()))
+	statusLock.Lock()
```
geoff-vball: This lock _is_ actually doing something, but it's only ensuring that there are no calls to either `tvm.AppSender.SendAppRequestF` or `tvm.AppSender.SendAppGossipF` between the two checks here, which I'm not sure if that's the desired effect.

plugin/evm/gossiper_eth_gossiping_test.go
```diff
@@ -93,17 +93,21 @@ func TestMempoolEthTxsAppGossipHandling(t *testing.T) {
...
 	var (
-		wg          sync.WaitGroup
 		txRequested bool
+		txResponded bool
+		statusLock  sync.Mutex
 	)
 	tvm.AppSender.CantSendAppGossip = false
 	tvm.AppSender.SendAppRequestF = func(context.Context, set.Set[ids.NodeID], uint32, []byte) error {
+		statusLock.Lock()
+		defer statusLock.Unlock()
```
geoff-vball: The point of this test is to check that this function _isn't_ called, correct? Can you just put a `require.FailNow()` inside this function?

alarso16: Sure!

plugin/evm/gossip_test.go
```diff
@@ -110,3 +112,26 @@ func setupPoolWithConfig(t *testing.T, config *params.ChainConfig, fundedAddress
 
 	return txPool
 }
+
+func getValidEthTxs(key *ecdsa.PrivateKey, count int, gasPrice *big.Int) []*types.Transaction {
```
alarso16: Unchanged helper function

JonathanOppenheimer: this only has a single usage - is it worth inlining this? esp with hardcoded numbers like `gasLimit = 37000`

alarso16: I did consider it, but I worried about readability of the other test.

JonathanOppenheimer (approved):

scripts/known_flakes.txt
```diff
@@ -1,6 +1,4 @@
-TestMempoolAtmTxsAppGossipHandlingDiscardedTx
-TestMempoolEthTxsAppGossipHandling
 TestResumeSyncAccountsTrieInterrupted
 TestResyncNewRootAfterDeletes
 TestTransactionSkipIndexing
-TestUpdatedKeyfileContents
+
```
StephenButtolph:
```suggestion
```

plugin/evm/atomic/vm/tx_gossip_test.go
```diff
@@ -311,18 +311,61 @@ func TestAtomicTxPushGossipInbound(t *testing.T) {
...
-	require.NoError(err)
+	inboundGossipMsg := buildAtomicPushGossip(t, tx)
+	require.NoError(vm.AppGossip(ctx, ids.EmptyNodeID, inboundGossipMsg))
+	require.True(vm.AtomicMempool.Has(tx.ID()))
+}
+
+// Tests that conflicting txs are handled, based on whether the original tx is
+// valid or not.
+func TestAtomicTxPushGossipInboundConflicting(t *testing.T) {
+	require := require.New(t)
```
powerslider: You don't need this. The only thing it achieves is implicitly passing `testing.T` to the concrete `require` assertion call, but it adds unnecessary noise.

PR #1361, opened 2025-10-24
test: Fix state sync flakes

### LINKED ISSUE #1357: Fix state sync flaky tests
There are two flaky state sync tests:
```
TestResumeSyncAccountsTrieInterrupted
TestResyncNewRootAfterDeletes
```

alarso16:
## Why this should be merged

Since the state syncer can have multiple workers, there may be multiple in flight requests when it should start returning errors. Therefore, the number may be atomically increased more than twice.

For the other test, I haven't been able to replicate it locally. In order to see it flake (if it still does), I think we should remove it from the list of flakes

Closes #1357

## How this works

Provides lower bound.

## How this was tested

Unable to reproduce, likely needs small machine.

## Need to be documented?

No.

No.

### DIFF
--- a/sync/statesync/sync_test.go
+++ b/sync/statesync/sync_test.go
@@ -260,7 +260,7 @@ func TestResumeSyncAccountsTrieInterrupted(t *testing.T) {
 		GetLeafsIntercept: intercept.getLeafsIntercept,
 	})
 
-	require.LessOrEqual(t, uint32(2), intercept.numRequests.Load())
+	require.GreaterOrEqual(t, intercept.numRequests.Load(), uint32(2))
 
 	testSync(t, syncTest{
 		prepareForTest: func(*testing.T, *rand.Rand) (ethdb.Database, ethdb.Database, *triedb.Database, common.Hash) {


powerslider (approved):

sync/statesync/sync_test.go
```diff
@@ -260,7 +260,7 @@ func TestResumeSyncAccountsTrieInterrupted(t *testing.T) {
 		GetLeafsIntercept: intercept.getLeafsIntercept,
 	})
 
-	require.Equal(t, uint32(2), intercept.numRequests)
+	require.LessOrEqual(t, uint32(2), intercept.numRequests.Load())
```
StephenButtolph: At a glance, I thought "less than or equal" "2" "numRequests" and incorrectly thought that the thing under test needed to be leq 2.
```suggestion
	require.GreaterOrEqual(t, intercept.numRequests.Load(), uint32(2))
```
Since we actually want the thing under test to be at _least_ 2, should we be using `GreaterOrEqual`?

PR #1365, opened 2025-10-24
test: Add race detection for bindings tests

alarso16:
## Why this should be merged

The golang binding tests had a race with libevm registrations, so I also fixed the flake.

Closes ava-labs/avalanchego#4556

## How this works

Add race detection, registers libevm globally

## How this was tested

Existing CI

## Need to be documented?

No

No

### DIFF
--- a/accounts/abi/bind/util_test.go
+++ b/accounts/abi/bind/util_test.go
@@ -40,29 +40,29 @@ import (
 	"github.com/ava-labs/libevm/common"
 	"github.com/ava-labs/libevm/core/types"
 	"github.com/ava-labs/libevm/crypto"
-	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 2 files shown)


accounts/abi/bind/bind_test.go
```diff
@@ -2179,6 +2169,29 @@ func golangBindings(t *testing.T, overload bool) {
 			}
 		})
 	}
+
+	// We msut also write a main_test.go file for libevm registrations.
```
JonathanOppenheimer:
```suggestion
	// We must also write a main_test.go file for libevm registrations.
```
nit

ARR4N (approved):
Thank you! TIL that those aren't separate test binaries (I hate the `bind` tests).

PR #1367, opened 2025-10-24
Use AvalancheGo database package 

JonathanOppenheimer:
## Why this should be merged
With the uplift of the `database` package to AvalancheGo, we can de-duplicate the code and remove the package here. 

See this PR: https://github.com/ava-labs/avalanchego/pull/4331

## How this was tested
Existing CI

## Need to be documented?
No

No


JonathanOppenheimer:
As this PR requires updating the version of AvalancheGo which is on 1.24.9, this PR is blocked by https://github.com/ava-labs/coreth/pull/1352

plugin/evm/atomic/state/atomic_trie.go
```diff
@@ -72,7 +72,7 @@ func newAtomicTrie(
 	}
 
 	trieDB := triedb.NewDatabase(
-		rawdb.NewDatabase(database.WrapDatabase(atomicTrieDB)),
+		rawdb.NewDatabase(database.New(atomicTrieDB)),
```
alarso16: This naming doesn't make any sense to me. The `atomicTrieDB` is wrapped twice before becoming a `TrieDB`? I realize this is slightly out of scope but `TrieDB` has a specific meaning, not just being a database containing a trie. This should be fixed throughout the atomic state package eventually.

JonathanOppenheimer: I created an issue for this: https://github.com/ava-labs/coreth/issues/1368

PR #1371, opened 2025-10-27
Remove `customlogs` package 

JonathanOppenheimer:
## Why this should be merged
See this comment https://github.com/ava-labs/avalanchego/pull/4332#discussion_r2461502218 by @joshua-kim. The package is not needed and shouldn't be removed. Where it is used can be adjusted, or the function can just be in-lined. 

## How this was tested
Existing CI 

## Need to be documented?
No

No

### DIFF
--- a/eth/filters/filter.go
+++ b/eth/filters/filter.go
@@ -33,7 +33,6 @@ import (
 	"fmt"
 	"math/big"
 
-	"github.com/ava-labs/coreth/plugin/evm/customlogs"
 	"github.com/ava-labs/coreth/rpc"
 	"github.com/ava-labs/libevm/common"
 	"github.com/ava-labs/libevm/core/bloombits"


eth/filters/filter.go
```diff
@@ -357,11 +358,14 @@ func (f *Filter) checkMatches(ctx context.Context, header *types.Header) ([]*typ
 	if err != nil {
 		return nil, err
 	}
-	unfiltered = unfiltered[:0]
+	var receiptLogs [][]*types.Log
 	for _, receipt := range receipts {
-		unfiltered = append(unfiltered, receipt.Logs...)
+		receiptLogs = append(receiptLogs, receipt.Logs)
+	}
+	logs = logs[:0] // reset the slice
```
joshua-kim: Is this a bug? We're resetting the slice - so all of the txLogs that matched the filter are now gone from the slice and are replaced by things in `receiptLogs` that match the filter. 

The old code reset the slice because it put the unfiltered receipt logs into that var - and copied over things that matched the filter in to `logs`. If we wanted to simplify we could get rid of `unfiltered` and just have a single `logs` var that we just add matches to like:

```
	for _, r := range receipts {
		logs = append(logs, filterLogs(r.Logs, nil, nil, f.addresses, f.topics)...)
	}
```

JonathanOppenheimer: No, I don't think so. first pass: we get some logs and filter them to find matches, then we check the log to see if has a `TxHash`, yes yes return, if not the logs are incomplete, so we get the "real" full logs from reciepts, but first we  throw away the incomplete logs (`logs = logs[:0]`) and replace them with the complete filtered logs from receipts.

joshua-kim: I don't know the difference between receipts and underived logs (the sentinel value of empty hash is confusing to me - I don't know when it's supposed to be used) - so the reviewer on EVM should double-check me here.

powerslider (approved):

eth/filters/filter.go
ceyonur: The change itself looks good. I'm just confused why we did not sync this https://github.com/ethereum/go-ethereum/pull/25199 as it uses the body to fill these missing data. Anyway this PR lgtm.

PR #1379, opened 2025-10-29
feat(vmsync): dynamic state sync with coordinator, pivot cadence, and engine-driven target updates

powerslider:
## Why this should be merged

check ava-labs/avalanchego#4582 

## How this works

- Add Coordinator to orchestrate dynamic state sync, enforce pivot cadence, and manage queue execution.
- Introduce engine hook OnEngineAccept to enqueue accepted blocks and advance the sync target.
- Implement pivot policy (every N blocks) and idempotence (skip behind/equal, allow same-height reorgs).

## How this was tested

still not expliciitly

## Need to be documented?
no

no

resolves ava-labs/avalanchego#4582


plugin/evm/vmsync/client.go
```diff
@@ -268,62 +305,113 @@ func (client *client) acceptSyncSummary(proposedSummary message.Syncable) (block
...
+// stateSyncDynamic launches the dynamic state sync in a goroutine and returns the mode.
+func (c *client) stateSyncDynamic(ctx context.Context, registry *SyncerRegistry) block.StateSyncMode {
+	c.coordinator = NewCoordinator(
+		registry,
+		Callbacks{
+			FinalizeVM: func(_ message.Syncable) error {
+				return c.finishSync()
+			},
+			ApplyBlock: func(b EthBlockWrapper) error {
+				return c.Chain.BlockChain().InsertBlock(b.GetEthBlock())
```
alarso16: This isn't sufficient. You need to do a very large portion of the operations in `wrappedBlock.Verify()` and `wrappedBlock.Accept()`.

plugin/evm/vmsync/coordinator.go
```diff
@@ -0,0 +1,194 @@
...
+	g := co.syncerRegistry.StartAsync(cctx, initial)
+
+	co.state.Store(int32(StateRunning))
+
+	go func() {
+		if err := g.Wait(); err != nil {
+			co.finish(err)
+			return
+		}
+		// All syncers finished successfully: finalize VM and execute the queued batch
```
alarso16: Have they finished? There might still be outstanding UpdateSyncTarget stuff (my solution was `Finalize`)

plugin/evm/vmsync/sync_target.go
```diff
@@ -0,0 +1,50 @@
...
+	"context"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/snow/engine/snowman/block"
+	"github.com/ava-labs/libevm/common"
+
+	"github.com/ava-labs/coreth/plugin/evm/message"
+)
+
+var _ message.Syncable = (*syncTarget)(nil)
```
alarso16: I don't think this is the right interface to satisfy. We should probably have the atomic root as well, but also, `Accept` is completely meaningless. This is intended to be something passed back to the engine to invoke the state sync at the beginning (this calls `acceptSyncSummary` currently)

plugin/evm/vmsync/client.go
```diff
@@ -104,88 +113,105 @@ type Client interface {
...
 
-	if err := client.registerSyncers(registry); err != nil {
+	ethb := b.GetEthBlock()
+	// Best-effort enqueue, ignored unless Running.
+	if ok := c.coordinator.AddBlock(b); !ok {
+		// TODO(powerslider): should we return an error here?
+		log.Warn("could not enqueue block for post-finalization execution", "hash", ethb.Hash(), "height", ethb.NumberU64())
+	}
+
+	syncTarget := newSyncTarget(ethb.Hash(), ethb.Root(), ethb.NumberU64())
```
alarso16: It's weird we have to make two separate calls to the coordinator here

plugin/evm/vmsync/client.go
```diff
@@ -104,88 +117,155 @@ type Client interface {
...
+// OnEngineAccept should be invoked by the engine when a block is accepted.
+// It best-effort enqueues the block for post-finalization execution and
+// advances the coordinator's sync target using the block's hash/root/height.
+func (c *client) OnEngineAccept(b EthBlockWrapper) error {
+	if !c.DynamicStateSyncEnabled || c.coordinator == nil {
+		return nil
+	}
+
+	// Skip notification if we're already processing the queue to avoid recursion.
+	if c.coordinator.CurrentState() == StateExecutingBatch {
```
alarso16: We still need to enqueue it. Say there's 5 blocks left to process, we can't process this block as normal, since its state probably depends on theirs.

plugin/evm/vmsync/client.go
```diff
@@ -104,88 +117,155 @@ type Client interface {
...
+
+// OnEngineVerify should be invoked by the engine when a block is verified.
+// It best-effort enqueues the block for post-finalization execution.
+func (c *client) OnEngineVerify(b EthBlockWrapper) error {
+	if !c.DynamicStateSyncEnabled || c.coordinator == nil {
+		return nil
+	}
+
+	// Skip notification if we're already processing the queue to avoid recursion.
+	if c.coordinator.CurrentState() == StateExecutingBatch {
```
alarso16: There's a lot of duplicated logic in these calls, can we push them down into the coordinator?

plugin/evm/wrapped_block.go
```diff
@@ -91,11 +92,13 @@ func wrapBlock(ethBlock *types.Block, vm *VM) (*wrappedBlock, error) {
...
 // Accept implements the snowman.Block interface
-func (b *wrappedBlock) Accept(context.Context) error {
-	vm := b.vm
-	// Although returning an error from Accept is considered fatal, it is good
-	// practice to cleanup the batch we were modifying in the case of an error.
-	defer vm.versiondb.Abort()
+func (b *wrappedBlock) Accept(ctx context.Context) error {
+	// Notify sync client that engine accepted a block.
+	if client := b.vm.SyncerClient(); client != nil {
+		if err := client.OnEngineAccept(b); err != nil {
```
alarso16: I don't think this will work. I know you made the comment about recursively finding a sync client, but you might just need to pass in a callback. Like:
```
if err := client.OnEngineAccept(b, (*wrappedBlock).accept)); err != nil {
```

plugin/evm/vmsync/client.go
alarso16: Nice, will you cleanup this function a little bit

plugin/evm/vmsync/client.go
alarso16: nit: I think it would be easier to pass the block to the coordinator, and then the coordinator can make the sync target, unless there's an import issue.

plugin/evm/wrapped_block.go
```diff
@@ -172,7 +189,21 @@ func (b *wrappedBlock) handlePrecompileAccept(rules extras.Rules) error {
 
 // Reject implements the snowman.Block interface
 // If [b] contains an atomic transaction, attempt to re-issue it
-func (b *wrappedBlock) Reject(context.Context) error {
+func (b *wrappedBlock) Reject(_ context.Context) error {
```
alarso16: You change this a lot. I think we prefer no name rather than `_`
```suggestion
func (b *wrappedBlock) Reject(context.Context) error {
```

PR #1383, opened 2025-10-30
test: Skip flaky ChainIndexer test

### LINKED ISSUE #1194: Fix `TestChainIndexerWithChildren` flaky test
See https://github.com/ava-labs/coreth/actions/runs/17594990875/job/49984976276?pr=886

alarso16:
## Why this should be merged

This test is occasionally flaky. 

Closes #1194

## How this works

This test is isolated from most of the core package, only effected by the chain indexer and bloombits code. However, we have no significant diff here from go-ethereum until they eventually replaced this entire functionality. We should not be editing this code anyway, so it's safe to skip.

## How this was tested

UT

## Need to be documented?

No

No


core/chain_indexer_test.go
```diff
@@ -51,6 +51,7 @@ func TestChainIndexerSingle(t *testing.T) {
 // Runs multiple tests with randomized parameters and different number of
 // chain backends.
 func TestChainIndexerWithChildren(t *testing.T) {
+	t.Skip("Flaky test from go-ethereum")
```
JonathanOppenheimer: for this again, I'd add some of the context from your PR comment so it's not lost!

PR #1388, opened 2025-10-31
refactor: use upstream bloombits package

alarso16:
## Why this should be merged

I was looking at a flaky test and discovered this package is exactly the same in libevm.

## How this works

Use libevm core/bloombits. Compare with this:

```
git diff --no-index ./core/bloombits ../libevm/core/bloombits
```

## How this was tested

UT

## Need to be documented?

No

No


JonathanOppenheimer (approved):
I love cleanups like this

StephenButtolph (approved):

PR #1391, opened 2025-11-03
ci: Add lint-fix to taskfile

alarso16:
## Why this should be merged

Some automatable fixes are rather difficult to do manually - specifically the auto fixes available form the avalanchego linter. This allows easy usage of automated lint fixes. This might not be the idiomatic way of restructuring these scripts, but I'm not a bash expert 🤷‍♂️

## How this works

Refactors upstream lint file creation into a separate file, altering a global variable that can be used anywhere when source'd. 

## How this was tested

Locally, any automatable fix can be created (like import ordering), and one can check that the fix is applied. Additionally, existing CI passes.

## Need to be documented?

No

No


scripts/lint.sh
```diff
@@ -122,26 +55,26 @@ function test_license_header {
       || return 1
   fi
 
-  if [[ ${#DEFAULT_FILES[@]} -gt 0 ]]; then
+  if [[ ${#AVALANCHE_FILES[@]} -gt 0 ]]; then
```
JonathanOppenheimer: Avalanche files makes sense to me.

scripts/lint_setup.sh
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
```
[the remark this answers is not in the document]
alarso16: We definitely could, I just thought this would be more convenient. But potentially not

scripts/lint.sh
```diff
@@ -2,6 +2,8 @@
 
 set -euo pipefail
 
+source ./scripts/lint_setup.sh
```
maru-ava: Maybe move this after the check for running from repo root given that the moved section depends on this to be true?

scripts/lint_setup.sh
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env bash
```
maru-ava: (No action required) Maybe document usage of what is effectively a library script e.g.

```bash
 # lint_setup.sh - Shared linting configuration setup
 #
 # Exports:
 #   AVALANCHE_FILES - Array of avalanche-specific Go files to lint
 #   UPSTREAM_FILES - Array of upstream Go files to lint separately
 #   AVALANCHE_LINT_FILE - Path to temporary avalanche-specific lint config
 #
 # Usage:
 #   source ./scripts/lint_setup.sh
 #   setup_lint
```

scripts/lint_setup.sh
```diff
@@ -0,0 +1,76 @@
...
+  for f in "${UPSTREAM_FILES[@]}"; do
+    default_exclude_args+=(! -path "$f")
+  done
+
+  # Now find default files (exclude already licensed ones)
+  # shellcheck disable=SC2034  # used by external scripts after sourcing
+  mapfile -t AVALANCHE_FILES < <(find . "${find_filters[@]}" "${default_exclude_args[@]}")
+
+  # copy avalanche file to temp directory to edit
+  AVALANCHE_LINT_FILE="$(mktemp -d)/.avalanche-golangci.yml"
```
maru-ava: Best practice when creating temporary files or directories is to make a best-effort attempt at removal on exit e.g.

```bash
   function setup_lint {
     ...
     TMP_DIR="$(mktemp -d)"
     trap rm -rf "${TMP_DIR}" EXIT
     AVALANCHE_LINT_FILE="${TMP_DIR}/.avalanche-golangci.yml"
     ...
   }
```

alarso16: Neat, hadn't heard of this! I'm also removing the output displaying where the file is if it's immediately going to be deleted

scripts/lint.sh
```diff
@@ -2,6 +2,7 @@
 
 set -euo pipefail
 
+
```
maru-ava: ni: non-functional formatting change

PR #1404, opened 2025-11-12
ci: Filters test corrections

### LINKED ISSUE #1377: Fix `TestFilters` flaky test
See https://github.com/ava-labs/coreth/actions/runs/18884698394/job/53896938049

alarso16:
## Why this should be merged

The timeout is flaky occasionally (see #1377) due to large differences in context handling with go-ethereum. I don't think this is worth investigating prior to repo merger.

Additionally, there is a race in a test because it accesses unexported variables

Closes #1377

## How this works

Skips test, adds lock

## How this was tested

UT

## Need to be documented?

No

No


JonathanOppenheimer (approved):

alarso16:
> I assume the list of flakes to fix properly is self-documenting? If it's skipped we need to properly fix it as the some point? (Post monorepo)

grep documents it! Seriously though, I don't know if there's a point in making an issue since they're so searchable and we hope to delete this code anyway. Not like avalanchego doesn't also have its share of `t.Skip`s...

PR #1406, opened 2025-11-12
Removed `avax.version` API

StephenButtolph:
## Why this should be merged

This information is already exposed by [info.getNodeVersion](https://build.avax.network/docs/rpcs/other/info-rpc#infogetnodeversion)

## How this works

Removes `avax.version`

## How this was tested

CI

## Need to be documented?

Yes

Yes

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -2,6 +2,10 @@
 
 ## Pending Release
 
+- Removes `avax.version` API
+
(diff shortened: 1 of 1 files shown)


JonathanOppenheimer (approved):
Is it worth putting into release notes for people who rely on this API method? 

Something along the lines of "removed `avax.version`, use `info.GetNodeVersion` in this manner etc. etc. instead"

Issue #1410, opened 2025-11-17
Context cancellation during shutdown logged as ERROR instead of INFO


powerslider:
**Describe the bug**

During graceful shutdown, when state syncers are cancelled via context cancellation, the registry logs these cancellations as `ERROR` level instead of `INFO`. This is misleading because context cancellation during shutdown is expected behavior, not an error condition. The ERROR logs create noise in monitoring systems and make it harder to identify actual problems.

**To Reproduce**

1. Start an AvalancheGo node with state syncing enabled.
2. Allow state sync to begin.
3. Gracefully shutdown the node (e.g., send SIGTERM or press Ctrl+C).
4. Observe the logs during shutdown.

**Expected behavior**

Context cancellation errors during shutdown should be logged as `INFO` level with a message like:

```
INFO syncer cancelled name="Code Syncer" summary=... height=...
```

Only actual errors (non-cancellation) should be logged as `ERROR`:

```
ERROR failed syncing name="Code Syncer" summary=... height=... err=<actual error>
```

**Logs**

Example of the problematic ERROR log during shutdown:

```
[11-17|12:46:44.161] ERROR <C Chain> plugin/evm/vmsync/registry.go:70 failed syncing
name="Code Syncer"
summary=0x87a65a4d1e92dab57ab1eaa2406fa3136f8498287e03eed2d3bc79c8a46142d3
height=47,775,744
err="could not get code (CodeRequest(Hashes=0x4f8836f69ba20dca430484f48c951316ea09138eefe2587b4b450447ece598c4, 0xfe9eb924450fc28ae05c8bb8be9386d8270cf7779921a66017e3d3ae410c4610, 0x5a42e53889f320f778ee2d98f9a18accf3bc5999966688939d9dd21a564b7d63, 0x225a631b0e2d6f1514f1ac51910ee735f6b0cec1ae95312358e1fcf13c1521ac, 0x332cf2b3017942cadf0656e3645de1e57a90ec199a3e971736da00e6ef2e3eca)): context canceled"
```

Logs can be found at `~/.avalanchego/logs/` by default.

**Operating System**

Any OS (reproduced on macOS, but should affect all platforms)

**Impact:**

- **Misleading logs** -  ERROR logs during normal shutdown operations create noise and make it harder to identify actual problems.
- **Monitoring confusion** - Error rate metrics may be inflated by expected shutdown cancellations.
- **Debugging difficulty** - Real errors may be obscured by cancellation noise.

**Additional Context:**

[...]

PR #1411, opened 2025-11-17
fix(vmsync,statesync): ensure syncers detect cancellation during graceful shutdown and fix registry cancellation logging

powerslider:
## Why this should be merged

During graceful shutdown, syncers cancelled via context cancellation were being logged as ERROR level. This is misleading since cancellation during shutdown is expected behavior, not an error condition.

Check #1410 

### Enable correct syncer cancellation logging in the `SyncerRegistry`

- Use `errors.Is()` to detect `context.Canceled` and `context.DeadlineExceeded` (handles wrapped errors) and log as INFO instead of ERROR.

- Separate `RunSyncerTasks()` logic into a synchronous wrapper and `StartAsync()` method for async execution to gain more flexibility and handle more use cases.

- Add early return optimization when context is already cancelled.

### Fix `StateSyncer` context propagation

During graceful shutdown, the State Syncer was hanging because multiple blocking operations did not check context cancellation. When shutdown occurred, these operations would block indefinitely, preventing syncers from detecting cancellation and exiting gracefully.

- Add `context.Context` parameter to `LeafSyncTask.OnLeafs()` interface to enable context propagation through the leaf processing call chain.

- Update `CodeQueue.AddCode()` to accept context and check `ctx.Done()` before blocking on channel sends, preventing indefinite blocking when Code Syncer stops consuming during shutdown.

- Update all OnLeafs implementations (`mainTrieTask`, `storageTrieTask`, `trieSegment`, atomic syncer) to accept and pass context through the call chain.

- Add context parameter to `startSyncing()` and `createSegments()` methods, checking cancellation before blocking channel sends to the segments work queue.

- Add context cancellation check in BlockSyncer before checking blocks on disk, ensuring it responds during the initial scan phase.

- Update `sync/client/leaf_syncer.go` to pass context to `OnLeafs()` callbacks.

This ensures all syncers detect cancellation immediately and exit gracefully instead of hanging until timeout.

## How this was tested

- Add tests for cancellation scenarios (`Canceled`, `DeadlineExceeded`, early return).
- Fix flakiness by adding `WaitGroup` synchronization and replacing channel-based coordination.
- Refactor tests to use `t.Context()` and extract common helpers.
- Performed a manual test by starting a node on Fuji Testnet. Check [here](https://github.com/ava-labs/coreth/pull/1411#issuecomment-3547650561).

## Need to be documented?

no

no

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -127,6 +127,9 @@ var (
 	latestGasCapacityGauge = metrics.NewRegisteredGauge("chain/latest/gas/capacity", nil)
 	latestGasTargetGauge   = metrics.NewRegisteredGauge("chain/latest/gas/target", nil)
 
+	latestMinDelayGauge       = metrics.NewRegisteredGauge("chain/latest/mindelay", nil)
+	latestMinDelayExcessGauge = metrics.NewRegisteredGauge("chain/latest/mindelay/excess", nil)
+
 	ErrRefuseToCorruptArchiver = errors.New("node has operated with pruning disabled, shutting down to prevent missing tries")
 
 	errFutureBlockUnsupported  = errors.New("future block insertion not supported")
@@ -1129,6 +1132,14 @@ func (bc *BlockChain) Accept(block *types.Block) error {
 		latestGasExcessGauge.Update(int64(s.Gas.Excess))
 		latestGasCapacityGauge.Update(int64(s.Gas.Capacity))
 		latestGasTargetGauge.Update(int64(s.Target()))
+	}
+	if extraConfig.IsGranite(block.Time()) {
+		extraHeader := customtypes.GetHeaderExtra(block.Header())
+		if extraHeader.MinDelayExcess != nil {
+			delayExcess := *extraHeader.MinDelayExcess
+			latestMinDelayGauge.Update(int64(delayExcess.Delay()))
+			latestMinDelayExcessGauge.Update(int64(delayExcess))
+		}
 	}
 	return nil
 }
--- a/plugin/evm/block_builder.go
+++ b/plugin/evm/block_builder.go
@@ -49,6 +49,9 @@ type blockBuilder struct {
 	// but at least after a minimum delay of minBlockBuildingRetryDelay.
 	lastBuildParentHash common.Hash
 	lastBuildTime       time.Time
+
+	chainHeadHash   common.Hash
+	mempoolHeadHash common.Hash
 }
 
 // NewBlockBuilder creates a new block builder. extraMempool is an optional mempool (can be nil) that
@@ -104,10 +107,13 @@ func (b *blockBuilder) awaitSubmittedTxs() {
 		extraChan = b.extraMempool.SubscribePendingTxs()
 	}
 
+	events := make(chan core.NewTxPoolReorgEvent)
+	sub := b.txPool.SubscribeNewReorgEvent(events)
+
 	b.shutdownWg.Add(1)
 	go b.ctx.Log.RecoverAndPanic(func() {
 		defer b.shutdownWg.Done()
-
+		defer sub.Unsubscribe()
 		for {
 			select {
 			case <-txSubmitChan:
@@ -118,6 +124,12 @@ func (b *blockBuilder) awaitSubmittedTxs() {
 				b.signalCanBuild()
 			case <-b.shutdownChan:
 				return
+			case event := <-events:
+				if event.Head == nil || event.Head.Number == nil {
+					log.Warn("nil head or block number in tx pool reorg event")
+					continue
+				}
+				b.setMempoolHeadHash(event.Head.Hash())
 			}
 		}
 	})
@@ -156,7 +168,7 @@ func (b *blockBuilder) waitForEvent(ctx context.Context, currentHeader *types.He
 func (b *blockBuilder) waitForNeedToBuild(ctx context.Context) (time.Time, common.Hash, error) {
 	b.buildBlockLock.Lock()
 	defer b.buildBlockLock.Unlock()
-	for !b.needToBuild() {
+	for !b.needToBuild() || b.pendingPoolUpdate() {
 		if err := b.pendingSignal.Wait(ctx); err != nil {
 			return time.Time{}, common.Hash{}, err
 		}
@@ -198,4 +210,34 @@ func minNextBlockTime(parent *types.Header) time.Time {
 	// so this should not overflow
 	requiredDelay := time.Duration(acp226DelayExcess.Delay()) * time.Millisecond
 	return parentTime.Add(requiredDelay)
+}
+
+func (b *blockBuilder) setChainHeadHash(hash common.Hash) {
+	b.buildBlockLock.Lock()
+	defer b.buildBlockLock.Unlock()
+
+	b.chainHeadHash = hash
+
+	if b.pendingPoolUpdate() {
+		return
+	}
+
+	b.pendingSignal.Broadcast()
+}
+
+func (b *blockBuilder) setMempoolHeadHash(hash common.Hash) {
+	b.buildBlockLock.Lock()
+	defer b.buildBlockLock.Unlock()
+
+	b.mempoolHeadHash = hash
+
+	if b.pendingPoolUpdate() {
+		return
+	}
+
+	b.pendingSignal.Broadcast()
+}
+
+func (b *blockBuilder) pendingPoolUpdate() bool {
+	return b.chainHeadHash != b.mempoolHeadHash
 }
--- a/plugin/evm/vm.go
+++ b/plugin/evm/vm.go
@@ -809,6 +809,8 @@ func (vm *VM) initBlockBuilding() error {
 	// NOTE: gossip network must be initialized first otherwise ETH tx gossip will not work.
 	vm.builderLock.Lock()
 	vm.builder = vm.NewBlockBuilder(vm.extensionConfig.ExtraMempool)
+	vm.builder.setChainHeadHash(vm.blockChain.CurrentBlock().Hash())
+	vm.builder.setMempoolHeadHash(vm.blockChain.CurrentBlock().Hash())
 	vm.builder.awaitSubmittedTxs()
 	vm.builderLock.Unlock()
 
@@ -1027,6 +1029,8 @@ func (vm *VM) SetPreference(ctx context.Context, blkID ids.ID) error {
 		return fmt.Errorf("failed to set preference to %s: %w", blkID, err)
 	}
 
+	vm.setPendingBlock(block.GetEthBlock().Hash())
+
 	return vm.blockChain.SetPreference(block.GetEthBlock())
 }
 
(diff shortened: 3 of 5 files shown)


plugin/evm/vmsync/registry_test.go
```diff
@@ -301,12 +300,225 @@ func TestSyncerRegistry_NoSyncersRegistered(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(t.Context())
 	t.Cleanup(cancel)
 
 	require.NoError(t, registry.RunSyncerTasks(ctx, newTestClientSummary(t)))
 }
 
+func TestSyncerRegistry_ContextCancellationErrors(t *testing.T) {
+	t.Parallel()
+
+	// Test scenario that reproduces the issue where context cancellation during shutdown
```
alarso16: I don't think this is the right way to write this comment, and also I don't think it's the right place to put it. Can you describe what is being tested on line 308, and what is the expected behavior? Why this is tested (a weird error log) won't be informative later.

plugin/evm/vmsync/registry_test.go
```diff
@@ -301,12 +300,225 @@ func TestSyncerRegistry_NoSyncersRegistered(t *testing.T) {
...
+			// Wait for syncers to start.
+			waitTimeout := 2 * time.Second
+			if tt.wantErr == context.DeadlineExceeded {
+				// For deadline exceeded, use shorter timeout since deadline is tight.
+				waitTimeout = 100 * time.Millisecond
+			}
+			utilstest.WaitGroupWithTimeout(t, startedWG, waitTimeout, "timed out waiting for syncers to start")
+
+			// Trigger cancellation for Canceled test (DeadlineExceeded will expire naturally).
+			if tt.wantErr == context.Canceled {
```
alarso16: Can you move this closer to where `cancel` is declared?

plugin/evm/vmsync/registry_test.go
```diff
@@ -301,12 +300,225 @@ func TestSyncerRegistry_NoSyncersRegistered(t *testing.T) {
...
+				cancel()
+			}
+
+			err := utilstest.WaitErrWithTimeout(t, doneCh, 4*time.Second)
+
+			// Verify error is returned directly (not wrapped).
+			require.ErrorIs(t, err, tt.wantErr)
+			require.Equal(t, tt.wantErr, err, "error should be returned directly, not wrapped")
```
alarso16: Does it matter if it's wrapped?

plugin/evm/vmsync/registry_test.go
```diff
@@ -301,12 +300,225 @@ func TestSyncerRegistry_NoSyncersRegistered(t *testing.T) {
...
+	require.ErrorIs(t, err, context.Canceled)
+	require.Equal(t, context.Canceled, err, "error should be context.Canceled directly, not wrapped")
+
+	// Verify that syncers were never started (early return optimization).
+	for i, mockSyncer := range mockSyncers {
+		require.False(t, mockSyncer.started, "Syncer %d should not have been started due to early return", i)
+	}
+}
+
+func TestSyncerRegistry_MixedCancellationAndSuccess(t *testing.T) {
```
alarso16: I think you could delete the two tests above and just use this test.

plugin/evm/vmsync/registry.go
```diff
@@ -51,22 +51,46 @@ func (r *SyncerRegistry) Register(syncer synccommon.Syncer) error {
...
+
+	log.Info("all syncers completed successfully", "count", len(r.syncers), "summary", summary.GetBlockHash().Hex())
+
+	return nil
+}
+
+// StartAsync launches all registered syncers and returns an [errgroup.Group]
+// whose Wait() completes when all syncers exit. The context returned will be
+// cancelled when any syncer fails, propagating shutdown to the others.
+func (r *SyncerRegistry) StartAsync(ctx context.Context, summary message.Syncable) *errgroup.Group {
```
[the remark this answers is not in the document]
powerslider: kinda true, but since I need this as well for the dynamic state sync implementation and is also related to better restructuring for better context cancellation handling, I decided to do it now, cause it also impacted the way I write the tests for context cancellation.

plugin/evm/vmsync/registry_test.go
```diff
@@ -301,12 +300,225 @@ func TestSyncerRegistry_NoSyncersRegistered(t *testing.T) {
...
+		timeout       time.Duration // Timeout duration (only used when wantErr is DeadlineExceeded)
+	}{
+		{
+			name:          "context canceled during active sync",
+			wantErr:       context.Canceled,
+			numSyncers:    3,
+			syncerTimeout: 5 * time.Second,
+		},
+		{
+			name:          "context deadline exceeded",
```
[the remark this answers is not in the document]
powerslider: yes, because I am checking for both `context.Canceled` and `context.DeadlineExceeded`. The conditions that trigger them are different, so you are testing two different scenarios.

plugin/evm/vmsync/registry_test.go
powerslider: yes, because you have to unwrap in order to determine if the error is `context.Canceled` in which case it is not a real error, but rather expected behaviour.

plugin/evm/vmsync/registry_test.go
```diff
@@ -301,12 +300,225 @@ func TestSyncerRegistry_NoSyncersRegistered(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(t.Context())
+	defer cancel()
+
+	// Start syncers - the wrapped error should be detected by errors.Is().
+	g := registry.StartAsync(ctx, newTestClientSummary(t))
+
+	err := g.Wait()
+
+	// Verify that errors.Is correctly identifies the wrapped context.Canceled
+	require.ErrorIs(t, err, context.Canceled, "wrapped context.Canceled should be detected by errors.Is")
```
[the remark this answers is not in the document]
powerslider: No, this tests the already unwrapped error. It is testing if `errors.Is` did its job of detecting the underlying `context.Canceled`.

plugin/evm/vmsync/registry_test.go
```diff
@@ -223,58 +228,50 @@ func TestSyncerRegistry_ErrorPropagatesAndCancelsOthers(t *testing.T) {
 
 	registry := NewSyncerRegistry()
 
-	ctx, cancel := utilstest.NewTestContext(t)
+	ctx, cancel := context.WithCancel(t.Context())
```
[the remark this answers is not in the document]
alarso16: See #1412

plugin/evm/vmsync/doubles_test.go
```diff
@@ -58,18 +62,18 @@ func NewErrorSyncer(trigger <-chan struct{}, errToReturn error) FuncSyncer {
...
-// NewCancelAwareSyncer closes `started` as soon as Sync begins, then waits for
+// NewCancelAwareSyncer calls startedWG.Done() as soon as Sync begins, then waits for
 // either:
-//   - `ctx` cancellation: closes `canceled` and returns ctx.Err; or
+//   - `ctx` cancellation: calls canceledWG.Done() and returns ctx.Err; or
 //   - `timeout` elapsing: returns an error indicating a timeout.
 //
 // Useful for asserting that cancellation propagates to the syncer under test.
-func NewCancelAwareSyncer(started, canceled chan struct{}, timeout time.Duration) FuncSyncer {
+func NewCancelAwareSyncer(startedWG, canceledWG *sync.WaitGroup, timeout time.Duration) FuncSyncer {
```
alarso16: Providing a `canceledWG` feels bug prone. Why isn't the channel sufficient?

plugin/evm/vmsync/registry_test.go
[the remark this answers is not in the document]
alarso16: I'm unconvinced both are necessary, but I digress

plugin/evm/vmsync/registry_test.go
alarso16: I think that you do not need to describe the situation that led to create the test (testing context cancellation is very typical). Additionally, for the remaining parts:
1. I don't think we need to guarantee that it isn't wrapped
2. You don't test  that we log as INFO, so it should be removed from the descriptions

plugin/evm/vmsync/registry_test.go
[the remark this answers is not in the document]
alarso16: That sounds like testing the `require.ErrorIs` implementation

sync/statesync/code_queue_test.go
```diff
@@ -96,13 +96,13 @@ func TestCodeQueue(t *testing.T) {
...
 				for _, add := range tt.addCode {
-					require.NoErrorf(t, q.AddCode(add), "%T.AddCode(%v)", q, add)
+					require.NoErrorf(t, q.AddCode(t.Context(), add), "%T.AddCode(%v)", q, add)
 				}
 
 				if tt.quitInsteadOfFinalize {
 					close(quit)
 					<-recvDone
-					require.ErrorIsf(t, q.AddCode(tt.addCodeAfter), errFailedToAddCodeHashesToQueue, "%T.AddCode() after `quit` channel closed", q)
+					require.ErrorIsf(t, q.AddCode(t.Context(), tt.addCodeAfter), errFailedToAddCodeHashesToQueue, "%T.AddCode() after `quit` channel closed", q)
```
alarso16: Our convention is to return the err first for `ErrorIs`
```suggestion
					err := q.AddCode(t.Context(), tt.addCodeAfter)
					require.ErrorIsf(t, err, errFailedToAddCodeHashesToQueue, "%T.AddCode() after `quit` channel closed", q)
```
Not sure why it wasn't caught by the linter though ...

plugin/evm/vmsync/registry_test.go
```diff
@@ -307,6 +299,164 @@ func TestSyncerRegistry_NoSyncersRegistered(t *testing.T) {
 	require.NoError(t, registry.RunSyncerTasks(ctx, newTestClientSummary(t)))
 }
 
+func TestSyncerRegistry_ContextCancellationErrors(t *testing.T) {
+	t.Parallel()
```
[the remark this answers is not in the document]
powerslider: The distinction is as follows:
- Function-level `t.Parallel()`:
   - Marks the test function as parallelizable with other test functions.
   - Allows `TestSyncerRegistry_ContextCancellationErrors` to run in parallel with `TestSyncerRegistry_ConcurrentStart`, `TestSyncerRegistry_ErrorPropagatesAndCancelsOthers`, etc.
- Subtest-level `t.Parallel()`:
   - Marks subtests as parallelizable with each other. Allows the `context canceled` and `context deadline exceeded` subtests to run in parallel. 

If your test logic can afford independence of other actions in its logic, I would always prefer to parallelize on all possible levels.

sync/blocksync/syncer.go
```diff
@@ -78,6 +78,11 @@ func (s *BlockSyncer) Sync(ctx context.Context) error {
 	// first, check for blocks already available on disk so we don't
 	// request them from peers.
 	for blocksToFetch > 0 {
+		// Check for context cancellation before checking each block.
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+
```
ceyonur: this looks redundant

plugin/evm/vmsync/registry.go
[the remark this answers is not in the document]
powerslider: no, because I will need it in another PR as well, but if you insist on unexporting it now, I don't mind. I can export it later again.

sync/blocksync/syncer.go
powerslider: `ReadBlock` is a fast database read, and the loop breaks as soon as a block isn't found, so it typically completes quickly. I added it just in case, but we could do without it.

ceyonur (approved):
I can confirm that this PR resolves the timeout issue in Shutdown.

ceyonur (commented):
Seems there is something weird happened on my node and it got stuck during shutdown. not sure if this was related with this or something else, but I removed this from the queue to investigate it further

PR #1414, opened 2025-11-18
Remove `TestMustNotImport` 

JonathanOppenheimer:
## Why this should be merged

Now that the `lint_allowed_eth_imports` test is added in https://github.com/ava-labs/coreth/pull/1348, and with the recent changes to libevm registration, `TestMustNotImport` should no longer be necessary, so we can remove it. 

## How this works
This just removes code. 

## How this was tested
CI

## Need to be documented?
No

No

### DIFF
--- a/plugin/evm/imports_test.go
+++ b/plugin/evm/imports_test.go
@@ -19,8 +19,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-const repo = "github.com/ava-labs/coreth"
-
 // TestLibevmImportsAreAllowed ensures that all libevm imports in the codebase
 // are explicitly allowed via the eth-allowed-packages.txt file.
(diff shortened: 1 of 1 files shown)


plugin/evm/imports_test.go
```diff
@@ -17,76 +17,8 @@ import (
...
-	}
-
-	// Start collecting dependencies
-	for _, pkg := range pkgs {
-		collectDeps(pkg)
-	}
-	return deps, nil
-}
-
-func TestMustNotImport(t *testing.T) {
```
alarso16: This doesn't seem resolved. Why can we remove this?

alarso16 (approved):
Given that we no longer need to support this (I'm deferring to @ARR4N for this point), the PR looks good to me

PR #1416, opened 2025-11-18
feat: enable archival mode for Firewood

RodrigoVillar:
## Why this should be merged

Updates `firewood-go-ethhash` to use the latest Firewood version (v0.0.15), which enables archival mode for Firewood.

## How this works

- Updates code to use new `ffi.Hash` type
- Removes restriction on using Firewood with pruning disabled
    - Imposes new restriction on opening Firewood without missing trie repopulation.
    - Firewood can now be configured to run in archival mode.
- Updates + adds tests

## How this was tested

- Updates the `TestArchiveBlockChainSnapsDisabled()` test to use Firewood
- Adds a `TestPruningToNonPruning()` test to verify expected behavior of HashDB and Firewood when switching from pruning to archival mode.

## Need to be documented?

Yes

Yes

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -6,6 +6,7 @@
 - Removes `customethclient` package in favor of `ethclient` package and temporary type registrations (`WithTempRegisteredLibEVMExtras`)
   - Also removes blockHook extension in `ethclient` package.
 - Enables Firewood to run with pruning disabled.
+  - This change modifies the filepath of Firewood and any nodes using Firewood will need to resync.
 
 ## [v0.16.0](https://github.com/ava-labs/coreth/releases/tag/v0.16.0)
 


triedb/firewood/database.go
```diff
@@ -493,7 +495,7 @@ func (reader *reader) Node(_ common.Hash, path []byte, _ common.Hash) ([]byte, e
 	// This function relies on Firewood's internal locking to ensure concurrent reads are safe.
 	// This is safe even if a proposal is being committed concurrently.
 	start := time.Now()
-	result, err := reader.db.fwDisk.GetFromRoot(reader.root.Bytes(), path)
+	result, err := reader.db.fwDisk.GetFromRoot(ffi.Hash(reader.root), path)
```
alarso16: nit: maybe store the FFI root in the struct instead? I don't have a strong opinion on that

core/blockchain_ext_test.go
```diff
@@ -166,27 +166,42 @@ func copyMemDB(db ethdb.Database) (ethdb.Database, error) {
...
-		name := e.Name()
-		data, err := os.ReadFile(filepath.Join(src, name))
-		require.NoError(t, err)
-		info, err := e.Info()
-		require.NoError(t, err)
-		require.NoError(t, os.WriteFile(filepath.Join(dst, name), data, info.Mode().Perm()))
-	}
+		// Calculate the relative path from src
+		relPath, err := filepath.Rel(src, path)
+		require.NoError(err)
```
alarso16: Why not return the error upward? I don't really know how `filepath.Walk` works, but I can look into it too

core/blockchain_test.go
```diff
@@ -409,6 +423,164 @@ func TestBlockChainOfflinePruningUngracefulShutdown(t *testing.T) {
...
+	// the last accepted block) upon restart.
+	for _, block := range prunedBlocks[:numStates-1] {
+		if blockchain.HasState(block.Root()) {
+			t.Fatalf("Expected blockchain to be missing state for intermediate block %d with pruning enabled", block.NumberU64())
+		}
+	}
+
+	blockchain.Stop()
+
+	archiveConfig := &CacheConfig{
```
alarso16: There's a lot of config copying in this file, maybe we should make a better abstraction for it (out of scope here though)

plugin/evm/vm.go
alarso16: I think we need some way to prevent the repopulate trees function from being called. The only reasonable places to return an error is in the config validation or in the function itself

triedb/firewood/database.go
RodrigoVillar: Done: e2f86b1

plugin/evm/vm.go
RodrigoVillar: Done: 3c4446c

alarso16 (commented):
You should also update `plugin/evm/config/config.md` and `RELEASES.md`

triedb/firewood/database.go
```diff
@@ -62,11 +67,12 @@ type ProposalContext struct {
 }
 
 type Config struct {
-	FilePath             string
+	ChainDataDir         string
 	CleanCacheSize       int  // Size of the clean cache in bytes
 	FreeListCacheEntries uint // Number of free list entries to cache
 	Revisions            uint
 	ReadCacheStrategy    ffi.CacheStrategy
+	PruningDisabled      bool
```
alarso16: I think a more logical name for this is "ArchiveMode" or something like that. Setting a negative boolean to positive is error-prone

triedb/firewood/database.go
```diff
@@ -62,11 +67,12 @@ type ProposalContext struct {
 }
 
 type Config struct {
-	FilePath             string
+	ChainDataDir         string
 	CleanCacheSize       int  // Size of the clean cache in bytes
 	FreeListCacheEntries uint // Number of free list entries to cache
 	Revisions            uint
```
alarso16: Maybe we should clarify this is revisions kept in memory, and can't be less than 2? At least in a comment

alarso16 (commented):
Will you rename this PR? It's mostly about enabling archival mode for firewood, which is a feature, not just updating dependencies

triedb/firewood/database.go
RodrigoVillar: Done: 0629148

alarso16 (approved):

